Issue #178, opened 2020-01-07
Wait.UntilPortsAreAvailable throws exception if the container doesn't have bash


jdelucaa:
**Describe the bug**
When building a container with `Wait.UntilPortsAreAvailable`, I get an expection when starting the container:

Docker.DotNet.DockerApiException: 'Docker API responded with status code=BadRequest, response={"message":"OCI runtime exec failed: exec failed: container_linux.go:346: starting container process caused \"exec: \\\"/bin/bash\\\": stat /bin/bash: no such file or directory\": unknown"}

This is because my container doesn't have `bash` installed, only `sh` or other implementation of shell..

**To Reproduce**
```
var fhirServerContainerBuilder = new TestcontainersBuilder<TestcontainersContainer>()
        .WithImage("conceptant/hapi-fhir")
        .WithPortBinding(8080, true)
        .WithCleanUp(true)
        .WithWaitStrategy(Wait.UntilPortsAreAvailable(8080));

      using (var fhirServerContainer = fhirServerContainerBuilder.Build())
      {
        await fhirServerContainer.StartAsync();
      }
```

**Expected behavior**
The commands used to verify the availability of the ports should support other shells.

**Screenshots**
Not necessary

**Desktop (please complete the following information):**
- Linux container
 - Docker for Windows version 19.03.5, build 633a0ea

**Additional context**
I did the following changes to the code that solved the problem

In `WaitUntilBashCommandsAreCompleted.cs` changed from `/bin/bash` to `bin/sh` - because `sh` will point to the actual implementation of the shell ( bash, dash, etc.. ), I think it'll work for most cases.

The expcetion was gone with that change, but the thread got stucked somewhere, I figured there was something wrong with the command used in `WaitUntilPortsAreAvailable.cs`, for some reason that command doesn't work in my container, it says the directory does not exist, so I changed the line 8 from `timeout 15 bash -c \"echo > /dev/tcp/localhost/8080\"` to a boolean expression with some ways to try and check if the port is available (like the testcontainers-java version does): `true && (cat /proc/net/tcp{{,6}} | awk '{{print $2}}' | grep -i :{port} || nc -vz -w 1 localhost {port} || /bin/bash -c '</dev/tcp/localhost/{port}')`;

Those changes saved my day... let me know if you think it's a good solution to open a PR to fix the issue.

HofmeisterAn:
[...]

PR #179, opened 2020-01-07
Change from bash to sh to execute shell commands

jdelucaa:
- `WaitUntilBashCommandsAreCompleted.cs`: changed from `/bin/bash` to `bin/sh` - `sh` will point to the actual implementation of the shell ( bash, dash, etc.. ).

-  `WaitUntilPortsAreAvailable.cs`: changed from `timeout 15 bash -c \"echo > /dev/tcp/localhost/8080\"` to a boolean expression with some ways to try and check if the port is available (like the testcontainers-java version does): `true && (cat /proc/net/tcp{{,6}} | awk '{{print $2}}' | grep -i :{port} || nc -vz -w 1 localhost {port} || /bin/bash -c '</dev/tcp/localhost/{port}')`;

- Renamed `WaitUntilBashCommandsAreCompleted.cs` to `WaitUntilShellCommandsAreCompleted.cs`

Issue #178


HofmeisterAn (approved):
LGTM! Do you need a new minor release or is the beta version enough? I’ll merge it later that day.

jdelucaa:
Both are fine to me atm, whatever is easier for you :)

PR #193, opened 2020-01-16
Default HostedServiceContainer hostname to DockerEndpoint hostname if possible

PSanetra:
At the moment the default value of a `HostedServiceContainer` hostname is always `localhost`. This does not make sense for Docker endpoints which are accessed via TCP. If the Docker endpoint scheme is `tcp` the default value of the `HostedServiceContainer` should be the hostname of the remote Docker endpoint.

This PR keeps `localhost` as default value if the Docker endpoint scheme is not `tcp`. Otherwise we will use the hostname of the Docker endpoint as default.

This PR makes it more easy to use Testcontainers in GitLab-CI as a pipeline job in GitLab-CI will access a _remote_ Docker daemon in most cases. If I start a Database as Testcontainer then this Database is accessible via the hostname of the remote Docker daemon.

### DIFF
--- a/src/DotNet.Testcontainers/Containers/Configurations/Abstractions/HostedServiceConfiguration.cs
+++ b/src/DotNet.Testcontainers/Containers/Configurations/Abstractions/HostedServiceConfiguration.cs
@@ -26,8 +26,6 @@ namespace DotNet.Testcontainers.Containers.Configurations.Abstractions
 
     public int Port { get; set; }
 
-    public virtual string Hostname { get; set; } = "localhost";
-
     public virtual string Username { get; set; }
 
     public virtual string Password { get; set; }
--- a/src/DotNet.Testcontainers/Containers/IDockerContainer.cs
+++ b/src/DotNet.Testcontainers/Containers/IDockerContainer.cs
@@ -32,6 +32,12 @@ namespace DotNet.Testcontainers.Containers
     [NotNull]
     string MacAddress { get; }
 
+    /// <summary>Gets the Testcontainer hostname.</summary>
+    /// <value>Returns the Docker container hostname if present or an empty string instead.</value>
+    /// <exception cref="InvalidOperationException">If container was not created.</exception>
+    [NotNull]
+    string Hostname { get; }
+
     /// <summary>
     /// Gets the public host port associated with the private container port.
     /// </summary>
--- a/src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
+++ b/src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
@@ -65,6 +65,9 @@ namespace DotNet.Testcontainers.Containers.Modules
       }
     }
 
+    /// <inheritdoc />
+    public string Hostname { get; }
+
     private TestcontainersState State
     {
       get
@@ -84,6 +87,10 @@ namespace DotNet.Testcontainers.Containers.Modules
     {
       this.client = new TestcontainersClient(configuration.Endpoint);
       this.configuration = configuration;
+
+      this.Hostname = "tcp" != this.configuration.Endpoint.Scheme
+        ? "localhost"
+        : this.configuration.Endpoint.Host;
     }
 
     public ushort GetMappedPublicPort(int privatePort)
--- a/src/DotNet.Testcontainers.Tests/Unit/Containers/Linux/TestcontainersContainerTest.cs
+++ b/src/DotNet.Testcontainers.Tests/Unit/Containers/Linux/TestcontainersContainerTest.cs
@@ -258,6 +258,38 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Linux
         }
       }
 
+      [Theory]
+      [InlineData("npipe://./pipe/docker_engine")]
+      [InlineData("unix:/var/run/docker.sock")]
+      public async Task HostnameShouldDefaultToLocalhostIfDockerEndpointSchemeIsNotTcp(string endpoint)
+      {
+        // Given
+        var testcontainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
+          .WithDockerEndpoint(endpoint);
+
+        // When
+        // Then
+        await using (var testcontainer = testcontainersBuilder.Build())
+        {
+          Assert.Equal("localhost", testcontainer.Hostname);
+        }
+      }
+
+      [Fact]
+      public async Task HostnameShouldDefaultToDockerEndpointHostnameIfDockerEndpointSchemeIsTcp()
+      {
+        // Given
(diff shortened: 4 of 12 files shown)


HofmeisterAn:
Thanks for the PR. Yes, this makes no sense 😆. Here are my thoughts:

1. Wouldn't it be easier to replace `this.Hostname` with `this.IpAddress` in each `TestcontainerDatabase` and `TestcontainerMessageBroker` implementation and remove `this.Hostname` entirely?

2. Or remove `container.Hostname = configuration.Hostname` in the `TestcontainersBuilderDatabaseExtension` and `TestcontainersBuilderMessageBrokerExtension` and set `Hostname` in `HostedServiceContainer` via `"tcp".Equals(configuration.Endpoint.Scheme) ? configuration.Endpoint.Host : "localhost"` (remove the setter in this case).

3. Or simply delegate `this.Hostname` in `HostedServiceContainer` to `this.IpAddress` and remove the `Hostname` configuration.

I would like to avoid the virtual member call in the `HostedServiceContainer` constructor.

It would be also nice if you could use CamelCase instead of Snake_Case in the unit tests.

HofmeisterAn (commented):
Thanks for the PR. Yes, this makes no sense 😆. Here are my thoughts:

1. Wouldn't it be easier to replace `this.Hostname` with `this.IpAddress` in each `TestcontainerDatabase` and `TestcontainerMessageBroker` implementation and remove `this.Hostname` entirely?

2. Or remove `container.Hostname = configuration.Hostname` in the `TestcontainersBuilderDatabaseExtension` and `TestcontainersBuilderMessageBrokerExtension` and set `Hostname` in `HostedServiceContainer` via `"tcp".Equals(configuration.Endpoint.Scheme) ? configuration.Endpoint.Host : "localhost"` (remove the setter in this case).

3. Or simply delegate `this.Hostname` in `HostedServiceContainer` to `this.IpAddress` and remove the `Hostname` configuration.

I would like to avoid the virtual member call in the `HostedServiceContainer` constructor and the internal method. It would be also nice if you could use CamelCase instead of Snake_Case in the unit tests.

Thanks again!

PSanetra:
Regarding using the ip address instead of the hostname: I'm not sure if this would lead to problems with TLS certificates which are only valid for specific hostnames. I guess it would. :confused:

I will take another look on your other suggestions.

HofmeisterAn:
> Regarding using the ip address instead of the hostname: I'm not sure if this would lead to problems with TLS certificates which are only valid for specific hostnames. I guess it would. 😕

Good point! Suggestion 2.  would be a suitable. I don't think we need `Hostname` in the configuration at all. Let me have another look as well.

Maybe another property in `TestcontainersContainer` would be a good idea?

HofmeisterAn:
@PSanetra What about something like:

```csharp
public Uri Endpoint
{
  get
  {
    return DockerApiEndpoint.Local.Equals(this.configuration.Endpoint) ? new Uri("tcp://localhost") : this.configuration.Endpoint;
  }
}
```

in `TestcontainersContainer` as implementation of `IDockerContainer`. I'm just a bit unhappy with the name. Furthermore, I am not sure if we should return the Uri or the name.

src/DotNet.Testcontainers/Containers/IDockerContainer.cs
```diff
@@ -32,6 +32,12 @@ public interface IDockerContainer : IAsyncDisposable
     [NotNull]
     string MacAddress { get; }
 
+    /// <summary>Gets the Testcontainer hostname.</summary>
+    /// <value>Returns the Docker container hostname if present or an empty string instead.</value>
+    /// <exception cref="InvalidOperationException">If container was not created.</exception>
+    [NotNull]
+    string Hostname { get; }
```
PSanetra, line 39: Instead of introducing an `Endpoint` property I have moved the `Hostname` from the `HostedServiceContainer` class to this interface. I think only providing the hostname instead of an uri makes more sense since we don't know which protocol and port is useful for the user of this property.

src/DotNet.Testcontainers/Containers/Configurations/Abstractions/HostedServiceConfiguration.cs
```diff
@@ -26,8 +26,6 @@ protected HostedServiceConfiguration(string image, int defaultPort, int port)
 
     public int Port { get; set; }
 
-    public virtual string Hostname { get; set; } = "localhost";
```
PSanetra, line 29: The removal of this configuration property is a breaking change, so we might need to increment the major version of this library.

src/DotNet.Testcontainers.Tests/Unit/Containers/Linux/Database/PostgreSqlTestcontainerTest.cs
```diff
@@ -1,5 +1,6 @@
 namespace DotNet.Testcontainers.Tests.Unit.Containers.Linux.Database
 {
+  using System.Threading;
```
HofmeisterAn: Remove unnecessary using.

src/DotNet.Testcontainers/Containers/Configurations/Abstractions/HostedServiceConfiguration.cs
HofmeisterAn: You're right about that. But I don't think the property was actually used and I don't care about breaking changes. Too much work for such a small project.

src/DotNet.Testcontainers/Containers/Modules/Abstractions/TestcontainerDatabase.cs
```diff
@@ -1,5 +1,6 @@
 namespace DotNet.Testcontainers.Containers.Modules.Abstractions
 {
+  using Configurations.Abstractions;
```
HofmeisterAn: Remove unnecessary using.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -84,6 +87,10 @@ protected TestcontainersContainer(ITestcontainersConfiguration configuration)
     {
       this.client = new TestcontainersClient(configuration.Endpoint);
       this.configuration = configuration;
+
+      this.Hostname = "tcp" != this.configuration.Endpoint?.Scheme
```
HofmeisterAn: `Endpoint` is not nullable.

HofmeisterAn (changes requested):
It would be nice if you could rebase on develop and maybe consider [CONTRIBUTING.md](https://github.com/HofmeisterAn/dotnet-testcontainers/blob/develop/CONTRIBUTING.md#labels) for commit messages. Then I'll merge it later.

PSanetra:
@HofmeisterAn Done

PR #214, opened 2020-02-27
Support testcontainers to run side of container

### LINKED ISSUE #213: Support testcontainers to run inside a docker container with mounted docker.socket
testcontainers-java supports running testcontainers inside a containers and access side spawned containers

Instead of using localhost(127.0.0.1), testcontainers should use the gateway ip container if it is running inside a container: https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/TestContainers/Core/Containers/Container.cs#L195-L212

farodin91:
Fixes #213

### DIFF
--- a/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
+++ b/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
@@ -261,36 +261,28 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
       }
 
       [Theory]
-      [InlineData("npipe://./pipe/docker_engine")]
-      [InlineData("unix:/var/run/docker.sock")]
-      public async Task HostnameShouldDefaultToLocalhostIfDockerEndpointSchemeIsNotTcp(string endpoint)
+      [InlineData("localhost", "npipe://./pipe/docker_engine")]
+      [InlineData("localhost", "unix:/var/run/docker.sock")]
+      [InlineData("docker", "http://docker")]
+      [InlineData("docker", "https://docker")]
+      [InlineData("docker", "tcp://docker")]
+      [InlineData("1.1.1.1", "http://1.1.1.1")]
+      [InlineData("1.1.1.1", "https://1.1.1.1")]
+      [InlineData("1.1.1.1", "tcp://1.1.1.1")]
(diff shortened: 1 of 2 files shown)


src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -196,6 +197,7 @@ public TDockerContainer Build()
 
     private static ITestcontainersConfiguration Apply(
       Uri endpoint = null,
+      string gateway = null,
```
HofmeisterAn: Do we really need an additional parameter `gateway`? Couldn't we just use `endpoint`? I think that's much simpler.

HofmeisterAn (commented):
I would prefere to reuse the `endpoint . Everything else looks good, thanks.

~~We could call `GetDefaultGateway` from the `Hostnamen` property or add an `Endpoint` property which returns the original endpoint and rename `Hostnamen` to `Gateway`~~.

What do you think about this:

```csharp
// TestcontainersContainer.cs
public string Hostname
{
  get
  {
    switch (this.configuration.Endpoint.Scheme)
    {
      case "unix":
      case "npipe":
        return "localhost";
      case "tcp":
      case "http":
      case "https":
        return this.configuration.Endpoint.Host;
      default:
        this.ThrowIfContainerHasNotBeenCreated();
        return this.container.NetworkSettings.Networks.First().Value.Gateway;
    }
  }
}
```

```csharp
// TestcontainersContainerTest.cs
[Theory]
[InlineData("localhost", "npipe://./pipe/docker_engine")]
[InlineData("localhost", "unix:/var/run/docker.sock")]
[InlineData("docker", "http://docker")]
[InlineData("docker", "https://docker")]
[InlineData("docker", "tcp://docker")]
[InlineData("1.1.1.1", "http://1.1.1.1")]
[InlineData("1.1.1.1", "https://1.1.1.1")]
[InlineData("1.1.1.1", "tcp://1.1.1.1")]
public async Task HostnameShouldMatchDockerGatewayAddress(string expectedHostname, string endpoint)
{
  // Given
  var testcontainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
    .WithDockerEndpoint(endpoint);

  // When
  // Then
  await using (var testcontainer = testcontainersBuilder.Build())
  {
    Assert.Equal(expectedHostname, testcontainer.Hostname);
  }
}
```

farodin91:
@HofmeisterAn I updated the hostname method.

PR #227, opened 2020-03-25
Feature/evd anton feature oracle testcontainer

HofmeisterAn:


EvdAnton:
there are only one thing, you can change command in wait strategy like 
`echo "exit" | $ORACLE_HOME/bin/sqlplus -s -L system/oracle@localhost:1521/XE`

Issue #237, opened 2020-04-23
ImageFromDockerfileBuilder does not tag image 


Jejuni:
**Describe the bug**
I have a Dockerfile which consists only of the following:
```
FROM rabbitmq:3.8-management-alpine

COPY rabbitmq_delayed_message_exchange* /plugins/

RUN rabbitmq-plugins enable rabbitmq_delayed_message_exchange
```

I have the delayed exchange plugin in the same folder as the dockerfile.

Using 
```
var imgBuilder = new ImageFromDockerfileBuilder()
                .WithName("rabbitmqtest:custom")
                .WithDockerfileDirectory("docker/rabbitmq");
string img = await imgBuilder.Build();
```
I could not get the image to be tagged. It would always be none:none.
I noticed that when debugging and stopping in 
```
using (var unused = await this.Docker.Images.BuildImageFromDockerfileAsync(stream, new ImageBuildParameters { Dockerfile = config.Dockerfile, Tags = new[] {config.Image.FullName } }, ct))
{
    // New Docker image built, ready to use.
}
```
and remaining there the image would be tagged.
It seemed to me that creation and tagging are two separate processes in `BuildImageFromDockerfileAsync` and if the stream is disposed to quickly the image will not get tagged. 
That also explains why for very simple Dockerfile the tagging will succeed.

**To Reproduce**
Execute the code mentioned above in a unit test.

**Expected behavior**
The image should be tagged.

**Solution**
Reading the stream to the end delays further execution until the image is tagged.
I have provided a PR #238 but your build pipeline seems to be failing during an initial step unrelated to the PR.

HofmeisterAn:
Included in #238 and published in `1.3.0-beta.20200423.2`. If you need a proper release let me know.

Jejuni:
The beta release will be fine.
Thanks!

PR #238, opened 2020-04-23
fix: Waiting for docker to process image building

Jejuni:
Added stream reading to the returned stream to delay dispose until docker is done with the image.

See issue #237


HofmeisterAn:
👍 Thanks

Issue #258, opened 2020-07-03
Add Couchbase Testcontainer


mustafaonuraydin:
Couchbase Server is an open source, distributed, JSON document database. In dotnet-testcontainers project there is some pre-configured containers for easy to use. 

The issue is add couchbase testcontainer as pre-configured container.

This container should provide following features;
1. Create initial cluster and create customizable admin user with environment parameter
2. Create initial bucket or buckets
3. Add startup strategy for Couchbase
4. Create bucket function
5. Flush bucket function

HofmeisterAn:
We should merge #261. Based on the existing example configurations you can create the configuration for Couchbase. This should be pretty straight forward. It would be nice if you keep the other configurations in mind and share patterns, use interfaces, add tests, etc.

HofmeisterAn:
Closed in #296, #298 and published in `1.4.0-beta.20200721.12`.

PR #293, opened 2020-07-19
[#258] #IMPLEMENT 'assemblyName: DotNet.Testcontainers; function: CouchbaseTestcontainer' {Add pre-configured Couchbase Testcontainer.}

mustafaonuraydin:
[#258] #IMPLEMENT 'assemblyName: DotNet.Testcontainers; function: CouchbaseTestcontainer' {Add pre-configured Couchbase Testcontainer.}

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -82,7 +82,7 @@ Task("Tests")
     DotNetCoreTest(testProject.Path.FullPath, new DotNetCoreTestSettings
     {
       Configuration = param.Configuration,
-      Verbosity = param.Verbosity,
+      Verbosity = Cake.Common.Tools.DotNetCore.DotNetCoreVerbosity.Normal,
       NoRestore = true,
       NoBuild = true,
       Logger = "trx",
--- a/src/DotNet.Testcontainers.Tests/DotNet.Testcontainers.Tests.csproj
+++ b/src/DotNet.Testcontainers.Tests/DotNet.Testcontainers.Tests.csproj
@@ -5,6 +5,7 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="CouchbaseNetClient" Version="3.0.1" />
     <PackageReference Include="coverlet.msbuild" Version="2.9.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.1" />
     <PackageReference Include="MyCouch" Version="7.2.1" />
--- a/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
+++ b/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
@@ -1,0 +1,157 @@
+namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix.Database
+{
+  using System;
+  using System.Threading.Tasks;
+  using Couchbase;
+  using Testcontainers.Containers.Builders;
+  using Testcontainers.Containers.Configurations.Databases;
+  using Testcontainers.Containers.Modules.Databases;
+  using Xunit;
+
+  public class CouchbaseTestcontainerTest
+  {
+
+    /// <summary>
+    /// Sample Couchbase document class for test
+    /// </summary>
+    private class Customer
+    {
+      public string Name { get; set; }
+      public int Age{ get; set; }
+    }
+
+    [Fact]
+    public async Task ConnectionEstablished_WithCrudTest()
+    {
+      //Given
+      var couchbaseTestcontainer = new TestcontainersBuilder<CouchbaseTestcontainer>()
+        .WithDatabase(new CouchbaseTestcontainerConfiguration("Administrator", "password", "Customer"))
+        .WithExposedPort(8091)
+        .WithExposedPort(11210)
+        .WithPortBinding(8091, 8091)
+        .WithPortBinding(11210, 11210)
+        .Build();
+
+      await couchbaseTestcontainer.StartAsync();
+
+      //When & Then
+      var cluster = await Cluster.ConnectAsync(couchbaseTestcontainer.ConnectionString, "Administrator","password");
+      var bucket = await cluster.BucketAsync("Customer");
+      var collection = bucket.DefaultCollection();
+
+      // Create
+      await collection.InsertAsync(
+        "customer-id-1",
+        new Customer {Name = "Mustafa", Age = 29}
+        );
+
+      // Read
+      var customer = (await collection.GetAsync("customer-id-1")).ContentAs<Customer>();
+      Assert.True(customer.Name == "Mustafa");
+      Assert.True(customer.Age == 29);
+
+      //Update
+      customer.Age = 30;
+      customer.Name = "Onur";
+      await collection.UpsertAsync("customer-id-1", customer);
+
+      var updatedCustomer = (await collection.GetAsync("customer-id-1")).ContentAs<Customer>();
+      Assert.True(updatedCustomer.Name == "Onur");
+      Assert.True(updatedCustomer.Age == 30);
+
+      //Delete
+      await collection.RemoveAsync("customer-id-1");
+      var existsResult = await collection.ExistsAsync("customer-id-1");
+      Assert.False(existsResult.Exists);
+
+      await couchbaseTestcontainer.StopAsync();
+    }
+
+    [Fact]
+    public void ShouldNotThrowsArgumentOutOfRangeException_WhenClusterConfigurationIsValid()
+    {
+      //Given
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "dotnet-testcontainers");
+
+      //When & Then
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration
+        .ClusterRamSize(1024)
+        .ClusterIndexRamSize(512)
+        .ClusterEventingRamSize(256)
+        .ClusterFtsRamSize(256)
+        .ClusterAnalyticsRamSize(1024);;
+    }
+
+    [Fact]
+    public void ShouldThrowsArgumentOutOfRangeException_WhenClusterRamSizeIsNotValid()
+    {
+      //Given
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "test");
+
+      //When
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration.ClusterRamSize(1023);
+
+      //Then
+      var exception = Assert.Throws<ArgumentOutOfRangeException>(action);
+      Assert.True(exception.Message == "Couchbase clusterRamSize ram size can not be less than 1024 MB. (Parameter 'clusterRamSize')");
+    }
+
+    [Fact]
+    public void ShouldThrowsArgumentOutOfRangeException_WhenClusterAnalyticsRamSizeIsNotValid()
+    {
+      //Given
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "test");
+
+      //When
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration.ClusterAnalyticsRamSize(1023);
+
+      //Then
+      var exception = Assert.Throws<ArgumentOutOfRangeException>(action);
+      Assert.True(exception.Message == "Couchbase clusterAnalyticsRamSize ram size can not be less than 1024 MB. (Parameter 'clusterAnalyticsRamSize')");
+    }
+
+    [Fact]
+    public void ShouldThrowsArgumentOutOfRangeException_WhenClusterEventingRamSizeIsNotValid()
+    {
+      //Given
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "test");
+
+      //When
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration.ClusterEventingRamSize(255);
+
+      //Then
+      var exception = Assert.Throws<ArgumentOutOfRangeException>(action);
+      Assert.True(exception.Message == "Couchbase clusterEventingRamSize ram size can not be less than 256 MB. (Parameter 'clusterEventingRamSize')");
+    }
+
+    [Fact]
+    public void ShouldThrowsArgumentOutOfRangeException_WhenClusterFtsRamSizeIsNotValid()
+    {
+      //Given
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "test");
+
+      //When
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration.ClusterFtsRamSize(255);
+
+      //Then
+      var exception = Assert.Throws<ArgumentOutOfRangeException>(action);
+      Assert.True(exception.Message == "Couchbase clusterFtsRamSize ram size can not be less than 256 MB. (Parameter 'clusterFtsRamSize')");
+    }
+
+    [Fact]
+    public void ShouldThrowsArgumentOutOfRangeException_WhenClusterIndexRamSizeIsNotValid()
+    {
+      //Given
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "test");
+
+      //When
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration.ClusterIndexRamSize(511);
+
+      //Then
+      var exception = Assert.Throws<ArgumentOutOfRangeException>(action);
+      Assert.True(exception.Message == "Couchbase clusterIndexRamSize ram size can not be less than 512 MB. (Parameter 'clusterIndexRamSize')");
+    }
+
+  }
+
+}
--- a/src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
+++ b/src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
@@ -1,0 +1,101 @@
+namespace DotNet.Testcontainers.Containers.Configurations.Databases
+{
+  using System;
+  using System.IO;
+  using Abstractions;
+  using OutputConsumers;
+  using WaitStrategies;
+
+  public sealed class CouchbaseTestcontainerConfiguration : TestcontainerDatabaseConfiguration
+  {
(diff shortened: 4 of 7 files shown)


build.cake
```diff
@@ -82,7 +82,7 @@ Task("Tests")
     DotNetCoreTest(testProject.Path.FullPath, new DotNetCoreTestSettings
     {
       Configuration = param.Configuration,
-      Verbosity = param.Verbosity,
```
HofmeisterAn, line 85: Undo change.

src/DotNet.Testcontainers.Tests/DotNet.Testcontainers.Tests.csproj
```diff
@@ -5,6 +5,7 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="CouchbaseNetClient" Version="3.0.1" />
```
HofmeisterAn, line 8: Latest version `3.0.3`.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+    {
+      public string Name { get; set; }
+      public int Age{ get; set; }
+    }
+
+    [Fact]
+    public async Task ConnectionEstablished_WithCrudTest()
+    {
+      //Given
+      var couchbaseTestcontainer = new TestcontainersBuilder<CouchbaseTestcontainer>()
```
HofmeisterAn, line 27: Add `using` statement. Maybe also consider disposing `cluster` and `bucket`. Or better use the `OracleFixture` approach with `IAsyncLifetime`.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+
+      // Create
+      await collection.InsertAsync(
+        "customer-id-1",
+        new Customer {Name = "Mustafa", Age = 29}
+        );
+
+      // Read
+      var customer = (await collection.GetAsync("customer-id-1")).ContentAs<Customer>();
+      Assert.True(customer.Name == "Mustafa");
```
HofmeisterAn, line 50: Use `Assert.Equal("Mustafa", customer.Name);` instead. Use this for the other assertions too please.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    [Fact]
+    public void ShouldNotThrowsArgumentOutOfRangeException_WhenClusterConfigurationIsValid()
+    {
+      //Given
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "dotnet-testcontainers");
+
+      //When & Then
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration
```
HofmeisterAn, line 77: You don't invoke the action and check the results here. I think you should at least call the action. If we switch to getters and setters we can even test the result. Would be nice.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+      var configuration = new CouchbaseTestcontainerConfiguration("Administrator", "password", "test");
+
+      //When
+      Func<CouchbaseTestcontainerConfiguration> action = () => configuration.ClusterIndexRamSize(511);
+
+      //Then
+      var exception = Assert.Throws<ArgumentOutOfRangeException>(action);
+      Assert.True(exception.Message == "Couchbase clusterIndexRamSize ram size can not be less than 512 MB. (Parameter 'clusterIndexRamSize')");
+    }
+
```
HofmeisterAn, line 154: Remove empty lines.

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+  using System;
+  using System.IO;
+  using Abstractions;
+  using OutputConsumers;
+  using WaitStrategies;
+
+  public sealed class CouchbaseTestcontainerConfiguration : TestcontainerDatabaseConfiguration
+  {
+    internal const string CouchbaseImage = "mustafaonuraydin/couchbase-testcontainer:6.0.0";
```
HofmeisterAn, line 11: Private modifier; why can we not use the official docket image?

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    private const int DefaultClusterRamSize = 1024;
+    private const int DefaultClusterIndexRamSize = 512;
+    private const int DefaultClusterEventingRamSize = 256;
+    private const int DefaultClusterFtsRamSize = 256;
+    private const int DefaultClusterAnalyticsRamSize = 1024;
+
+    private const int BootstrapHttpPort = 8091;
+
+    private Stream OutStream = new MemoryStream();
```
HofmeisterAn, line 24: `OutStream` and `ErrorStream` readonly. Maybe rename it to stdout and stderr.

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,101 @@
...
+      this.Username = username;
+      this.Password = password;
+      this.Port = BootstrapHttpPort;
+      this.Environments["ADMIN_NAME"] = username;
+      this.Environments["PASSWORD"] = password;
+      this.Environments["INITIAL_BUCKET_NAME"] = bucket;
+      this.Environments["INITIAL_BUCKET_PASSWORD"] = password;
+      this.Environments["INITIAL_BUCKET_TYPE"] = bucketType;
+      this.Environments["INITIAL_BUCKET_RAMSIZE"] = bucketRamSize.ToString();
+      this.InitializeDefaults();
```
HofmeisterAn, line 44: Can we call a private constructor instead?

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public CouchbaseTestcontainerConfiguration ClusterRamSize(int clusterRamSize)
```
HofmeisterAn, line 47: What do you think about getters and setters instead? Like `MySqlTestcontainerConfiguration`? This would keep the library more consistent. I would really appreciate that change.

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,101 @@
...
+      ValidateMinimumRequirement(nameof(clusterAnalyticsRamSize),clusterAnalyticsRamSize,DefaultClusterAnalyticsRamSize);
+      this.Environments["CLUSTER_EVENTING_RAMSIZE"] = clusterAnalyticsRamSize.ToString();
+      return this;
+    }
+
+    private static void ValidateMinimumRequirement(string name, int memory, int minimum)
+    {
+      if (memory < minimum)
+      {
+        throw new ArgumentOutOfRangeException(name,"Couchbase "+name+" ram size can not be less than "+minimum+" MB.");
```
HofmeisterAn, line 86: String interpolation would be nice.

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
```diff
@@ -0,0 +1,39 @@
...
+  {
+    internal CouchbaseTestcontainer(ITestcontainersConfiguration configuration) : base(configuration)
+    {
+    }
+
+    public override string ConnectionString => "couchbase://" + this.Hostname;
+
+    public async Task<long> CreateBucket(string bucket, int memory = 512, int withDelaySecond = 3)
+    {
+      var createBucketCommand =
```
HofmeisterAn, line 18: - String interpolation (`FlushBucket` too):
`$"/opt/couchbase/bin/couchbase-cli bucket-create -c 127.0.0.1:8091 --username {this.Username} --password {this.Password} --bucket={bucket} --bucket-type couchbase --bucket-ramsize {memory} --enable-flush 1 --bucket-replica 0";`

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
```diff
@@ -0,0 +1,39 @@
...
+        "bash /opt/couchbase/bin/couchbase-cli " +
+        "bucket-create -c 127.0.0.1:8091 " +
+        "--username " + this.Username + " --password " + this.Password + " " +
+        "--bucket=" + bucket + " --bucket-type couchbase --bucket-ramsize "+memory+" --enable-flush 1 --bucket-replica 0";
+      await Task.Delay(withDelaySecond * 1000);
```
HofmeisterAn, line 23: Why do we need this delay? I think we should remove it (`FlushBucket` too).

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
```diff
@@ -0,0 +1,39 @@
...
+      return await this.ExecAsync(createBucketCommand.Split(" ").ToList());
```
HofmeisterAn, line 24: This should `return this.ExecAsync(new[] { "/bin/sh", "-c", createBucketCommand });` (`FlushBucket` too).

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
```diff
@@ -0,0 +1,39 @@
...
+
+  public class CouchbaseTestcontainer : TestcontainerDatabase
+  {
+    internal CouchbaseTestcontainer(ITestcontainersConfiguration configuration) : base(configuration)
+    {
+    }
+
+    public override string ConnectionString => "couchbase://" + this.Hostname;
+
+    public async Task<long> CreateBucket(string bucket, int memory = 512, int withDelaySecond = 3)
```
HofmeisterAn, line 16: `FlushBucket` and `FlushBucket` not covered by tests. Would be nice to have tests here. Maybe consider xUnit `IAsyncLifetime`, example in `OracleFixture`. I think that's a good way to set-up containers for tests.

HofmeisterAn (changes requested):
Thanks for your PR! It would be nice if you rename your branch to `feature/...`. `develop` causes "build errors".
I think the requested changes can be implemented quite quickly. Not a big deal for you.

I really like to use getters and setters in the configuration classes. This change would be really nice in `CouchbaseTestcontainerConfiguration` for consistency.

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+
+    private void InitializeDefaults()
+    {
+      this.Environments["CLUSTER_RAMSIZE"] = DefaultClusterRamSize.ToString();
```
HofmeisterAn, line 92: This should call the equivalent setter, including others.

src/DotNet.Testcontainers.Tests/DotNet.Testcontainers.Tests.csproj
mustafaonuraydin: I am not sure the latest version is stable especially about N1QL calls. But latest version can cover to test Couchbase Testcontainer. I will update.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
mustafaonuraydin: IAsyncLifetime looks like more better solution for manage. I'll update.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
mustafaonuraydin: Sorry about that. I'm not familiar to use Assert functions over C# because I'm using FluentAssertions. I'll update.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
mustafaonuraydin: You are absolutely right! I will reconsider these tests after update configuration class with your suggestion. Also, I am looking something like Assert.NotThrows , do you have any advice?

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
mustafaonuraydin: Unfortunately the official image is not providing initialize cluster definition, create user, create bucket etc. On first use you should configure Couchbase over web ui or couchbase-cli. My image is using official image and running installation script over official image with necessary installment script.
My offer is, let's open docker id  like "dotnet-testcontainers" or "testcontainers". After that we can push this image to that repository. And also, we can create small git project containing files which is i wrote Dockerfile and entrypoint.sh . After that this image will be open to contribute and source code will be depended to dotnet-testcontainers project. Or adding docker image build files into dotnet-testcontainers project is also good.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
HofmeisterAn: I think there is:

```
var ex = Record.Exception(() => // Code that does not throw);
Assert.Null(ex);
```

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
mustafaonuraydin: We can remove that initializing after adding docker image sources into project. Default variables will be already set to defaults.

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
HofmeisterAn: Maybe we can add a comment to the image what it contains in addition. If you like you can also set-up your own GitHub Repository for it.

src/DotNet.Testcontainers/Containers/Configurations/Databases/CouchbaseTestcontainerConfiguration.cs
mustafaonuraydin: I will update these functions with your advice. Do you have any suggestion to validate function over setters?

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
mustafaonuraydin: Execution to these commands over Couchbase taking a bit time because of Couchbase index mechanism. It's totally depending to resources. In local environment you can see results without any delay. But depending CI/CI pipeline environment resources the delay may need to use. On the other hand, this is not a responsibility of library. Library user can consider this situation depending ci/ci pipeline. What is your suggestion? Should we remove?

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
mustafaonuraydin: I'll add new tests! Thank you.

mustafaonuraydin:
@HofmeisterAn thanks for your review. I'm so familiar trunk based development. That's the reason of developing without feature branch. But i agree to you in this pr will be more efficient to create feature branch. I will send new pr with new feature branch which is updated with your suggestions.  Many thanks for your time!

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
HofmeisterAn: Well I think that call should be executed after `this.ExecAsync()` (if at all) and then we should use `Task.DelayTimeSpan.FromSeconds(withDelaySecond))`.

However, getting rid of it would be the best. Can we check with `couchbase-cli` if the bucket exist? Then we can use the behaviour of the wait strategies. Such as:

```csharp
public Task<bool> DoesBucketExists(string bucketName, CancellationToken ct = default)
{
  return Task.FromResult(true); // Add couchbase-cli call here and check exit code or something like that.
}

public async Task<long> CreateBucket(string bucketName, int memory = 512, int timeout = 1000, CancellationToken ct = default)
{
  const string createBucketCommand = "couchbase-cli create bucket command";
  var exitCode = await this.ExecAsync(new[] { "/bin/sh", "-c", createBucketCommand }, ct);

  await WaitStrategy.WaitUntil(() => this.DoesBucketExists(bucketName, ct), timeout: timeout, ct: ct);
  return exitCode;
}
```

PR #298, opened 2020-07-21
[#258] #IMPLEMENT 'assemblyName: DotNet.Testcontainers; function: CouchbaseTestcontainer'

HofmeisterAn:
{Minor code improvments.}

### DIFF
--- a/src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/ModuleFixture.cs
+++ b/src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/ModuleFixture.cs
@@ -1,0 +1,14 @@
+namespace DotNet.Testcontainers.Tests.Fixtures.Containers.Modules
+{
+  using DotNet.Testcontainers.Containers.Modules.Abstractions;
+
+  public class ModuleFixture<T> where T : HostedServiceContainer
+  {
+    public ModuleFixture(T container)
+    {
+      this.Container = container;
+    }
+
+    public T Container { get; }
+  }
+}
--- a/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerConfigurationTest.cs
+++ b/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerConfigurationTest.cs
@@ -1,93 +1,115 @@
 namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix.Database
 {
   using System;
-  using Testcontainers.Containers.Configurations.Databases;
+  using DotNet.Testcontainers.Containers.Configurations.Databases;
   using Xunit;
 
   public class CouchbaseTestcontainerConfigurationTest
   {
     [Fact]
-    public void ShouldNotThrowsArgumentOutOfRangeException_WhenClusterConfigurationIsValid()
+    public void ShouldNotThrowArgumentOutOfRangeExceptionWhenClusterConfigurationIsValid()
     {
-      //Given
-      Action action = () => new CouchbaseTestcontainerConfiguration
+      // Given
+      CouchbaseTestcontainerConfiguration configuration = null;
+
+      Action action = () => configuration =  new CouchbaseTestcontainerConfiguration
       {
+        BucketName = "Bucket",
+        BucketType = "MEMCACHED",
         ClusterRamSize = "1024",
         ClusterIndexRamSize = "512",
         ClusterEventingRamSize = "256",
         ClusterFtsRamSize = "256",
         ClusterAnalyticsRamSize = "1024"
       };
-      //When
-      var exception = Record.Exception(action);
-      //Then
+
+      // When
+      var exception = Record.Exception(action.Invoke);
+
+      // Then
       Assert.Null(exception);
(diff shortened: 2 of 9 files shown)


src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
```diff
@@ -14,19 +16,14 @@ internal CouchbaseTestcontainer(ITestcontainersConfiguration configuration) : ba
 
     public async Task<long> CreateNewBucket(string bucket, int memory = 128)
     {
-      var createBucketCommand =
-        "bash /opt/couchbase/bin/couchbase-cli bucket-create -c 127.0.0.1:8091 " +
-        $"--username {this.Username} --password {this.Password} " +
-        $"--bucket={bucket} --bucket-type couchbase --bucket-ramsize {memory.ToString()} --enable-flush 1 --bucket-replica 0";
-      return await this.ExecAsync(new[] {"/bin/sh", "-c", createBucketCommand});
+      var createBucketCommand = $"{couchbaseCli} bucket-create -c 127.0.0.1:8091 --username {this.Username} --password {this.Password} --bucket={bucket} --bucket-type couchbase --bucket-ramsize {memory} --enable-flush 1 --bucket-replica 0";
```
mustafaonuraydin, line 19: @HofmeisterAn what do you think to pass --bucket-type with a parameter which has default value "couchbase" like following;
CreateNewBucket(string bucket, int memory = 128,string bucketType=couchbase)

src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/ModuleFixture.cs
```diff
@@ -0,0 +1,14 @@
+namespace DotNet.Testcontainers.Tests.Fixtures.Containers.Modules
+{
+  using DotNet.Testcontainers.Containers.Modules.Abstractions;
+
+  public class ModuleFixture<T> where T : HostedServiceContainer
```
mustafaonuraydin, line 5: I really like it. Common approach for Couchbase and Oracle.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerConfigurationTest.cs
```diff
@@ -1,93 +1,115 @@
...
     {
-      //Given
-      Action action = () => new CouchbaseTestcontainerConfiguration
+      // Given
+      CouchbaseTestcontainerConfiguration configuration = null;
+
+      Action action = () => configuration =  new CouchbaseTestcontainerConfiguration
       {
+        BucketName = "Bucket",
+        BucketType = "MEMCACHED",
```
mustafaonuraydin, line 18: BucketName and BucketType has already default values and not required to set. These parameters is not validating. Also regarding Couchbase document, accepted bucket types are "couchbase", "ephemeral", and "memcached". The "couchbase" bucket is the standard bucket type. I'm not sure uppercase is valid.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
```diff
@@ -1,106 +1,118 @@
...
-        "1",
-        new { Phone = "123456" }
-      );
+          // When
+          await this.couchbaseFixture.Container.FlushBucket(bucketName);
 
-      //When
-      await this.couchbase.Container.FlushBucket("Example");
+          // Then
+          var result = await cluster.QueryAsync<long>($"SELECT * FROM {bucketName}"); // Don't do this.
```
mustafaonuraydin, line 105: It's possible to execute count query. Or maybe better to fetch only document which has id "1". Do you have any suggestion?

mustafaonuraydin (commented):
I add some comments. Not critical but can be reviewable.

src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
HofmeisterAn: I was also thinking about that. We can changes this with a new feature.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerConfigurationTest.cs
HofmeisterAn: Thanks 👍. I added this to test the getters. We can improve this with further prs. Maybe with an enum or something like that.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Database/CouchbaseTestcontainerTest.cs
HofmeisterAn: I don't know a lot about couchbase. But it looks like a regular SQL query. I'm pretty sure this will work. Should be some improvement for the next issue / pr.

PR #355, opened 2021-01-19
Add KafkaTestcontainer

PSanetra:
This PR adds a `KafkaTestcontainer`. There are a lot of changes, because the container needs to be implemented kind of hacky. Kafka needs to know on which external network interface it is reachable. Therefore it is needed to start the container before configuring Kafka via environment variables. The KafkaTestcontainer will wait after startup for a startup script to be created inside the Container. The startup script will contain the randomly assigned host port and will start the Kafka instance.

This approach is very similar to the approach, taken by the Java Testcontainers project.

This PR also adds the possibility to add a file to a `TestcontainersContainer` and fixes a possible source for issues by letting the Docker daemon decide on random port assignments instead of the client in seperate commits.

### DIFF
--- a/src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/AlpineFixture.cs
+++ b/src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/AlpineFixture.cs
@@ -14,9 +14,10 @@ namespace DotNet.Testcontainers.Tests.Fixtures.Containers.Modules
     {
     }
 
-    public Task InitializeAsync()
+    public async Task InitializeAsync()
     {
-      return Task.WhenAll(this.Container.StartAsync(), this.Container.StopAsync());
+      await this.Container.StartAsync().ConfigureAwait(false);
+      await this.Container.StopAsync().ConfigureAwait(false);
     }
 
     public Task DisposeAsync()
--- a/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
+++ b/src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
@@ -1,8 +1,10 @@
 namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
 {
   using System;
+  using System.Collections.Generic;
   using System.IO;
   using System.Net;
+  using System.Text;
   using System.Threading.Tasks;
   using DotNet.Testcontainers.Clients;
   using DotNet.Testcontainers.Containers;
@@ -355,6 +357,29 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
         {
           await testcontainer.StartAsync();
           Assert.Equal(255, await testcontainer.ExecAsync(new[] { "/bin/sh", "-c", "exit 255" }));
+        }
+      }
+
+      [Fact]
+      public async Task CopyFileToRunningContainer()
+      {
+        // Given
+        const string dayOfWeekFilePath = "/tmp/dayOfWeek";
+
+        var dayOfWeek = DateTime.Now.DayOfWeek.ToString();
+
+        var testcontainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
+          .WithImage("alpine")
+          .WithCommand("/bin/sh", "-c", "tail -f /dev/null");
+
+        // When
+        // Then
+        await using (IDockerContainer testcontainer = testcontainersBuilder.Build())
+        {
+          await testcontainer.StartAsync();
+          await testcontainer.CopyFileAsync(dayOfWeekFilePath, Encoding.UTF8.GetBytes(dayOfWeek));
+          Assert.Equal(0, await testcontainer.ExecAsync(new[] { "/bin/sh", "-c", $"test \"$(cat {dayOfWeekFilePath})\" = \"{dayOfWeek}\"" }));
+          Assert.Equal(0, await testcontainer.ExecAsync(new[] { "/bin/sh", "-c", $"stat {dayOfWeekFilePath} | grep 0600" }));
         }
       }
     }
--- a/src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
+++ b/src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
@@ -119,6 +119,21 @@ namespace DotNet.Testcontainers.Clients
       return long.MinValue;
     }
 
+    public Task ExtractArchiveToContainerAsync(string id, string path, Stream tarStream, CancellationToken ct = default)
+    {
+      Logger.LogInformation("Copying tar stream to {path} at container {id}", path, id);
+
+      return this.Docker.Containers.ExtractArchiveToContainerAsync(
+        id,
+        new ContainerPathStatParameters()
+        {
+          Path = path,
+          AllowOverwriteDirWithFile = false
+        },
+        tarStream,
+        ct);
+    }
+
     public async Task<string> RunAsync(ITestcontainersConfiguration configuration, CancellationToken ct = default)
     {
       var converter = new TestcontainersConfigurationConverter(configuration);
--- a/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
+++ b/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
@@ -4,6 +4,7 @@ namespace DotNet.Testcontainers.Clients
   using System.Collections.Generic;
   using System.Diagnostics;
   using System.IO;
+  using System.Text;
   using System.Threading;
   using System.Threading.Tasks;
   using Docker.DotNet.Models;
@@ -11,6 +12,7 @@ namespace DotNet.Testcontainers.Clients
   using DotNet.Testcontainers.Containers.OutputConsumers;
   using DotNet.Testcontainers.Images.Configurations;
   using DotNet.Testcontainers.Services;
+  using ICSharpCode.SharpZipLib.Tar;
 
   internal sealed class TestcontainersClient : ITestcontainersClient
   {
@@ -130,6 +132,34 @@ namespace DotNet.Testcontainers.Clients
     public Task<long> ExecAsync(string id, IList<string> command, CancellationToken ct = default)
     {
       return this.containers.ExecAsync(id, command, ct);
+    }
+
+    public async Task CopyFileAsync(string id, string filePath, byte[] fileContent, int accessMode, int userId, int groupId, CancellationToken ct = default)
+    {
+      await using var memStream = new MemoryStream();
+      await using var tarOutputStream = new TarOutputStream(memStream, Encoding.Default)
+      {
+        IsStreamOwner = false
+      };
+
+      tarOutputStream.PutNextEntry(new TarEntry(new TarHeader()
+      {
+        Name = filePath,
+        UserId = userId,
+        GroupId = groupId,
+        Mode = accessMode,
+        Size = fileContent.Length
+      }));
+
+      await tarOutputStream.WriteAsync(fileContent, ct);
+
+      tarOutputStream.CloseEntry();
+      tarOutputStream.Close();
+      await tarOutputStream.FlushAsync(ct);
+
+      memStream.Position = 0;
+
+      await this.containers.ExtractArchiveToContainerAsync(id, "/", memStream, ct);
     }
 
     public async Task<string> RunAsync(ITestcontainersConfiguration configuration, CancellationToken ct = default)
--- a/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
+++ b/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
@@ -151,15 +151,18 @@ namespace DotNet.Testcontainers.Clients
         Size = fileContent.Length
       }));
 
-      await tarOutputStream.WriteAsync(fileContent, ct);
+      await tarOutputStream.WriteAsync(fileContent, ct)
+        .ConfigureAwait(false);
 
       tarOutputStream.CloseEntry();
       tarOutputStream.Close();
-      await tarOutputStream.FlushAsync(ct);
+      await tarOutputStream.FlushAsync(ct)
+        .ConfigureAwait(false);
 
       memStream.Position = 0;
 
-      await this.containers.ExtractArchiveToContainerAsync(id, "/", memStream, ct);
+      await this.containers.ExtractArchiveToContainerAsync(id, "/", memStream, ct)
+        .ConfigureAwait(false);
     }
 
     public async Task<string> RunAsync(ITestcontainersConfiguration configuration, CancellationToken ct = default)
--- a/src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
+++ b/src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
@@ -1,6 +1,8 @@
 namespace DotNet.Testcontainers.Containers.Builders
 {
   using System;
+  using System.Threading;
+  using System.Threading.Tasks;
   using DotNet.Testcontainers.Containers.OutputConsumers;
   using DotNet.Testcontainers.Containers.WaitStrategies;
   using DotNet.Testcontainers.Images;
@@ -191,6 +193,14 @@ namespace DotNet.Testcontainers.Containers.Builders
     /// <remarks>Multiple wait strategies are executed one after the other in th…</remarks>
     [PublicAPI]
     ITestcontainersBuilder<TDockerContainer> WithWaitStrategy(IWaitForContainerOS waitStrategy);
+
+    /// <summary>
+    /// Sets a startup callback to be executed after starting the container but before executing the wait strategy.
+    /// </summary>
+    /// <param name="startupCallback">The callback function to be executed.</param>
+    /// <returns>A configured instance of <see cref="ITestcontainersBuilder{TDockerContainer}" />.</returns>
+    [PublicAPI]
+    ITestcontainersBuilder<TDockerContainer> WithStartupCallback(Func<IDockerContainer, CancellationToken, Task> startupCallback);
 
     /// <summary>
     /// Builds the instance of <see cref="IDockerContainer" /> with the given configuration.
--- a/src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
+++ b/src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
@@ -4,6 +4,8 @@ namespace DotNet.Testcontainers.Containers.Builders
   using System.Collections.Generic;
   using System.Linq;
   using System.Reflection;
+  using System.Threading;
+  using System.Threading.Tasks;
   using DotNet.Testcontainers.Clients;
   using DotNet.Testcontainers.Containers.Configurations;
   using DotNet.Testcontainers.Containers.OutputConsumers;
@@ -188,6 +190,12 @@ namespace DotNet.Testcontainers.Containers.Builders
     }
 
     /// <inheritdoc />
+    public ITestcontainersBuilder<TDockerContainer> WithStartupCallback(Func<IDockerContainer, CancellationToken, Task> startupCallback)
+    {
+      return Build(this, Apply(startupCallback: startupCallback));
+    }
+
+    /// <inheritdoc />
     public TDockerContainer Build()
     {
       // Create container instance.
@@ -216,6 +224,7 @@ namespace DotNet.Testcontainers.Containers.Builders
       IEnumerable<IBind> mounts = null,
       IOutputConsumer outputConsumer = null,
       IEnumerable<IWaitUntil> waitStrategies = null,
+      Func<IDockerContainer, CancellationToken, Task> startupCallback = null,
       bool cleanUp = true)
     {
       return new TestcontainersConfiguration(
@@ -233,6 +242,7 @@ namespace DotNet.Testcontainers.Containers.Builders
         mounts,
         outputConsumer,
         waitStrategies,
+        startupCallback,
         cleanUp);
     }
 
@@ -259,6 +269,7 @@ namespace DotNet.Testcontainers.Containers.Builders
       var authConfig = new[] { next.AuthConfig, previous.configuration.AuthConfig }.First(config => config != null);
       var outputConsumer = new[] { next.OutputConsumer, previous.configuration.OutputConsumer }.First(config => config != null);
       var waitStrategies = new[] { next.WaitStrategies, previous.configuration.WaitStrategies }.First(config => config != null);
+      var startupCallback = Merge(next.StartupCallback, previous.configuration.StartupCallback);
 
       var mergedConfiguration = Apply(
         endpoint,
@@ -275,6 +286,7 @@ namespace DotNet.Testcontainers.Containers.Builders
         mounts,
         outputConsumer,
         waitStrategies,
+        startupCallback,
         cleanUp);
 
       return new TestcontainersBuilder<TDockerContainer>(mergedConfiguration, moduleConfiguration ?? previous.moduleConfiguration);
--- a/src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
+++ b/src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
@@ -141,14 +141,14 @@ namespace DotNet.Testcontainers.Containers.Builders
     /// <inheritdoc />
     public ITestcontainersBuilder<TDockerContainer> WithPortBinding(string port, bool assignRandomHostPort = false)
     {
-      var hostPort = assignRandomHostPort ? $"{TestcontainersNetworkService.GetAvailablePort()}" : port;
+      var hostPort = assignRandomHostPort ? null : port;
       return this.WithPortBinding(hostPort, port);
     }
 
     /// <inheritdoc />
     public ITestcontainersBuilder<TDockerContainer> WithPortBinding(string hostPort, string containerPort)
     {
-      var portBindings = new Dictionary<string, string> { { hostPort, containerPort } };
+      var portBindings = new Dictionary<string, string> { { containerPort, hostPort } };
       return Build(this, Apply(portBindings: portBindings));
     }
 
--- a/src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilderKafkaExtension.cs
+++ b/src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilderKafkaExtension.cs
@@ -1,0 +1,31 @@
+namespace DotNet.Testcontainers.Containers.Builders
+{
+  using System.Linq;
+  using System.Text;
+  using System.Threading;
+  using System.Threading.Tasks;
+  using Configurations.MessageBrokers;
+  using DotNet.Testcontainers.Containers.Modules.MessageBrokers;
+
+  /// <summary>
+  /// This class applies the extended Testcontainer configurations for Kafka.
+  /// </summary>
+  public static class TestcontainersBuilderKafkaExtension
+  {
+    public static ITestcontainersBuilder<KafkaTestcontainer> WithKafka(this ITestcontainersBuilder<KafkaTestcontainer> builder, KafkaTestcontainerConfiguration configuration)
+    {
+      builder = configuration.Environments.Aggregate(builder, (current, environment) => current.WithEnvironment(environment.Key, environment.Value));
+
+      return builder
+        .WithImage(configuration.Image)
+        .WithCommand(configuration.Command)
+        .WithPortBinding(configuration.Port, configuration.DefaultPort)
+        .WithWaitStrategy(configuration.WaitStrategy)
+        .WithStartupCallback(configuration.StartupCallback)
+        .ConfigureContainer(container =>
+        {
+          container.Port = configuration.Port;
+        });
+    }
+  }
+}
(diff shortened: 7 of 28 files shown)


src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/AlpineFixture.cs
```diff
@@ -14,9 +14,10 @@ public AlpineFixture()
     {
     }
 
-    public Task InitializeAsync()
+    public async Task InitializeAsync()
     {
-      return Task.WhenAll(this.Container.StartAsync(), this.Container.StopAsync());
+      await this.Container.StartAsync();
```
HofmeisterAn: Depends on the `semaphoreSlim` (see comment `TestcontainersClient`), but `return Task.WhenAll(this.Container.StartAsync(), this.Container.StopAsync());` should be fine.

src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/AlpineFixture.cs
```diff
@@ -14,9 +14,10 @@ public AlpineFixture()
...
+      await this.Container.StopAsync();
```
HofmeisterAn: `.ConfigureAwait(false)`

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/MessageBroker/KafkaTestContainerTest.cs
```diff
@@ -0,0 +1,53 @@
...
+  using DotNet.Testcontainers.Containers.Modules.MessageBrokers;
+  using Xunit;
+
+  public class KafkaTestContainerTest
+  {
+    [Fact]
+    public async Task StartsWorkingKafkaInstance()
+    {
+      // Given
+      var testcontainersBuilder = new TestcontainersBuilder<KafkaTestcontainer>()
```
HofmeisterAn: It would be nice if we could use xUnit `IAsyncLifetime` like `CouchbaseFixture` does.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
```diff
@@ -357,6 +359,34 @@ await using (IDockerContainer testcontainer = testcontainersBuilder.Build())
           Assert.Equal(255, await testcontainer.ExecAsync(new[] { "/bin/sh", "-c", "exit 255" }));
         }
       }
+
+      [Fact]
+      public async Task CopyFileToRunningContainer()
```
HofmeisterAn, line 364: 👍 Just an idea to keep it similar to the other tests:

```csharp
// Given
const string dayOfWeekFilePath = "/tmp/dayOfWeek";

var dayOfWeek = DateTime.Now.DayOfWeek.ToString();

var testcontainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
  .WithImage("alpine")
  .WithCommand("/bin/sh", "-c", "tail -f /dev/null");

// When
// Then
await using (IDockerContainer testcontainer = testcontainersBuilder.Build())
{
  await testcontainer.StartAsync();
  await testcontainer.CopyFileAsync(dayOfWeekFilePath, Encoding.UTF8.GetBytes(dayOfWeek));
  Assert.Equal(0, await testcontainer.ExecAsync(new[] { "/bin/sh", "-c", $"test \"$(cat {dayOfWeekFilePath})\" = \"{dayOfWeek}\"" }));
  Assert.Equal(0, await testcontainer.ExecAsync(new[] { "/bin/sh", "-c", $"stat {dayOfWeekFilePath} | grep 0600" }));
}
```

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -119,6 +119,21 @@ await this.Docker.Exec.StartContainerExecAsync(created.ID, ct)
       return long.MinValue;
     }
 
+    public Task ExtractArchiveToContainerAsync(string id, string path, Stream tarStream, CancellationToken ct = default)
+    {
+      Logger.LogInformation("Copying tar stream to {path} in container {id}", path, id);
```
HofmeisterAn: `at container`.

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
```diff
@@ -4,16 +4,20 @@ namespace DotNet.Testcontainers.Clients
...
   using Docker.DotNet.Models;
   using DotNet.Testcontainers.Containers.Configurations;
   using DotNet.Testcontainers.Containers.OutputConsumers;
   using DotNet.Testcontainers.Images.Configurations;
   using DotNet.Testcontainers.Services;
+  using ICSharpCode.SharpZipLib.Tar;
 
   internal sealed class TestcontainersClient : ITestcontainersClient
   {
+    private readonly SemaphoreSlim semaphoreSlim = new SemaphoreSlim(1, 1);
```
HofmeisterAn: Why did you move `semaphoreSlim`? I think it belongs to `TestcontainersContainer`. It's not necessary to sync the client among different containers.

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -139,14 +139,14 @@ public ITestcontainersBuilder<TDockerContainer> WithPortBinding(int hostPort, in
...
-      var hostPort = assignRandomHostPort ? $"{TestcontainersNetworkService.GetAvailablePort()}" : port;
+      var hostPort = assignRandomHostPort ? null : port;
       return this.WithPortBinding(hostPort, port);
     }
 
     /// <inheritdoc />
     public ITestcontainersBuilder<TDockerContainer> WithPortBinding(string hostPort, string containerPort)
     {
-      var portBindings = new Dictionary<string, string> { { hostPort, containerPort } };
+      var portBindings = new Dictionary<string, string> { { containerPort, hostPort } };
```
HofmeisterAn, line 151: Changing `GetAvailablePort` looks good, but what happens for something like:

```chsarp
.WithPortBinding(80, 81)
.WithPortBinding(80, 82);
```
?

src/DotNet.Testcontainers/Containers/Configurations/Abstractions/TestcontainerDatabaseConfiguration.cs
```diff
@@ -7,7 +7,7 @@ namespace DotNet.Testcontainers.Containers.Configurations.Abstractions
   /// </summary>
   public abstract class TestcontainerDatabaseConfiguration : HostedServiceConfiguration
   {
-    protected TestcontainerDatabaseConfiguration(string image, int defaultPort) : base(image, defaultPort, TestcontainersNetworkService.GetAvailablePort())
+    protected TestcontainerDatabaseConfiguration(string image, int defaultPort) : base(image, defaultPort)
```
HofmeisterAn, line 10: Maybe we can change both `TestcontainerDatabaseConfiguration` and `TestcontainerMessageBrokerConfiguration ` to something like:

```chsarp
protected TestcontainerXYZConfiguration(string image, int defaultPort, int port = 0) : base(image, defaultPort, port)
{
}
```

src/DotNet.Testcontainers/Containers/Configurations/Abstractions/TestcontainerMessageBrokerConfiguration.cs
```diff
@@ -7,7 +7,7 @@ namespace DotNet.Testcontainers.Containers.Configurations.Abstractions
   /// </summary>
   public abstract class TestcontainerMessageBrokerConfiguration : HostedServiceConfiguration
   {
-    protected TestcontainerMessageBrokerConfiguration(string image, int defaultPort) : base(image, defaultPort, TestcontainersNetworkService.GetAvailablePort())
+    protected TestcontainerMessageBrokerConfiguration(string image, int defaultPort) : base(image, defaultPort)
```
HofmeisterAn, line 10: See `TestcontainerDatabaseConfiguration`.

src/DotNet.Testcontainers/Containers/Modules/Abstractions/HostedServiceContainer.cs
```diff
@@ -11,12 +12,12 @@ protected HostedServiceContainer(ITestcontainersConfiguration configuration) : b
...
-    public virtual int Port { get; set; }
+    public virtual int ContainerPort { get; set; }
+
+    public int Port => this.GetMappedPublicPort(this.ContainerPort);
 
     public virtual string Username { get; set; }
 
     public virtual string Password { get; set; }
-
-    public abstract string ConnectionString { get; }
```
HofmeisterAn, line 20: Maybe move that to `TestcontainerDatabase`.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -127,84 +125,40 @@ public ushort GetMappedPublicPort(string privatePort)
...
     {
-      await this.semaphoreSlim.WaitAsync(ct)
-        .ConfigureAwait(false);
-
-      try
-      {
         this.container = await this.Create(ct)
           .ConfigureAwait(false);
-        this.container = await this.Start(this.Id, ct)
+        await this.Start(this.Id, ct)
```
HofmeisterAn: This is necessary for `ContainerHasBeenCreatedStates`.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -15,8 +15,6 @@ public class TestcontainersContainer : IDockerContainer
   {
     private static readonly TestcontainersState[] ContainerHasBeenCreatedStates = { TestcontainersState.Created, TestcontainersState.Running, TestcontainersState.Exited };
 
-    private readonly SemaphoreSlim semaphoreSlim = new SemaphoreSlim(1, 1);
```
HofmeisterAn: I think we should keep the `semaphoreSlim` implementation here.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -127,84 +125,40 @@ public ushort GetMappedPublicPort(string privatePort)
...
     }
 
     public async Task<long> ExecAsync(IList<string> command, CancellationToken ct = default)
     {
-      await this.semaphoreSlim.WaitAsync(ct)
-        .ConfigureAwait(false);
-
-      try
-      {
         return await this.client.ExecAsync(this.Id, command, ct)
```
HofmeisterAn: There are some of them. You don't need to `await`, things like that can be immediately returned.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -217,7 +171,12 @@ public virtual async ValueTask DisposeAsync()
       var cleanOrStopTask = this.configuration.CleanUp ? this.CleanUpAsync() : this.StopAsync();
       await cleanOrStopTask.ConfigureAwait(false);
 
-      this.semaphoreSlim.Dispose();
+      this.client.Dispose();
+    }
+
+    internal virtual Task OnContainerStartedAsync(CancellationToken ct)
```
HofmeisterAn: `protected`.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -233,21 +192,23 @@ private async Task<ContainerListResponse> Create(CancellationToken ct = default)
...
     {
       await this.client.AttachAsync(id, this.configuration.OutputConsumer, ct)
         .ConfigureAwait(false);
       await this.client.StartAsync(id, ct)
         .ConfigureAwait(false);
 
+      this.container = await this.client.GetContainer(id, ct)
+        .ConfigureAwait(false);
+
+      await this.OnContainerStartedAsync(ct);
```
HofmeisterAn: Couldn't we add a function within the configuration like `.With...`? I think that would suite more the implementation and keep the configuration together.

src/DotNet.Testcontainers/Images/Builders/ImageFromDockerfileBuilder.cs
```diff
@@ -53,11 +53,11 @@ public IImageFromDockerfileBuilder WithDeleteIfExists(bool deleteIfExists)
     }
 
     /// <inheritdoc />
-    public Task<string> Build()
+    public async Task<string> Build()
     {
       using (var client = new TestcontainersClient())
       {
-        return client.BuildAsync(this.configuration);
+        return await client.BuildAsync(this.configuration);
```
HofmeisterAn: Why should we await instead of return the task?

src/DotNet.Testcontainers/Internals/Mappers/TestcontainersConfigurationConverter.cs
```diff
@@ -92,7 +92,9 @@ private sealed class ToPortBindings : DictionaryConverter<IEnumerable<KeyValuePa
 
       public override IEnumerable<KeyValuePair<string, IList<PortBinding>>> Convert(IEnumerable<KeyValuePair<string, string>> source)
       {
-        return source?.Select(portBinding => new KeyValuePair<string, IList<PortBinding>>($"{portBinding.Value}/tcp", new List<PortBinding> { new PortBinding { HostPort = portBinding.Key } }));
+        return source?.Select(portBinding => new KeyValuePair<string, IList<PortBinding>>(
+          $"{portBinding.Key}/tcp",
+          new List<PortBinding> { new PortBinding { HostPort = portBinding.Value == null || portBinding.Value == "0" ? null : portBinding.Value } }));
```
HofmeisterAn: `portBinding.Value == null` is not necessary, it will assign null anyway.

src/DotNet.Testcontainers/Containers/Modules/MessageBrokers/KafkaTestcontainer.cs
```diff
@@ -0,0 +1,39 @@
...
+    internal const int ZookeeperPort = 2181;
+    internal const int BrokerPort = 9093;
+
+    internal KafkaTestcontainer(ITestcontainersConfiguration configuration) : base(configuration)
+    {
+    }
+
+    public string BootstrapServers => $"{this.Hostname}:{this.Port}";
+
+    internal override async Task OnContainerStartedAsync(CancellationToken ct)
```
HofmeisterAn: I think this should be part of the configuration itself. This is useful for other containers too, something like `InvokeOnStartup(Action action…)`.

src/DotNet.Testcontainers/Containers/Modules/MessageBrokers/KafkaTestcontainer.cs
```diff
@@ -0,0 +1,39 @@
+namespace DotNet.Testcontainers.Containers.Modules.MessageBrokers
+{
+  using System.Collections.Generic;
+  using System.Text;
+  using System.Threading;
+  using System.Threading.Tasks;
+  using Abstractions;
+  using Configurations;
```
HofmeisterAn: `Abstractions`, `Configurations` please use the full namespace. I think I found a couple more in other files too.

src/DotNet.Testcontainers/Containers/Configurations/MessageBrokers/KafkaTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,41 @@
+namespace DotNet.Testcontainers.Containers.Configurations.MessageBrokers
+{
+  using DotNet.Testcontainers.Containers.Configurations.Abstractions;
+  using DotNet.Testcontainers.Containers.WaitStrategies;
+  using Modules.MessageBrokers;
```
HofmeisterAn: Full namespace.

src/DotNet.Testcontainers.Tests/Unit/Containers/Unix/MessageBroker/KafkaTestContainerTest.cs
```diff
@@ -0,0 +1,53 @@
...
+  using System;
+  using System.Collections.Generic;
+  using System.Threading.Tasks;
+  using Confluent.Kafka;
+  using DotNet.Testcontainers.Containers.Builders;
+  using DotNet.Testcontainers.Containers.Configurations.MessageBrokers;
+  using DotNet.Testcontainers.Containers.Modules.MessageBrokers;
+  using Xunit;
+
+  public class KafkaTestContainerTest
```
HofmeisterAn: It's `KafkaTestcontainerTest` (lower `C`).

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilderKafkaExtension.cs
```diff
@@ -0,0 +1,29 @@
+namespace DotNet.Testcontainers.Containers.Builders
+{
+  using System.Linq;
+  using Configurations.MessageBrokers;
+  using DotNet.Testcontainers.Containers.Configurations.Abstractions;
+  using DotNet.Testcontainers.Containers.Modules.Abstractions;
+  using Modules.MessageBrokers;
```
HofmeisterAn: Full namespace.

HofmeisterAn (changes requested):
Thanks a lot for the pull request. I added a few comments. Most of them are minor changes, but I would like to discuss 2 major change requests:

1. Why didn’t you keep the ` semaphoreSlim` in `TestcontainersContainers`? I think it’s not necessary to synchronize the client among different containers. It’s only important that the container itself is synchronized. The implementation will only allow few public calls, and those are synchronized.
2. I suggest replacing `OnContainerStartedAsync` with a function that can be added to the Testcontainers builder (`ITestcontainersConfiguration`). Similar to `WithWaitStrategy`.
    - This will keep the configuration together.
        - `HostedServiceConfiguration` → Contains the configuration before the container starts (everything to start).
        - `HostedServiceContainer` → Contains the configuration after the container starts (everything to interact).
    - This will follow the current approach which is familiar for other developers.
    - You can move your configuration from `KafkaTestcontainer` to `KafkaTestcontainerConfiguration` where it belongs imo.

Again, thanks for your work!

src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/AlpineFixture.cs
PSanetra: In the tests I got an exception in the `StopAsync` method, because the container was not started yet. And I think this makes sense as `StartAsync` and `StopyAsync` are executed here in parallel. So it is actually kind of random that it has worked with the `semaphoreslim`.

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
PSanetra: By introducing the `OnContainerStartedAsync` Callback it is possible to run into a deadlock, because `OnContainerStartedAsync` itself runs in a part with those semaphore locks. So when I call `ExecAsync` or `CopyFileAsync` inside `OnContainerStartedAsync` the semaphore is already locked. Therefore I thought it would be better to move the semaphore to the client as I thought that the semaphore was actually introduced because we should not call methods of the client in parallel.

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
PSanetra: It does not makes sense to bind the same host port to two different container ports, does it? I guess it is not even possible to do that.

src/DotNet.Testcontainers.Tests/Fixtures/Containers/Modules/AlpineFixture.cs
HofmeisterAn: Yep, I agree. Your suggestion is much cleaner, thanks.

src/DotNet.Testcontainers/Images/Builders/ImageFromDockerfileBuilder.cs
PSanetra: The problem was here that the `Dispose` method of `client` was called before the Task of `BuildAsync` was finished as this method without the await would return before the Task was finished. So it was possible that the `BuildAsync` Task continued while the `client` was actually already disposed.

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
HofmeisterAn: Yep, it doesn't make sense, but I was wondering what Docker will do. In the previous implementation the value was simple overwritten.

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
HofmeisterAn: Good point 👍 but I think we need to take a closer look at this. Locking **each** individual method in the client is not enough. This will not synchronize the public methods. E. g. we don't know if anything else is called between `AttachAsync` and `StartAsync`. This applies to all public methods. Furthermore, the client is responsible for more than just 1 container.

We need to protect the public methods and the execution order. I need some time to think about a proper solution:

- Maybe we can implement it with a second semaphore.
- Do we really need to lock `ExecAsync` or `CopyFileAsync`?
    - We can provide a safe interface to the builder such as:
    ```csharp
    public interface ITestcontainersBuilder<out TDockerContainer>
      where TDockerContainer : IDockerContainer
    {
      ITestcontainersBuilder<TDockerContainer> OnContainerStarted(Action<ISafeInterfaceWithExecAndCopyFileAsync> container);
    }

    new TestcontainersBuilder<TestcontainersContainer>()
      .OnContainerStarted(container => container.ExecAsync(new [] { "" }))
    ```

Do you have any other ideas?

PSanetra: I am not sure why it is important that `StartAsync` and `AttachAsync` are called subsequently without any other calls between them. Those other calls are always independent from these calls and can not corrupt any state, can they?

src/DotNet.Testcontainers/Images/Builders/ImageFromDockerfileBuilder.cs
HofmeisterAn: Oh wow 😀 obvious it isn't `IAsyncDisposable`. Then it should be:

```csharp
public async Task<string> Build()
{
  using (var client = new TestcontainersClient())
  {
    return await client.BuildAsync(this.configuration)
      .ConfigureAwait(false);
  }
}
```

Thanks

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
HofmeisterAn: I have implemented a small [example](https://pastebin.com/raw/0CnDJqUg) using both approaches. You can see that the results of both implementations are different. Your approach is now no longer deterministic and can lead to different results. 

I don't know if this breaks existing code or if it's just theoretical (ofc the example code force this behavior).

src/DotNet.Testcontainers/Containers/Configurations/Abstractions/TestcontainerDatabaseConfiguration.cs
PSanetra: 0 as default value is already set on the `HostedServiceConfiguration` constructor. `TestcontainerDatabaseConfiguration` and `TestcontainerMessageBrokerConfiguration` call that base constructor and use the same default implicitly.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
PSanetra: I have reverted this change

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
PSanetra: I have added a `WithStartupCallback` to the `ITestcontainersBuilder`

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
PSanetra: Yes, generally one should be careful with concurrency, but I don't know if we even need any semaphore. Which state or resource are we protecting from corruption using this semaphore? Is the dotnet Docker client not thread-safe?

If some application depends on `StartAsync` to be finished, then it should wait for the result of the `StartAsync`-Task. Otherwise it is kind of an accident that the application works.

HofmeisterAn: I’ll do the review later that day. I just noticed one thing, shouldn’t be the configuration from `TestcontainersBuilderKafkaExtension` in `KafkaTestcontainerConfiguration` like in the other modules?

<hr />

In reference to the semaphore:
> Which state or resource are we protecting from corruption using this semaphore?
1. Each container instance itself.
2. We protect the field `this.container` in `TestcontainersContainer`.
3. We protect the developers from running into unexpected behaviors and race conditions.

We’re moving the semaphore **only** to work around a deadlock. This will probably cause other issues and make some parts unreliable. I don’t think it’s worth it without considering other solutions.

To get this pull request done, I suggest following:
- We keep the semaphore in `TestcontainersContainer`.
- We do not enter the semaphore in `CopyFileAsync`, this won’t deadlock KafkaTestcontainer.
    - _I guess we can remove the semaphore from `ExecAsync` too. Multiple threads should be able to copy and execute._
- We create a new GitHub issue that describes the deadlock situation and we fix it in a new pull request.
    - _E. g. we could change the interface in `WithStartupCallback`, to a interface that cannot call methodes that enter the semaphore in `TestcontainersContainer` like `IRunningDockerContainer` or whatever._

I hope I don’t bother you too much. Have a nice day.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -218,6 +214,7 @@ public virtual async ValueTask DisposeAsync()
       await cleanOrStopTask.ConfigureAwait(false);
 
       this.semaphoreSlim.Dispose();
+      this.client.Dispose();
```
PSanetra, line 215: I have added the semaphore again, but kept disposing the client here. To me it looks like this was missing before as we are initializing the client in the constructor (line 29), but are not disposing it in the `Dispose` method.

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
PSanetra: I have added the semaphore again to `TestcontainersContainer` and removed its usage in `CopyFileAsync` and `ExecAsync`. Also I have moved the configuration code to `KafkaTestcontainerConfiguration`.

src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
```diff
@@ -192,6 +194,15 @@ public interface ITestcontainersBuilder<out TDockerContainer>
     [PublicAPI]
     ITestcontainersBuilder<TDockerContainer> WithWaitStrategy(IWaitForContainerOS waitStrategy);
 
+    /// <summary>
+    /// Sets a startup callback to be executed after starting the container but before executing the wait strategy.
+    /// </summary>
+    /// <param name="startupCallback">The callback function to be executed.</param>
+    /// <returns>A configured instance of <see cref="ITestcontainersBuilder{TDockerContainer}" />.</returns>
+    /// <remarks>Multiple wait strategies are executed one after the other in th…</remarks>
```
HofmeisterAn: You can remove this `<remarks>` node.

src/DotNet.Testcontainers/Containers/Configurations/ITestcontainersConfiguration.cs
```diff
@@ -2,10 +2,13 @@ namespace DotNet.Testcontainers.Containers.Configurations
 {
   using System;
   using System.Collections.Generic;
+  using System.Threading;
+  using System.Threading.Tasks;
   using DotNet.Testcontainers.Containers.OutputConsumers;
   using DotNet.Testcontainers.Containers.WaitStrategies;
   using DotNet.Testcontainers.Images;
   using JetBrains.Annotations;
+  using Modules;
```
HofmeisterAn: Full namespace.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -193,18 +193,14 @@ await this.semaphoreSlim.WaitAsync(ct)
 
     public async Task<long> ExecAsync(IList<string> command, CancellationToken ct = default)
     {
-      await this.semaphoreSlim.WaitAsync(ct)
+      return await this.client.ExecAsync(this.Id, command, ct)
```
HofmeisterAn: Return task immediately.

HofmeisterAn (approved):
👍 LGTM; I'll only add a few minor improvements and merge it later.

PSanetra:
@HofmeisterAn I have resolved those remaining issues.

HofmeisterAn:
I already prepared the pr yesterday. I added some minor changes in: https://github.com/HofmeisterAn/dotnet-testcontainers/commit/5b8c154b868d808b14390b781bdd9006b5f83939 and created an issue for the deadlock #356. Released in snapshot version: `1.5.0-beta.20210122.3`.

Issue #242, opened 2020-05-08
Force Clean Containers on build before start


mubeda:
.WithCleanUp(true) works well if you don't have any crash on your application. But after a crash you cannot build the container again. For Postgresql you get this exception ("Message = "An item with the same key has already been added. Key: 5432/tcp"). Basically you need to go to the docker app and remove the container manually.
It will be really nice that everytime you build a container with the WithCleanUp(true) it forces the delete of the container if it already exist.

Thanks!

HofmeisterAn:
@mubeda There is already a mechanism to delete remaining containers at the end. See: https://github.com/HofmeisterAn/dotnet-testcontainers/blob/d2469b35d8f7496411ac5ca8d3486b4ab65b8957/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs#L56-L57

Either the event is not triggered or finished in time. The [total execution time](https://docs.microsoft.com/dotnet/api/system.appdomain.processexit?view=netcore-3.1) of all finalizers is limited in .NET Framework. I'm aware that this implementation has some disadvantages. However, it is simple and has always worked in my use cases.

- Which framework do you use?
- Do you always use the same container name?
- Can you test if the event is fired at all and provide some more information?

----

In addition to that. I guess you are mapping the host port `5432` to the container port `5432`. It's recommended to use a random host port and assign that to the default container port. This is the default behavior and good practice, see: https://github.com/HofmeisterAn/dotnet-testcontainers/blob/4acf978e59578c1b99590eafef74edefb10791ff/src/DotNet.Testcontainers/Containers/Configurations/Abstractions/TestcontainerDatabaseConfiguration.cs#L10

It will prevent mapping the same host port multiple times (of course, it's just a workaround and not a proper fix).

mubeda:
Hey Andre! 
Thank you for the response. I use .net core 3.1, i always use use the same container name and fixed ports, i am trying to implement automatic test cases for all dbms i support in my application.
. This problem happens to me once when i stopped the execution of the process, then the code of PurgeOrphanedContainers is never executed.  Basically i had to go to docker and manually delete the running instance. So i was courious if this Purge can be done by instance name on app startup. Feel free to close this ticked if you thing it's not needed.

HofmeisterAn:
[...]

PR #360, opened 2021-01-31
Virtual start and stop

ilmax:
In order to facilitate with #242 I came with this 2 liner PR so we can manually attempt to clean up if start fails with something like the following pseudo:

```csharp
public class AutoCleanupTestContainer : TestcontainersContainer
{
    protected AutoCleanupTestContainer(ITestcontainersConfiguration configuration)
        : base(configuration)
    { }

    public async override Task StartAsync(CancellationToken ct)
    {
    retry:
        try
        {
            await StartAsync(ct).ConfigureAwait(false);
        }
        catch (DockerApiException dockerApiException) when (dockerApiException.StatusCode == HttpStatusCode.Conflict)
        {
            // Cleanup logic
            goto retry;
        }
    }
}
```

I know this is not the full fix discussed in 242 but it may be eligible as a workaround, what do you think about it?

### DIFF
--- a/src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
+++ b/src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
@@ -130,7 +130,7 @@ namespace DotNet.Testcontainers.Containers.Modules
       return this.client.GetContainerExitCode(this.Id, ct);
     }
 
-    public async Task StartAsync(CancellationToken ct = default)
+    public virtual async Task StartAsync(CancellationToken ct = default)
     {
       await this.semaphoreSlim.WaitAsync(ct)
         .ConfigureAwait(false);
@@ -148,7 +148,7 @@ namespace DotNet.Testcontainers.Containers.Modules
       }
     }
 
-    public async Task StopAsync(CancellationToken ct = default)
+    public virtual async Task StopAsync(CancellationToken ct = default)
     {
       await this.semaphoreSlim.WaitAsync(ct)
         .ConfigureAwait(false);


HofmeisterAn:
Thanks for sharing your ideas! It's something I haven't thought about it. It sounds pretty interesting. Especially, this will work for Unix and Windows. I have two concerns so far:

1. How do we get the id of the orphaned container.
2. How do we prevent developers from applying two different configurations (`WithCleanUp` + `AutoCleanupTestContainer`). I think that's confusing.

Perhaps we can find a proper way to solve both concerns. I attached a small [example](https://github.com/HofmeisterAn/dotnet-testcontainers/files/5899562/GitHub-PR360.patch.txt).

<hr />

On the other side, I'm thinking about a simpler solution. We can label the registered containers in `TestcontainersRegistryService.Register` and purge them on initialization or with a scheduler. I'm only thinking about the right labels. E. g. a container labeled with `Testcontainers` and the Docker status `Exited` we can remove immediately. This will cover the most cases and a good start, I think.

- We know the orphaned container id.
- We do not require any further configuration.

https://github.com/HofmeisterAn/dotnet-testcontainers/blob/5b8c154b868d808b14390b781bdd9006b5f83939/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs#L180

What do you think?

ilmax:
Hi @HofmeisterAn  thank you for the fast response.

> How do we get the id of the orphaned container.

I was achieving it via this code:

```csharp
var uri = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? new Uri("npipe://./pipe/docker_engine") : new Uri("unix:/var/run/docker.sock");
var dockerClient = new DockerClientConfiguration(uri).CreateClient();

// Stop the container if it's running and remove it
await dockerClient.Containers.RemoveContainerAsync(name, new ContainerRemoveParameters
{
    Force = true
});
```

so you actually can get away with it just knowing the name of the container you want to remove

> How do we prevent developers from applying two different configurations (WithCleanUp + AutoCleanupTestContainer). I think that's confusing.

Not sure I got this one, but my Idea is to allow a dev to implement it's own logic so I wasn't thinking about implementing the `AutoCleanupTestContainer` in the library itself. Just give the proper hooks so people can built their own version of it.

Looking at the example it may work for some but the implementation heavily relies on the test framework of choice and I'm not sure all of them supports the same concepts (i.e. a global cleanup) moreover IMO it's quite common to debug a unit test, find a bug and stop the debugging session (or at least I'm doing it all the time) and this will prevent the cleanup method from being called because AFAIK the process is killed.

So I really would like to have an hook to override wen we try to start a container so I can catch an exception and try to cleanup myself.

So to sum up I think a proper solution like the one you outlined above it's definitely a better approach but (unless I'm missing something 😅 ) it may not suffice to resolve all issues, thus I still think there's some value in allowing the end user to override the start and stop methods.

EDIT: The autopurge on startup based on labels would be awesome!

HofmeisterAn:
> So you actually can get away with it just knowing the name of the container you want to remove.

Unfortunately, that only applies to your example. Many tests don't use `WithName`. How would you cover those tests?

> Just give the proper hooks so people can built their own version of it.

Ok, got it. I still think two different ways to clean up containers are bad. At least `WithCleanUp` should do a better job 😀.

> Looking at the example it may work for some but the implementation heavily relies on the test framework of choice and I'm not sure all of them supports the same concepts (i.e. a global cleanup) moreover IMO it's quite common to debug a unit test, find a bug and stop the debugging session (or at least I'm doing it all the time) and this will prevent the cleanup method from being called because AFAIK the process is killed.

Which part? Do you talk about the diff? That was just an example to figure out what else is needed. The idea is still the same. Remove orphaned containers at the start.

> allowing the end user to override the start and stop methods.

Yep, that makes sense 👍. I'll add that in the next pull request.

> The autopurge on startup based on labels would be awesome!

I think too. I'll take a look into it at the end of next week.

ilmax:
> Unfortunately, that only applies to your example. Many tests don't use `WithName`. How would you cover those tests?

It won't but some docs/example can help people with that.
> Ok, got it. I still think two different ways to clean up containers are bad. At least `WithCleanUp` should do a better job 😀.

I agree that `WithCleanUp` can do a better job there, but in the meantime this will allow implementing our own hand rolled cleanup implementation and I don't see any issue in having virtual methods in the future when `WithCleanUp` will be smarter.
> Yep, that makes sense 👍. I'll add that in the next pull request.

If you like the idea, we can merge this one in, let me know if you want me to open an issue first.
> I think too. I'll take a look into it at the end of next week.

Thank you for that 🙏 Let me know if I can help in any way.

HofmeisterAn:
> If you like the idea, we can merge this one in, let me know if you want me to open an issue first.

HofmeisterAn:
> If you like the idea, we can merge this one in, let me know if you want me to open an issue first.

If you don't mind, I'll just merge 59d1b4f039100e922181457ae4e7b2830c74e95a from this pull request tomorrow. I don't think we really need the test.

> Thank you for that 🙏 Let me know if I can help in any way.

Ofc. I will get in touch with you end of next week.
Thanks.

ilmax:
Awesome, thank you!

ilmax:
@HofmeisterAn just dropped the test commit so you can merge this one in directly if you want

HofmeisterAn (approved):

HofmeisterAn:
@ilmax FYI it's deployed in `1.5.0-beta.20210201.3`.

ilmax:
Thank you, I'm gonna try this out later on today

ilmax:
@HofmeisterAn Yesterday I tried to implement the autocleanup at startup and come up with this code:

```csharp
public sealed class SqlServerTestcontainer : TestcontainerDatabase
{
    internal SqlServerTestcontainer(ITestcontainersConfiguration configuration)
        : base(configuration)
    { }

    public override string ConnectionString => $"Server=127.0.0.1,{Port};Database={Database};User Id={Username};Password={Password};";

    public override async Task StartAsync(CancellationToken cancellationToken = default)
    {
        bool retry = true;
    tryAgain:
        try
        {
            await base.StartAsync(cancellationToken);
        }
        catch (DockerApiException dockerApiException) when (retry && dockerApiException.StatusCode == HttpStatusCode.Conflict)
        {
            retry = false;
            await NukeItAsync("sql-server-db");
            goto tryAgain;
        }
    }

    private async Task NukeItAsync(string name)
    {
        var uri = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? new Uri("npipe://./pipe/docker_engine") : new Uri("unix:/var/run/docker.sock");
        var dockerClient = new DockerClientConfiguration(uri).CreateClient();

        // Stop the container if it's running and remove it
        await dockerClient.Containers.RemoveContainerAsync(name, new ContainerRemoveParameters { Force = true });
    }
}
```

used like this:

```csharp
var builder = new TestcontainersBuilder<SqlServerTestcontainer>()
    .WithName("sql-server-db")
    .WithDatabase(new MsSqlTestcontainerConfiguration("mcr.microsoft.com/mssql/server:2019-latest")
    {
        Password = "Guess_me",
        Port = 1535
    });

_container = builder.Build();
await _container.StartAsync();
```

This works like a charm for me, only catch is that I cannot use the Name property on the cleanup since it's only available after the container has bee successfully initialized.

HofmeisterAn:
I started yesterday with the other implementation I mentioned above. I'll get back to you as soon as it's ready to test.

stevo-knievo:
I'm happy to test it too

HofmeisterAn:
@stevo-knievo 👍 FYI: In general, containers will be removed through `Dispose()` (this is preferred). Orphaned containers will be removed within the next start if their status is exited.

PR #361, opened 2021-02-03
Feature/242 label each container to cleanup at next start

HofmeisterAn:
@ilmax I added the first part to clean orphaned containers from previous runs. The idea is to label each container and remove the remaining ones in the next run. Until now, this will only cover containers with the status `exited`. If a container is still running, we don’t know if it is created from another test process.

I'll try to cover that in the next commit. We can probably assign a kind of session id to each container, to identify multiple processes.

### DIFF
--- a/src/DotNet.Testcontainers.Tests/Unit/GitHub.cs
+++ b/src/DotNet.Testcontainers.Tests/Unit/GitHub.cs
@@ -1,0 +1,17 @@
+namespace DotNet.Testcontainers.Tests.Unit
+{
+  using System;
+  using System.Threading.Tasks;
+  using DotNet.Testcontainers.Clients;
+  using Xunit;
+
+  public class GitHub
+  {
+    [Fact]
+    public Task PullRequest361()
+    {
+      Assert.NotEqual(Guid.Empty, new TestcontainersSession().Id);
+      return Task.CompletedTask;
+    }
+  }
+}
--- a/src/DotNet.Testcontainers/Clients/IHasListOperations.cs
+++ b/src/DotNet.Testcontainers/Clients/IHasListOperations.cs
@@ -8,6 +8,8 @@ namespace DotNet.Testcontainers.Clients
   {
     Task<IEnumerable<T>> GetAllAsync(CancellationToken ct = default);
 
+    Task<IEnumerable<T>> GetOrphanedObjects(CancellationToken ct = default);
+
     Task<T> ByIdAsync(string id, CancellationToken ct = default);
 
     Task<T> ByNameAsync(string name, CancellationToken ct = default);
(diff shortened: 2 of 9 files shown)


src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
```diff
@@ -167,13 +172,22 @@ await this.containers.ExtractArchiveToContainerAsync(id, "/", memStream, ct)
...
+        .Select(container => this.containers.RemoveAsync(container.ID, ct));
+
       if (!await this.images.ExistsWithNameAsync(configuration.Image.FullName, ct)
         .ConfigureAwait(false))
       {
         await this.images.CreateAsync(configuration.Image, configuration.AuthConfig, ct)
           .ConfigureAwait(false);
       }
 
+      await Task.WhenAll(removeOrphanedContainersTasks)
```
ilmax, line 190: Shouldn't this one be done before the image creation?

ilmax:
@HofmeisterAn I went through the code and I left a nit, this change looks good to me. Solving the other problem may be a bit tricky so I was thinking maybe the user can decide what to do with an overload of the `WithCleanup` that allows to remove images for running container

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
HofmeisterAn: No. We can pull not existing images while we're removing orphaned containers. We will not remove images.

src/DotNet.Testcontainers/Clients/IHasListOperations.cs
```diff
@@ -8,6 +8,8 @@ internal interface IHasListOperations<T>
   {
     Task<IEnumerable<T>> GetAllAsync(CancellationToken ct = default);
 
+    Task<IEnumerable<T>> GetOrphanedObjects(CancellationToken ct = default);
```
ilmax, line 11:
```suggestion
    Task<IEnumerable<T>> GetOrphanedObjectsAsync(CancellationToken ct = default);
```

src/DotNet.Testcontainers.Tests/Unit/GitHub.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+  using System;
+  using System.Threading.Tasks;
+  using DotNet.Testcontainers.Clients;
+  using Xunit;
+
+  public class GitHub
+  {
+    [Fact]
+    public Task PullRequest361()
```
ilmax, line 11:
```suggestion
    public void PullRequest361()
```

HofmeisterAn: This was just a test commit to see if it works on Unix. `mapName` is not supported. I will change it to `CreateFromFile` in the next days and test it on a separate branch. Than I'll merge it back to this pr.

HofmeisterAn:
Sharing the same session between multiple processes, do not work right away with Unix systems and .NET Core. To clean up orphaned containers that are still **running** we need a different approach. We probably can merge e0495011ac79d3fbca077c016d2a18873d89d779. This will still help in a lot of cases, but I would like to do some more tests. I’ll close this draft in the meantime.

PR #362, opened 2021-02-08
added hostname on container creating

Eshva:
Why I need it. I need to write integration and end-to-end tests using
Kafka. Kafka needs Zookeeper. When you start them in containers you
need to configure it in some specific way. For that you need to
reference Kafka in container one time with `localhost` and another with
its hostname. If you don't know or don't control this hostname your
Kafka will be unreachable. The easiest way to accomplish this is to
set it during the container creating. Docker Compose for instance allows
to set the container internal hostname (and a Docker network hostname)
in the compose-file. Docker API allows to do it but this library for
some reason ignores this setting. I've added this missing ability.
I've seen a different approach to fix this problem in samples but it
requires some shell-scripts to run during the Kafka container startup.
It's too complex and not flexible. What if you can't (and in most cases
can't) alter your Kafka image? My way to fix this problem is much more
elegant and easy but requires to change some internals of this library.

### DIFF
--- a/src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
+++ b/src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
@@ -48,6 +48,13 @@ namespace DotNet.Testcontainers.Containers.Builders
     ITestcontainersBuilder<TDockerContainer> WithName(string name);
 
     /// <summary>
+    /// Sets the hostname of the Testcontainer.
+    /// </summary>
+    /// <param name="hostname">Testcontainers name.</param>
+    /// <returns>A configured instance of <see cref="ITestcontainersBuilder{TDockerContainer}" />.</returns>
+    ITestcontainersBuilder<TDockerContainer> WithHostname(string hostname);
+
+    /// <summary>
     /// Overrides the working directory of the Testcontainer for the instruction sets.
     /// </summary>
     /// <param name="workingDirectory">Working directory.</param>
--- a/src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
+++ b/src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
@@ -50,7 +50,7 @@ namespace DotNet.Testcontainers.Containers.Builders
     /// <summary>
     /// Sets the hostname of the Testcontainer.
     /// </summary>
-    /// <param name="hostname">Testcontainers name.</param>
+    /// <param name="hostname">Testcontainers hostname.</param>
     /// <returns>A configured instance of <see cref="ITestcontainersBuilder{TDockerContainer}" />.</returns>
     ITestcontainersBuilder<TDockerContainer> WithHostname(string hostname);
 
--- a/src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
+++ b/src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
@@ -93,7 +93,7 @@ namespace DotNet.Testcontainers.Containers.Modules
               return "localhost";
             }
           default:
-            return this.IpAddress;
+            return this.configuration.Hostname;
         }
       }
     }
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Keep in mind to enable the correct Docker engine on Windows host systems to matc
 - `WithEntrypoint` specifies and overrides the `ENTRYPOINT` that will run as an executable.
 - `WithCommand` specifies and overrides the `COMMAND` instruction provided from the Dockerfile.
 - `WithName` sets the container name e. g. `--name nginx`.
+- `WithHostname` sets the container hostname e. g. `--hostname my-nginx`.
 - `WithEnvironment` sets an environment variable in the container e. g. `-e, --env "test=containers"`.
 - `WithLabel` applies metadata to a container e. g. `-l, --label dotnet.testcontainers=awesome`.
 - `WithExposedPort` exposes a port inside the container e. g. `--expose=80`.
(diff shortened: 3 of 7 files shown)


Eshva:
I configure Kafka for Testcontainers.NET this way:
```csharp
    public override IReadOnlyCollection<KeyValuePair<string, string>> GetVariables() => new[]
    {
      new KeyValuePair<string, string>(
        "KAFKA_LISTENERS",
        $"INTERNAL://:{DefaultInternalCommunicationPort},EXTERNAL://:{BootstrapPort}"),
      new KeyValuePair<string, string>(
        "KAFKA_ADVERTISED_LISTENERS",
        $"INTERNAL://{ContainerName}:{DefaultInternalCommunicationPort},EXTERNAL://localhost:{BootstrapPort}"),
      new KeyValuePair<string, string>("KAFKA_LISTENER_SECURITY_PROTOCOL_MAP", "INTERNAL:PLAINTEXT,EXTERNAL:PLAINTEXT"),
      new KeyValuePair<string, string>("KAFKA_INTER_BROKER_LISTENER_NAME", "INTERNAL"),
      new KeyValuePair<string, string>("KAFKA_ZOOKEEPER_CONNECT", ZookeeperAddress),
      new KeyValuePair<string, string>("ALLOW_PLAINTEXT_LISTENER", "true"),
      new KeyValuePair<string, string>("KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR", "1"),
      new KeyValuePair<string, string>("KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE", "true")
    };
```
Notice this part:
```csharp
      new KeyValuePair<string, string>(
        "KAFKA_ADVERTISED_LISTENERS",
        $"INTERNAL://{ContainerName}:{DefaultInternalCommunicationPort},EXTERNAL://localhost:{BootstrapPort}"),
```
For internal communications (in one node cluster it is the single node itself) Kafka uses it's container hostname but for external it advertised itself as `localhost`. And for my tests running on local computer Kafka container exposes itself on the standard port on `localhost`.

P.S. The excerpt is from my own KafkaContainer class that incapsulates Kafka and Zookeeper in a single class. And it works.

Eshva:
Seams like you need to fix 3 tests for CouchbaseTestcontainerTest:
```
[xUnit.net 00:03:43.56]     DotNet.Testcontainers.Tests.Unit.Containers.Unix.Database.CouchbaseTestcontainerTest.ShouldFlushBucket [FAIL]
[xUnit.net 00:03:43.60]     DotNet.Testcontainers.Tests.Unit.Containers.Unix.Database.CouchbaseTestcontainerTest.ConnectionEstablishedWithCrudTest 
[xUnit.net 00:03:47.81]     DotNet.Testcontainers.Tests.Unit.Containers.Unix.Database.CouchbaseTestcontainerTest.ShouldCreateBucket [FAIL]
```
They are broken from 5 February.

Please review my changes.

HofmeisterAn:
Thanks! I'll do the review later that day.

src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
```diff
@@ -47,6 +47,13 @@ public interface ITestcontainersBuilder<out TDockerContainer>
     [PublicAPI]
     ITestcontainersBuilder<TDockerContainer> WithName(string name);
 
+    /// <summary>
+    /// Sets the hostname of the Testcontainer.
+    /// </summary>
+    /// <param name="hostname">Testcontainers name.</param>
```
HofmeisterAn: `Testcontainers hostname`

HofmeisterAn (commented):
PR looks good to me 👍, thanks! I have two short questions:

1. >I've seen a different approach to fix this problem in samples but it
requires some shell-scripts to run during the Kafka container startup.
It's too complex and not flexible.

    I'm not familiar with Kafka, Zookeeper, etc. But can we remove the `StartupCallback` in `KafkaTestcontainerConfiguration`? It would be very kind if you can add your approach to the configuration instead.

2. Does it make sense to return the configured hostname in `TestcontainersContainer.Hostname`
    https://github.com/HofmeisterAn/dotnet-testcontainers/blob/a991d569e98286b0a506fa57b3a02d7733c263b7/src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs#L74
    I don't think, but I would like to double check.

HofmeisterAn:
> Seams like you need to fix 3 tests for CouchbaseTestcontainerTest:

Odd, I'm not able to pull the images too. I'll have a look at it. If you like skip the Couchbase tests.

Eshva:
> 1. I'm not familiar with Kafka, Zookeeper, etc. But can we remove the `StartupCallback` in `KafkaTestcontainerConfiguration`? It would be very kind if you can add your approach to the configuration instead.

It's an another PR. 😄 
And probably before that I want to discus/fix another problem about `/global.json` file.

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
```diff
@@ -93,7 +93,7 @@ public string Hostname
               return "localhost";
             }
           default:
-            return this.IpAddress;
+            return this.configuration.Hostname;
```
Eshva, line 96: I guess you mean this case only. I'm not so sure about other cases.

src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
Eshva: done

src/DotNet.Testcontainers/Containers/Modules/TestcontainersContainer.cs
HofmeisterAn: Yep, but I don't think it will resolve the name immediately. At least in my small example it didn't. The IP should be better, for now. I can update it.

HofmeisterAn (approved):

HofmeisterAn:
Published in `1.5.0-beta.20210210.3`.

PR #363, opened 2021-02-10
fix build in the current version of VS

Eshva:
If I open the solution in VS2019 of the current version all around of
the code I see missing references to string, Guid and similar staff. If
I delete global.json from the root the problem is fixed immediately.
Seams like this file isn't required anymore because the only thing it
does it pinpoints solution projects to an old version of .NET Core.
There was a problem with version incompatibility between used by VS and
the latest .NET Core for that moment but isn't an issue anymore.
May be you need it to build on Azure or something else I don't know.
Fill free to reject if you really need it.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,0 @@
-{
-  "sdk": {
-    "version": "3.1.100",
-    "rollForward": "latestPatch"
-  }
-}


HofmeisterAn:
This behavior is on purpose. It pins the .NET Core version, see: https://docs.microsoft.com/de-de/dotnet/core/tools/global-json. Either install a compatible version (3.1.100; 00 is the patch) or set the version property to a SDK on your computer. Don't commit changes to that file.

Eshva:
Ok. No problem.

PR #372, opened 2021-03-08
WIP: Add WithDockerEndpoint overload with credentials

prskr:
Please note that this is a work in progress PR to get early feedback.

I'll squash all commits according to the commit message format described in the guidelines as soon as the PR is completed and ready to be merged.


global.json
```diff
@@ -1,6 +1,6 @@
 {
```
HofmeisterAn, line 1: Do not commit this change or update the pipeline as well.

src/DotNet.Testcontainers.Tests/Unit/Containers/TestcontainersAccessInformationTest.cs
```diff
@@ -2,6 +2,7 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers
 {
   using System;
   using System.Threading.Tasks;
+  using Docker.DotNet;
```
HofmeisterAn: `using Docker.DotNet;` is not necessary.

src/DotNet.Testcontainers/Clients/DockerApiEndpoint.cs
```diff
@@ -1,10 +1,54 @@
 namespace DotNet.Testcontainers.Clients
 {
   using System;
-  using DotNet.Testcontainers.Services;
+  using System.IO;
+  using System.Linq;
+  using Docker.DotNet;
+  using Docker.DotNet.X509;
+  using Services;
```
HofmeisterAn: Use full namespaces.

src/DotNet.Testcontainers/Clients/DockerApiEndpoint.cs
```diff
@@ -1,10 +1,54 @@
...
   {
-    public static Uri Local => TestcontainersHostService.GetService<IOperatingSystem>().DockerApiEndpoint;
+    private const string DockerTlsVerifyEnvName = "DOCKER_TLS_VERIFY";
+    private const string DockerCertPathEnvName = "DOCKER_CERT_PATH";
+
+    private static readonly string[] RequiredCertFiles = { "ca.pem", "cert.pem", "key.pem" };
+
+    public static Uri Default => TestcontainersHostService.GetService<IOperatingSystem>().DockerApiEndpoint;
+
+    public static Credentials DefaultCredentials => DetermineCredentialsFromEnv();
```
HofmeisterAn: ~~This should be part of `TestcontainersBuilder`. The first task I mentioned simply allows the configuration for the client certificate authentication through the builder pattern. Like it's done for every other case~~. I have not expressed myself very well here. IMO it should be more like `IOperatingSystem` and part of its own class.

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -6,18 +6,19 @@ namespace DotNet.Testcontainers.Clients
...
+  using Docker.DotNet;
   using Docker.DotNet.Models;
-  using DotNet.Testcontainers.Containers.Configurations;
-  using DotNet.Testcontainers.Containers.OutputConsumers;
-  using DotNet.Testcontainers.Internals.Mappers;
-  using DotNet.Testcontainers.Services;
+  using Containers.Configurations;
+  using Containers.OutputConsumers;
+  using Internals.Mappers;
+  using Services;
```
HofmeisterAn: Use full namespaces.

src/DotNet.Testcontainers/Clients/DockerImageOperations.cs
```diff
@@ -6,15 +6,16 @@ namespace DotNet.Testcontainers.Clients
...
+  using Docker.DotNet;
   using Docker.DotNet.Models;
-  using DotNet.Testcontainers.Containers.Configurations;
-  using DotNet.Testcontainers.Images;
-  using DotNet.Testcontainers.Images.Archives;
-  using DotNet.Testcontainers.Images.Configurations;
+  using Containers.Configurations;
+  using Images;
+  using Images.Archives;
+  using Images.Configurations;
```
HofmeisterAn: Use full namespaces.

src/DotNet.Testcontainers/Clients/DockerSystemOperations.cs
```diff
@@ -3,10 +3,12 @@ namespace DotNet.Testcontainers.Clients
   using System;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet;
+  using JetBrains.Annotations;
```
HofmeisterAn: `using JetBrains.Annotations;` is not necessary.

src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs
```diff
@@ -176,6 +177,16 @@ public interface ITestcontainersBuilder<out TDockerContainer>
...
 
+
+    /// <summary>
+    /// TODO
+    /// </summary>
+    /// <param name="endpoint"></param>
+    /// <param name="credentials"></param>
+    /// <returns></returns>
+    [PublicAPI]
+    ITestcontainersBuilder<TDockerContainer> WithDockerEndpoint(string endpoint, Credentials credentials);
```
HofmeisterAn: Do not use Docker remote API specific types here.  To keep the API more stable use primitive types or an own interface for .NET Testcontainers (like `WithRegistryAuthentication`, `IDockerImage`, etc.).

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -7,6 +7,7 @@ namespace DotNet.Testcontainers.Containers.Builders
   using System.Reflection;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet;
```
HofmeisterAn, line 10: Do not use Docker remote API specific types here.

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -238,7 +246,8 @@ public TDockerContainer Build()
       bool cleanUp = true)
     {
       return new TestcontainersConfiguration(
-        endpoint ?? DockerApiEndpoint.Local,
+        endpoint ?? DockerApiEndpoint.Default,
+        endpointCredentials ?? DockerApiEndpoint.DefaultCredentials,
```
HofmeisterAn: Replace this with:

```csharp
return new TestcontainersConfiguration(
  endpoint,
  endpointCredentials,
  ...,
```

and apply the default configuration at Line:44 (maybe similar to `AuthenticationConfiguration`?):

```csharp
public TestcontainersBuilder() : this(
  Apply(
    authConfig: new AuthenticationConfiguration(),
    labels: new DefaultLabels(),
    endpoint: .... /* Something like that. */
    useTlsVerification: ....
    certificatePath: ....
    outputConsumer: Consume.DoNotConsumeStdoutAndStderr(),
    waitStrategies: Wait.ForUnixContainer().Build(),
    startupCallback: (_, _) => Task.CompletedTask),
  _ => { })
{
}
```

If one of the env variables are set, apply the configuration here.

src/DotNet.Testcontainers/Containers/Configurations/ITestcontainersConfiguration.cs
```diff
@@ -4,6 +4,7 @@ namespace DotNet.Testcontainers.Containers.Configurations
   using System.Collections.Generic;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet;
```
HofmeisterAn: Do not use Docker remote API specific types here.

src/DotNet.Testcontainers/Containers/Configurations/TestcontainersConfiguration.cs
```diff
@@ -4,6 +4,7 @@ namespace DotNet.Testcontainers.Containers.Configurations
   using System.Collections.Generic;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet;
```
HofmeisterAn: Do not use Docker remote API specific types here.

src/DotNet.Testcontainers/DotNet.Testcontainers.csproj
```diff
@@ -4,6 +4,7 @@
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Docker.DotNet.X509" Version="3.125.2" />
```
HofmeisterAn: Put it below `Docker.DotNet`.

src/DotNet.Testcontainers/Services/IOperatingSystem.cs
```diff
@@ -7,6 +7,8 @@ namespace DotNet.Testcontainers.Services
   /// </summary>
   internal interface IOperatingSystem
   {
+    internal const string DockerHostEnvName = "DOCKER_HOST";
```
HofmeisterAn, line 10: Register it as an implementation of IOperatingSystem in TestcontainersHostService. Resolve it first if the env variable is set. Maybe we can use the same approach for the other env variables too.

Something like (maybe we can improve this)? What do you think?

```csharp
internal readonly struct DockerHostOperatingSystem : IOperatingSystem
{
  private readonly IOperatingSystem os;

  public DockerHostOperatingSystem(IOperatingSystem os)
  {
    this.os = os;
    this.DockerApiEndpoint = new Uri(Environment.GetEnvironmentVariable("DOCKER_HOST") ?? this.os.DockerApiEndpoint.AbsoluteUri);
  }

  public Uri DockerApiEndpoint { get; }

  public string NormalizePath(string path)
  {
    return this.os.NormalizePath(path);
  }
}
```

src/DotNet.Testcontainers/Services/Unix.cs
```diff
@@ -1,13 +1,16 @@
...
   using System;
+  using static IOperatingSystem;
 
   /// <summary>
   /// Unix operating system.
   /// </summary>
   internal readonly struct Unix : IOperatingSystem
   {
-    private static readonly Uri endpoint = new Uri("unix:/var/run/docker.sock");
+    private const string DefaultDockerSocketPath = "unix:/var/run/docker.sock";
```
HofmeisterAn: See `IOperatingSystem`.

src/DotNet.Testcontainers/Services/Windows.cs
```diff
@@ -9,7 +10,8 @@ namespace DotNet.Testcontainers.Services
   {
 #pragma warning disable S1075
 
-    private static readonly Uri endpoint = new Uri("npipe://./pipe/docker_engine");
+    private const string DefaultDockerSocketPath = "npipe://./pipe/docker_engine";
```
HofmeisterAn: See `IOperatingSystem`.

HofmeisterAn (changes requested):
Not that much to change. Just a few requests to keep the implementation more consistent:

- Do not use `Docker.DotNet` in the mentioned classes. I like to hide it, to keep the API more stable.
- Move the default configuration into `TestcontainersBuilder` line 44.
    - Maybe put the certificate configuration into its own class and register it. Similar to `IOperatingSystem`.
    I attached an example for `DOCKER_HOST` below.

Thanks for your work and the pull request.

prskr:
I updated the PR and tried to apply as much of your suggestions as possible.

Right now this would be a breaking change but I'm not sure how you would like to avoid this therefore I thought the easiest way to discuss this would be to push the already broken code and see if we can find a good solution if there is any.

I'm also not that satisfied with the naming of the `DockerClientAuthConfig` and I thought it me be even better to make a breaking change now and introduce something like a `DockerClientConfig` that also includes the endpoint URI to be able to extend it in the future.

The cert loading is still under construction because I'd like to actually parse the certificate labels to not introduce kinda 50% solution.

I hope I applied the `IOperatingSystem` suggestion correct.

global.json
prskr: Updated the Azure pipeline config - anything else?

HofmeisterAn: Yes, your change in `azure-pipelines.yml` is correct. But it would be enough to keep `5.0.200` in `global.json`. See [rollForward](https://docs.microsoft.com/de-de/dotnet/core/tools/global-json?tabs=netcore3x#rollforward) (latestPatch).

HofmeisterAn:
> Right now this would be a breaking change but I'm not sure how you would like to avoid this therefore I thought the easiest way to discuss this would be to push the already broken code and see if we can find a good solution if there is any.

I'll take a look at it in the next days. I'm pretty sure we will find an implementation that won't be a breaking change.

> I'm also not that satisfied with the naming of the DockerClientAuthConfig and I thought it me be even better to make a breaking change now and introduce something like a DockerClientConfig that also includes the endpoint URI to be able to extend it in the future.

I try to add my suggestions on top of your pull request.

> The cert loading is still under construction because I'd like to actually parse the certificate labels to not introduce kinda 50% solution.

👍

global.json
prskr: Interesting, had some issues when loading the project. I'll try to revert the change here but I'd keep the latest version in the Azure pipeline, no reason to keep the old one I guess?

HofmeisterAn: >Interesting, had some issues when loading the project

That's on purpose if you do not have a compatible patch Version. This will prevent different behaviors with different Versions of .NET Core. This file pins the exact version.

>  I'll try to revert the change here

I have added some changes to your forked branch by pull request. I included that one as well.

> no reason to keep the old one I guess?

Yep, we can update the pipeline to a newer version.

global.json
```diff
@@ -1,6 +1,6 @@
...
   "sdk": {
-    "version": "5.0.100",
+    "version": "5.0.103",
```
HofmeisterAn, line 3: This should be `5.0.200` again 😀.

prskr:
I did some work yesterday to move the cert loading to `PemCertificateCredentials` and to move back to `Credentials`.
To be honest I'm not sure why I'd need another implementation of `IDockerClientConfiguration` because most of the logic required to support other credentials could be move either to constructors (e.g. one that takes username, password to initialize HTTP basic credentials) or to custom `Credentials` implementations like the `PemCertificateCredentials`.

I would now add a base class for `PemCertificateCredentials` and `X509CertificateCredentials` that takes `(bool tlsVerify, X509Certificate2 caCert)` to handle the certificate validation at the base class and then all the magic for handling different cert mechanisms goes to inherited classes? Or do you prefer a different approach?

BTW: the certs I added are just random generated ones from Docker-in-Docker for testing the cert loading mechanism.

HofmeisterAn (commented):
> I would now add a base class for `PemCertificateCredentials` and `X509CertificateCredentials` that takes `(bool tlsVerify, X509Certificate2 caCert)` to handle the certificate validation at the base class and then all the magic for handling different cert mechanisms goes to inherited classes? Or do you prefer a different approach?

Sounds good. I'll resolve the conversations from the previous review and look at your changes if your done with `PemCertificateCredentials` and `X509CertificateCredentials`, ok? I took a quick look at the changes from yesterday. Looks good so far 👍 . As I said, if you need any help -  let me know.

global.json
prskr: I know :sweat_smile: I'll try to keep this in mind but I worked on a different machine yesterday with a different patch level and I wanted to see if I understood the config option in the `global.json` correctly (I did not :sweat_smile: ) therefore I kept it like this but eventually I will make sure it matches the one in the Azure pipeline

HofmeisterAn: > I worked on a different machine yesterday with a different patch level and I wanted to see if I understood the config option in the `global.json` correctly (I did not 😅)

`200` is not the patch version. It's a bit different than semantic versioning. It's `x.y.znn`:

- `x` is the major version.
- `y` is the minor version.
- `z` is the feature band.
- `nn` is the patch version.

`5.0.100` is valid for the local installed sdks `5.0.101`, `5.0.102`, `5.0.103`, ... **but** not for `>= 5.0.200`.
I hope that helps 😅.

prskr: Thanks for clarification!
Looking back it feels kinda stupid expecting Microsoft to stick to something like semantic versioning - of course there's more! :roll_eyes: but at least I do understand now why I broke stuff...

prskr:
Tests are still missing at this point but I thought it might be good to first check if the design is okay for you before putting too much effort into something that does not fit the overall architecture.

I actually ditched the whole x509 NuGet package and implemented the whole thing myself after having a look at the original implementation which isn't flexible enough to allow all different variants that are actually possible - or I completely missed something but I think the current solution fits at least okay-ish into the project, covers all cases and is still flexible enough to be extended later on if necessary.

I won't be able to continue on this before next week and if you haven't had time till then to look into it I'll continue to write some tests anyway to see if I missed something :see_no_evil: hence - no reason to hurry! :blush: 
Thanks for all your effort!

src/DotNet.Testcontainers/Containers/Configurations/DockerClientConstants.cs
```diff
@@ -0,0 +1,9 @@
+namespace DotNet.Testcontainers.Containers.Configurations
+{
+  internal static class DockerClientConstants
```
prskr, line 3: TODO: check if these can be moved to `DockerCertificatesDirectory`

HofmeisterAn:
> Tests are still missing at this point but I thought it might be good to first check if the design is okay for you before putting too much effort into something that does not fit the overall architecture.
> 
> I actually ditched the whole x509 NuGet package and implemented the whole thing myself after having a look at the original implementation which isn't flexible enough to allow all different variants that are actually possible - or I completely missed something but I think the current solution fits at least okay-ish into the project, covers all cases and is still flexible enough to be extended later on if necessary.
> 
> I won't be able to continue on this before next week and if you haven't had time till then to look into it I'll continue to write some tests anyway to see if I missed something 🙈 hence - no reason to hurry! 😊
> Thanks for all your effort!

Sorry – I had a very busy week. I needed some break from coding. I will look at it next week during the holidays.

.editorconfig
```diff
@@ -196,6 +196,10 @@ csharp_prefer_static_local_function = true:warning
 csharp_prefer_simple_using_statement = true:suggestion
 dotnet_diagnostic.IDE0063.severity = suggestion
 
+resharper_csharp_space_in_singleline_method = true
+resharper_csharp_space_between_accessors_in_singleline_property = true
+resharper_csharp_space_within_single_line_array_initializer_braces = true
```
HofmeisterAn, line 201: The end of `.editorconfig` contains a dedicated block for ReSharper / Rider:

```
##########################################
# JetBrains Rider
##########################################
```

src/DotNet.Testcontainers.Tests/Unit/Containers/DockerCertDirTest.cs
```diff
@@ -0,0 +1,23 @@
+namespace DotNet.Testcontainers.Tests.Unit.Containers
+{
+  using System.IO;
+  using Testcontainers.Containers.Configurations;
```
HofmeisterAn, line 4: Full namespace.

src/DotNet.Testcontainers.Tests/Unit/Containers/DockerClientConfigurationTest.cs
```diff
@@ -0,0 +1,34 @@
+namespace DotNet.Testcontainers.Tests.Unit.Containers
+{
+  using System;
+  using Testcontainers.Containers.Configurations;
```
HofmeisterAn, line 4: Full namespace. Haven't reviewed anything else, test class is still in progress.

src/DotNet.Testcontainers.Tests/Unit/Containers/TlsCredentialsTest.cs
```diff
@@ -0,0 +1,17 @@
+namespace DotNet.Testcontainers.Tests.Unit.Containers
+{
+  using System.IO;
+  using Testcontainers.Containers.Configurations;
```
HofmeisterAn, line 4: Full namespace.

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -97,12 +97,7 @@ public async Task AttachAsync(string id, IOutputConsumer outputConsumer, Cancell
     {
       Logger.LogInformation("Attaching {outputConsumer} at container {id}", outputConsumer.GetType(), id);
 
-      var attachParameters = new ContainerAttachParameters
-      {
-        Stdout = true,
-        Stderr = true,
-        Stream = true,
-      };
+      var attachParameters = new ContainerAttachParameters { Stdout = true, Stderr = true, Stream = true };
```
HofmeisterAn, line 100: Remove unnecessary changes (formatting).

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -121,8 +116,9 @@ public async Task<long> ExecAsync(string id, IList<string> command, Cancellation
       await this.Docker.Exec.StartContainerExecAsync(created.ID, ct)
         .ConfigureAwait(false);
 
-      for (ContainerExecInspectResponse response; (response = await this.Docker.Exec.InspectContainerExecAsync(created.ID, ct)
-        .ConfigureAwait(false)) != null;)
+      for (ContainerExecInspectResponse response;
```
HofmeisterAn, line 119: Remove unnecessary changes (formatting).

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -139,9 +135,9 @@ public async Task<string> RunAsync(ITestcontainersConfiguration configuration, C
 
       var hostConfig = new HostConfig
       {
-        // AutoRemove = configuration.CleanUp, TODO: Should we keep this? If the Docker daemon remove containers we're no longer able to call e. g. `CleanUp(true)` + `GetExitCode()`.
-        PortBindings = converter.PortBindings,
-        Mounts = converter.Mounts,
+        // TODO: Should we keep this? If the Docker daemon remove containers we're no longer able to call e. g. `CleanUp(true)` + `GetExitCode()`.
+        // AutoRemove = configuration.CleanUp,
+        PortBindings = converter.PortBindings, Mounts = converter.Mounts
```
HofmeisterAn, line 140:
```
PortBindings = converter.PortBindings,
Mounts = converter.Mounts
```

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -7,11 +7,13 @@ namespace DotNet.Testcontainers.Containers.Builders
   using System.Reflection;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet;
   using DotNet.Testcontainers.Clients;
   using DotNet.Testcontainers.Containers.Configurations;
   using DotNet.Testcontainers.Containers.OutputConsumers;
   using DotNet.Testcontainers.Containers.WaitStrategies;
   using DotNet.Testcontainers.Images;
+  using DockerClientConfiguration = Configurations.DockerClientConfiguration;
```
HofmeisterAn, line 16: Remove unnecessary change.

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -170,20 +173,33 @@ public ITestcontainersBuilder<TDockerContainer> WithMount(string source, string
     /// <inheritdoc />
     public ITestcontainersBuilder<TDockerContainer> WithCleanUp(bool cleanUp)
     {
-      return Build(this, Apply(cleanUp: cleanUp))
-        .WithLabel(TestcontainersClient.TestcontainersCleanUpLabel, cleanUp.ToString().ToLowerInvariant());
+      return Build(this, Apply(cleanUp: cleanUp)).WithLabel(TestcontainersClient.TestcontainersCleanUpLabel, cleanUp.ToString().ToLowerInvariant());
```
HofmeisterAn, line 176: Remove unnecessary changes (formatting).

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -306,7 +322,7 @@ public TDockerContainer Build()
     }
 
     /// <summary>
-    /// Returns the changed Testcontainer configuration object. If there is no change, the previous Testcontainer configuration object is returned.
+    ///   Returns the changed Testcontainer configuration object. If there is no change, the previous Testcontainer configuration object is returned.
```
HofmeisterAn, line 325: Remove unnecessary changes (formatting).

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -320,7 +336,7 @@ private static T Merge<T>(T next, T previous, T defaultConfiguration = null)
     }
 
     /// <summary>
-    /// Merges all existing and new Testcontainer configuration changes. If there are no changes, the previous Testcontainer configurations are returned.
+    ///   Merges all existing and new Testcontainer configuration changes. If there are no changes, the previous Testcontainer configurations are returned.
```
HofmeisterAn, line 339: Remove unnecessary changes (formatting).

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -333,14 +349,12 @@ private static IEnumerable<T> Merge<T>(IEnumerable<T> next, IEnumerable<T> previ
...
-      {
-        return next.Concat(previous).ToArray();
-      }
+
+      return next.Concat(previous).ToArray();
     }
 
     /// <summary>
-    /// Merges all existing and new Testcontainer configuration changes. If there are no changes, the previous Testcontainer configurations are returned.
+    ///   Merges all existing and new Testcontainer configuration changes. If there are no changes, the previous Testcontainer configurations are returned.
```
HofmeisterAn, line 357: Remove unnecessary changes (formatting).

src/DotNet.Testcontainers/Containers/Configurations/DockerClientEnvironmentConfiguration.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// Initializes a new instance of the <see cref="DockerClientEnvironmentConfiguration" /> class.
+    /// </summary>
+    /// <param name="clientEndpoint">The Docker client endpoint.</param>
+    /// <param name="certificatesDirectory">The TLS certificates base directory.</param>
+    /// <param name="isTlsVerificationEnabled">Is TLS verification enabled.</param>
+    public DockerClientEnvironmentConfiguration(Uri clientEndpoint, string certificatesDirectory, bool isTlsVerificationEnabled)
+    {
+      this.Endpoint = clientEndpoint;
+      var dockerCertificates = new DockerCertificatesDirectory(certificatesDirectory);
+      this.Credentials = new TlsCredentials(dockerCertificates.CaCertificate, dockerCertificates.ClientCertificate, isTlsVerificationEnabled, isTlsVerificationEnabled);
```
HofmeisterAn, line 28: Haven't tested it yet, but does it work if e. g. only `DOCKER_HOST` is set?

src/DotNet.Testcontainers/Containers/Builders/TestcontainersBuilder.cs
```diff
@@ -170,20 +173,33 @@ public ITestcontainersBuilder<TDockerContainer> WithMount(string source, string
...
-        .WithLabel(TestcontainersClient.TestcontainersCleanUpLabel, cleanUp.ToString().ToLowerInvariant());
+      return Build(this, Apply(cleanUp: cleanUp)).WithLabel(TestcontainersClient.TestcontainersCleanUpLabel, cleanUp.ToString().ToLowerInvariant());
     }
 
     /// <inheritdoc />
-    public ITestcontainersBuilder<TDockerContainer> WithDockerEndpoint(string endpoint)
+    public ITestcontainersBuilder<TDockerContainer> WithDockerEndpoint(string clientEndpoint, bool tlsVerify = true)
     {
-      return Build(this, Apply(endpoint: new Uri(endpoint)));
+      var endpoint = new Uri(clientEndpoint);
```
HofmeisterAn, line 182: Maybe put both codes to create `clientAuthConfig` into a static method.

HofmeisterAn (changes requested):
> Tests are still missing at this point but I thought it might be good to first check if the design is okay for you before putting too much effort into something that does not fit the overall architecture.

I think the overall implementation looks good! This allows different implementations of `IDockerClientConfiguration` in the future too. Just a few code formatting and consistency issues are left (I stopped at some point 😀). I can take care of them as soon as the implementation is ready.

**Edit**: Maybe it's better to split the pull request into at least two parts. In the first pull request we prepare the `IDockerClientConfiguration` and update the current implementation to use the interface. In the next pull requests we implement the different `Credentials`.

HofmeisterAn:
I am closing this PR due to inactivity.

prskr:
Yeah the use case I have/had got postponed hence I prioritized other things over this.
Is this still of interest or is there no other one requiring this? :sweat_smile:

HofmeisterAn:
> Is this still of interest or is there no other one requiring this? 😅

I think it's still necessary, but as long as no one is working on this feature, it doesn't need to be open. It's probably better to slice it into smaller chunks, too. First part could be just the interface to pass the necessary properties to the client, etc.

PR #377, opened 2021-03-29
CreateTopic method and tests added

ErenArslan:


HofmeisterAn (commented):
Thanks for you pull request. IMO module specific implementations shouldn't be part of this library. I know we did this in the past, but I would like to avoid it in the future and suggest creating a module specific package. This will reduce the complexity a lot.

If we need any changes to support module specific package, I'm happy to help.
If you need changes urgent, we can discuss about it in an issue.

@mustafaonuraydin what do you think?

mustafaonuraydin:
Hello @ErenArslan and @HofmeisterAn ,

In my opinion we can consider to module specific implementations when these functions are impossible to do with client libraries. For example you can not create Couchbase bucket via couchbase dotnet sdk. In this way, for applying the integration test like production environment we can consider to implement some function like CreateBucket.

My question, is it possible to create topic via Confluent.Kafka client library? If it's impossible, this pull request is acceptable in my opinion.

andreycha:
> My question, is it possible to create topic via Confluent.Kafka client library? If it's impossible, this pull request is acceptable in my opinion.

Yes, it is possible to create a topic using `AdminClient`, see [example](https://github.com/confluentinc/confluent-kafka-dotnet/blob/b7b04fed82762c67c2841d7481eae59dee3e4e20/examples/AdminClient/Program.cs#L79) in the official repository.

HofmeisterAn (commented):
> If it's impossible, this pull request is acceptable in my opinion.

But even then a module specific package would be "cleaner". It's not necessary to ship the entire implementation for modules if you only need a fraction. I think it's even more reliable and secure to create module packages.


> Yes, it is possible to create a topic using `AdminClient`, see [example](https://github.com/confluentinc/confluent-kafka-dotnet/blob/b7b04fed82762c67c2841d7481eae59dee3e4e20/examples/AdminClient/Program.cs#L79) in the official repository.

Yep, I have checked that too.

PR #384, opened 2021-05-19
initial attempt to add networking

### LINKED ISSUE #324: support for --link while building container
Looking to achieve something like below,

`docker run --rm --name web2 --link db:db training/webapp env`

is "--link" supported with any of existing supported command in the library?

brethubbard:
Fixes #324 

I started approaching this a couple different ways but ended up here. I am definitely open to suggestions on how to improve this.


HofmeisterAn:
Thanks, four your pull request! Great feature 👍 . I'll look at it later that day.

HofmeisterAn (commented):
I created a pull request with some suggestions in your fork. Most parts look good. Only a few things are left. We can discuss my changes there.

I've added a builder to create the network configuration and changed `Tuple<string, string>` to an `IEnumerable<IDockerNetwork>` to support multiple networks.

Once we have finished the remaining todos we can merge this pr.

src/DotNet.Testcontainers/Clients/DockerNetworkOperations.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    public async Task<IEnumerable<NetworkResponse>> GetAllAsync(CancellationToken ct = default)
+    {
+      return (await this.Docker.Networks.ListNetworksAsync(new NetworksListParameters(), ct)
+        .ConfigureAwait(false)).ToArray();
+    }
+
+    public async Task<IEnumerable<NetworkResponse>> GetOrphanedObjects(CancellationToken ct = default)
+    {
+      var filters = new FilterByProperty {{"label", $"{TestcontainersClient.TestcontainersCleanUpLabel}=true"}, {"dangling", "true"}};
```
brethubbard: Just thinking about cleaning up resources from previous runs, we may need to think of something more holistic than this. Since the network is created before the containers are started, the containers will likely still be connected to the network so we they won't be dangling.

Because of this, we'll need to clean up the old containers before we clean up the old networks. Not sure how to approach this one. I'll think about it some more.

HofmeisterAn (commented):
- [x] Find a better name for `ExistingDockerNetwork` and `DockerNetwork`.
- [x] Fix `UtilityFunctions` (I don't lik "utility classes"; maybe move it to somewhere else and change it to an extension methode; I have to think about it)
- [x] Take a closer look into `removeOrphanedNetworksTasks`.
- [x] Add more unit tests.
- [x] Add missing documentation.

src/DotNet.Testcontainers/Clients/DockerNetworkOperations.cs
HofmeisterAn: Yeah, I haven't thought about that.... We probably can't implement it that way at all. A network resource has no status. We do not know at all when it's no longer needed.

src/DotNet.Testcontainers/Networks/Configurations/NetworkDriver.cs
```diff
@@ -0,0 +1,26 @@
...
+
+  /// <summary>
+  ///
+  /// </summary>
+  public readonly struct NetworkDriver
+  {
+    /// <summary>
+    ///
+    /// </summary>
+    public static readonly NetworkDriver Bridge = new NetworkDriver("bridge");
```
brethubbard, line 13: I wonder if we should add any other drivers like `overlay`, `host`, `macvlan`, or `none`.

brethubbard:
Thanks for continuing to work on this! Unfortunately I have a busy weekend coming up so I won't be able to contribute until next week. I'm happy to pick up any of the unfinished items then!

src/DotNet.Testcontainers/Networks/Configurations/NetworkDriver.cs
HofmeisterAn: Yep, I'll add them too.

HofmeisterAn:
> Thanks for continuing to work on this! Unfortunately I have a busy weekend coming up so I won't be able to contribute until next week. I'm happy to pick up any of the unfinished items then!

I can add the unfinished parts in the next days.

src/DotNet.Testcontainers/Clients/DockerNetworkOperations.cs
HofmeisterAn: I removed the part for now. I don't know how to remove orphaned networks properly. This is tricky. I think it requires an overall solution incl. containers.

src/DotNet.Testcontainers/Networks/Configurations/NetworkDriver.cs
HofmeisterAn: I've added `host` too. I think the other drivers require a bit more configuration.

HofmeisterAn (commented):
@brethubbard It would be kind if you can take a look at the last changes. After that I'll squash the commits and merge.

brethubbard:
That all looks good! The choice for adding only the `bridge` and `host` docker network makes sense as well as figuring out what docker cleanup looks like holistically. I wonder if this would be a good candidate for `ryuk`. ~I'm not sure if that manages only containers or networks as well. If it only manages containers, the cleanup of networks like we were doing would probably work since `ryuk` would clean up the containers automatically.~ It looks like `ryuk` can remove networks as well.

HofmeisterAn:
> It looks like `ryuk` can remove networks as well.

Yep, it can (and other resources too). Back then, mixed platforms weren't supported, or maybe I just didn't know about it. I need something that covers Windows containers too. Maybe I should take a closer look at `ryuk` again. Anyway, I'll squash and merge the pull request tomorrow. Thanks.

HofmeisterAn (approved):

Issue #390, opened 2021-08-15
Accessing the output of ExecAsync


fabiogouw:
**Is your feature request related to a problem? Please describe.**
I'm writing code to test an application against a database and I need to run some scripts before the test runs (create some tables, etc.). I was able to copy the script into the container (using CopyFileAsync method) and I'm trying to run it (using ExecAsync method), but I can't see the output (stdout and stderr) when running the command (to debug and see errors if they happen), only the result code.

**Describe the solution you'd like**
I'd like to have some way to see what happened during the command execution. The ExecAsync could return an object that has this information, not only the return code.
The Java version of TestContainers has the [Container.ExecResult](https://www.testcontainers.org/features/commands/) object, I guess we might have it here too.

**Describe alternatives you've considered**
I considered to run the scripts outside TestContainers, using plain ADO.NET commands, but I think it would be nice if we can do it using the framework (maybe adding some extension methods to do it in the databases modules).

HofmeisterAn:
👍 I think we just need to pass a stream to:

https://github.com/HofmeisterAn/dotnet-testcontainers/blob/8b3860c0bb27d8f82f747fdba07870ef33a18bad/src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs#L117

Either we use the output consumer or create a scoped stream and return a dedicated object with the output result and exit code.

fabiogouw:
Andre, I took your second suggestion and it worked, I can make a PR with this change. But before, I'd like to discuss two options with you:

1. Create a new method (for instance, _ExecWithOutputAsync_) that returns the new object (exit code + stdout + stderr), similar to the solution in the Java version of this lib
2. Alter the existing method _ExecAsync_ to return this object. This will break existing clients.

HofmeisterAn:
Thanks. I'll take a look at it later that day.

HofmeisterAn:
`1.6.0-beta.1966`

PR #392, opened 2021-08-19
Add exec function that returns the stdout and stderr of the container

fabiogouw:
This PR address the issue #390 , I'm taking the approach that creates a new method so no usage is broken. Please let me know if it's ok or if it needs any kind of improvement. Thank you.

### DIFF
--- a/src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
+++ b/src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
@@ -8,6 +8,7 @@ namespace DotNet.Testcontainers.Clients
   using System.Threading.Tasks;
   using Docker.DotNet.Models;
   using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Containers;
   using Microsoft.Extensions.Logging;
 
   internal sealed class DockerContainerOperations : DockerApiClient, IDockerContainerOperations
@@ -110,31 +111,38 @@ namespace DotNet.Testcontainers.Clients
         .ConfigureAwait(false);
     }
 
-    public async Task<long> ExecAsync(string id, IList<string> command, CancellationToken ct = default)
+    public async Task<ExecResult> ExecAsync(string id, IList<string> command, CancellationToken ct = default)
     {
       this.logger.LogInformation("Executing {command} at container {id}", command, id);
 
       var execCreateParameters = new ContainerExecCreateParameters
       {
         Cmd = command,
+        AttachStdout = true,
+        AttachStderr = true,
       };
 
-      var created = await this.Docker.Exec.ExecCreateContainerAsync(id, execCreateParameters, ct)
+      var execCreateResponse = await this.Docker.Exec.ExecCreateContainerAsync(id, execCreateParameters, ct)
         .ConfigureAwait(false);
 
-      await this.Docker.Exec.StartContainerExecAsync(created.ID, ct)
+      var stdOutAndErrStream = await this.Docker.Exec.StartAndAttachContainerExecAsync(execCreateResponse.ID, false, ct)
         .ConfigureAwait(false);
 
-      for (ContainerExecInspectResponse response; (response = await this.Docker.Exec.InspectContainerExecAsync(created.ID, ct)
+      for (ContainerExecInspectResponse response; (response = await this.Docker.Exec.InspectContainerExecAsync(execCreateResponse.ID, ct)
         .ConfigureAwait(false)) != null;)
       {
-        if (!response.Running)
+        if (response.Running)
         {
-          return response.ExitCode;
+          continue;
         }
+
+        var (stdout, stderr) = await stdOutAndErrStream.ReadOutputToEndAsync(ct)
+          .ConfigureAwait(false);
+
+        return new ExecResult(stdout, stderr, response.ExitCode);
       }
 
-      return long.MinValue;
+      return ExecResult.Failure;
     }
 
     public async Task<string> RunAsync(ITestcontainersConfiguration configuration, CancellationToken ct = default)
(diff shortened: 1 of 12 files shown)


src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -137,6 +137,38 @@ await this.Docker.Exec.StartContainerExecAsync(created.ID, ct)
...
+      {
+        Cmd = command,
+        AttachStderr = true,
+        AttachStdout = true,
+      };
+
+      var created = await this.Docker.Exec.ExecCreateContainerAsync(id, execCreateParameters, ct)
+        .ConfigureAwait(false);
+
+      var stream = await this.Docker.Exec.StartAndAttachContainerExecAsync(created.ID, false, ct);
```
HofmeisterAn: `.ConfigureAwait(false);`

src/DotNet.Testcontainers/Clients/ExecResult.cs
```diff
@@ -0,0 +1,11 @@
+namespace DotNet.Testcontainers.Clients
```
HofmeisterAn: Move the file to `src\DotNet.Testcontainers\Containers\`, namespace `DotNet.Testcontainers.Containers`.

src/DotNet.Testcontainers/Clients/ExecResult.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+  public class ExecResult
```
HofmeisterAn: Please use constructor initialization.

```cs
public readonly struct ExecResult
{
  public ExecResult(string stdOut, string stdErr, long exitCode)
  {
    this.StdOut = stdOut;
    this.StdErr = stdErr;
    this.ExitCode = exitCode;
  }

  public string StdOut { get; }

  public string StdErr { get; }

  public long ExitCode { get; }
}
```

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -137,6 +137,38 @@ await this.Docker.Exec.StartContainerExecAsync(created.ID, ct)
...
+      var results = new ExecResult();
+
+      for (ContainerExecInspectResponse response; (response = await this.Docker.Exec.InspectContainerExecAsync(created.ID, ct)
+        .ConfigureAwait(false)) != null;)
+      {
+        if (!response.Running)
+        {
+          (results.StdOut, results.StdErr) = await stream.ReadOutputToEndAsync(ct);
+          results.ExitCode = response.ExitCode;
+          break;
```
HofmeisterAn: `return new ExecResult(stdOut, stdErr, response.ExitCode);` see `ExecResult` suggestions.

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -137,6 +137,38 @@ await this.Docker.Exec.StartContainerExecAsync(created.ID, ct)
...
+        }
+      }
+
+      return results;
```
HofmeisterAn: `return new ExecResult(string.Empty, string.Empty, long.MinValue)`.

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -137,6 +137,38 @@ await this.Docker.Exec.StartContainerExecAsync(created.ID, ct)
...
+      var stream = await this.Docker.Exec.StartAndAttachContainerExecAsync(created.ID, false, ct);
+
+      var results = new ExecResult();
+
+      for (ContainerExecInspectResponse response; (response = await this.Docker.Exec.InspectContainerExecAsync(created.ID, ct)
+        .ConfigureAwait(false)) != null;)
+      {
+        if (!response.Running)
+        {
+          (results.StdOut, results.StdErr) = await stream.ReadOutputToEndAsync(ct);
```
HofmeisterAn: `.ConfigureAwait(false);`

HofmeisterAn (changes requested):
PR looks good! Thanks. I suggest only a few small changes. If you don't mind, I think we should use the existing `ExecAsync` method. Even it is a breaking change, I think it's better to stay close to the Java implementation. Version 1.6 has breaking changes anyway (renaming namespaces).

Thanks again.

src/DotNet.Testcontainers/Clients/DockerContainerOperations.cs
fabiogouw: Done!

fabiogouw:
Thanks for your considerations, @HofmeisterAn , I made all the changes. However, the pipeline now hangs in the Tests stage, I'll take a look at the root cause.

HofmeisterAn:
> Thanks for your considerations, @HofmeisterAn , I made all the changes. However, the pipeline now hangs in the Tests stage, I'll take a look at the root cause.

It's because the wait strategies never return, see:

https://github.com/HofmeisterAn/dotnet-testcontainers/blob/8b3860c0bb27d8f82f747fdba07870ef33a18bad/src/DotNet.Testcontainers/Configurations/WaitStrategies/UntilUnixCommandIsCompleted.cs#L29

We need to double check the `ExecAsync` calles. I'll take a look at it tomorrow, but I'm pretty sure there are not many left.

fabiogouw:
> > Thanks for your considerations, @HofmeisterAn , I made all the changes. However, the pipeline now hangs in the Tests stage, I'll take a look at the root cause.
> 
> It's because the wait strategies never return, see:
> 
> https://github.com/HofmeisterAn/dotnet-testcontainers/blob/8b3860c0bb27d8f82f747fdba07870ef33a18bad/src/DotNet.Testcontainers/Configurations/WaitStrategies/UntilUnixCommandIsCompleted.cs#L29
> 
> We need to double check the `ExecAsync` calles. I'll take a look at it tomorrow, but I'm pretty sure there are not many left.

Great, that was the problem! Thanks

HofmeisterAn (approved):

Issue #399, opened 2021-09-28
Helper methods for database script execution


fabiogouw:
**Is your feature request related to a problem? Please describe.**
If I need to run scripts to prepare the database before running tests against it, then I must code all the steps:
- by using the database API to execute the script
- or by copying the file inside the container and executing the command line tool (sqlcmd, mysql, ...)

**Describe the solution you'd like**
A good way to make my test code smaller is creating a helper method in the database modules:

```csharp
var results = await dbTestContainer.ExecScript("CREATE DATABASE xzy; ...");
```

This `ExecScript` method can encapsulate the task of creating the script file, copying it inside the container and running the command line tool to execute it.

HofmeisterAn:
> to prepare the database before running tests

Shouldn't that be part of the Docker image?

> This ExecScript method can encapsulate the task of creating the script file, copying it inside the container and running the command line tool to execute it.

Sounds good as long as it covers Windows containers (PowerShell scripts, etc.) too.

fabiogouw:
> Shouldn't that be part of the Docker image?]

This is an approach, but I'm looking for scenarios where it's easier to prepare the content of the container instead building a new image for each test case. Let's say we're building a data access component and we want to validate the behavior when the component uses an invalid user, user with wrong permissions, access a table that doesn't exists in the MySQL database. Instead of having an image for each case, it's easier to run a script to give wrong permissions to the user before the test.

> Sounds good as long as it covers Windows containers (PowerShell scripts, etc.) too.

Hm, I was thinking in having this feature only for specific databases modules (at least for now), not for generic images: running SQL scripts against MS SQL, MySQL, Oracle, Postgre; Lua scripts for Redis; N1QL for Couchbase. I've created a pull request #401 so we can easily discuss about this feature

HofmeisterAn:
> Hm, I was thinking in having this feature only for specific databases modules (at least for now), not for generic images: running SQL scripts against MS SQL, MySQL, Oracle, Postgre; Lua scripts for Redis; N1QL for Couchbase. I've created a pull request #401 so we can easily discuss about this feature

Ah I see. I was thinking about a "more" generic implementation like `IDockerContainer.ExecAsync`. Where you pass the content of the script and the executable to the function.

HofmeisterAn:
#401

PR #401, opened 2021-09-30
[#399] #IMPLEMENT Helper methods for database script execution

fabiogouw:
Methods to run scripts against databases:

- SQL scripts for MS SQL, MySQL, Oracle, Postgre;
- Lua scripts for Redis; 
- N1QL scripts for Couchbase

### DIFF
--- a/tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CouchbaseTestcontainerTest.cs
+++ b/tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CouchbaseTestcontainerTest.cs
@@ -233,6 +233,25 @@ namespace DotNet.Testcontainers.Tests.Unit
       Assert.Equal("Couchbase ClusterAnalyticsRamSize ram size can not be less than 1024 MB. (Parameter 'ClusterAnalyticsRamSize')", exception.Message);
     }
 
+    [Fact]
+    public async Task ExecScriptInRunningContainer()
+    {
+      // Given
+      _ = await this.couchbaseFixture.Container.CreateBucket("MyBucket")
+        .ConfigureAwait(false);
+      const string script = @"
+        CREATE PRIMARY INDEX ON `MyBucket`;
+        SELECT * FROM system:indexes;
+        ";
+
+      // When
+      var results = await this.couchbaseFixture.Container.ExecScriptAsync(script)
+        .ConfigureAwait(false);
+
+      // Then
+      Assert.Contains("MyBucket", results.Stdout);
+    }
+
     private readonly struct Customer
     {
       public Customer(string name, int age)
--- a/src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
+++ b/src/DotNet.Testcontainers/Containers/Modules/Databases/CouchbaseTestcontainer.cs
@@ -1,5 +1,6 @@
 namespace DotNet.Testcontainers.Containers
 {
+  using System.Text;
   using System.Threading.Tasks;
   using DotNet.Testcontainers.Configurations;
   using JetBrains.Annotations;
@@ -49,6 +50,19 @@ namespace DotNet.Testcontainers.Containers
     {
       var flushBucketCommand = $"yes | {CouchbaseCli} bucket-flush -c 127.0.0.1:8091 --username {this.Username} --password {this.Password} --bucket {bucket}";
       return this.ExecAsync(new[] { "/bin/sh", "-c", flushBucketCommand });
+    }
+
+    /// <summary>
+    /// Executes a N1QL script in the database container.
+    /// </summary>
+    /// <param name="scriptContent">The content of the N1QL script to be executed.</param>
+    /// <returns>Task that completes when the script has been executed.</returns>
+    public override async Task<ExecResult> ExecScriptAsync(string scriptContent)
+    {
+      var tempScriptFile = this.GetTempScriptFile();
+      await this.CopyFileAsync(tempScriptFile, Encoding.ASCII.GetBytes(scriptContent));
+      var execScriptCommand = $"cbq -user {this.Username} -password {this.Password} -file={tempScriptFile}";
+      return await this.ExecAsync(new[] { "/bin/sh", "-c", execScriptCommand });
     }
   }
 }
--- a/src/DotNet.Testcontainers/Containers/Modules/Databases/MsSqlTestcontainer.cs
+++ b/src/DotNet.Testcontainers/Containers/Modules/Databases/MsSqlTestcontainer.cs
@@ -1,5 +1,7 @@
 namespace DotNet.Testcontainers.Containers
 {
+  using System.Text;
+  using System.Threading.Tasks;
   using DotNet.Testcontainers.Configurations;
   using JetBrains.Annotations;
   using Microsoft.Extensions.Logging;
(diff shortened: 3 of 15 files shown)


HofmeisterAn:
I'll do the review in the next days. Did you check the test issue `ShouldExecuteCommandsWhenScriptExecuted`?

fabiogouw:
I'm still working on it, Andre, trying to figure it out.

fabiogouw:
> I'm still working on it, Andre, trying to figure it out.

I guess it's ok now!

HofmeisterAn:
> I guess it's ok now!

👍  I'll do the review end of next week. Unfortunately, I'm not able to do it earlier. Thanks.

tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CouchbaseTestcontainerTest.cs
```diff
@@ -233,6 +233,25 @@ public void ShouldThrowArgumentOutOfRangeExceptionWhenClusterAnalyticsRamSizeIsN
       Assert.Equal("Couchbase ClusterAnalyticsRamSize ram size can not be less than 1024 MB. (Parameter 'ClusterAnalyticsRamSize')", exception.Message);
     }
 
+    [Fact]
+    public async Task ShouldExecuteCommandsWhenScriptExecuted()
```
HofmeisterAn: A consistent test method name (for all script tests) would be nice. What about something like `ExecScriptInRunningContainer`?

tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CouchbaseTestcontainerTest.cs
```diff
@@ -233,6 +233,25 @@ public void ShouldThrowArgumentOutOfRangeExceptionWhenClusterAnalyticsRamSizeIsN
...
+    {
+      // Given
+      _ = await this.couchbaseFixture.Container.CreateBucket("MyBucket")
+        .ConfigureAwait(false);
+      var script = @"
```
HofmeisterAn: Please use `const string script` instead of `var script`, other tests too.

src/DotNet.Testcontainers/Containers/Modules/IScriptExecution.cs
```diff
@@ -0,0 +1,19 @@
+namespace DotNet.Testcontainers.Containers.Modules
+{
+  using System.Threading.Tasks;
+  using JetBrains.Annotations;
+
+  /// <summary>
+  /// Define the contract for database modules that has the script execution capability.
+  /// </summary>
+  public interface IScriptExecution
```
HofmeisterAn: What about `IDatabaseScript`?

src/DotNet.Testcontainers/Containers/Modules/IScriptExecution.cs
```diff
@@ -0,0 +1,19 @@
...
+  {
+    /// <summary>
+    /// Executes a script in the database container.
+    /// </summary>
+    /// <param name="scriptContent">The content of the script to be executed.</param>
+    /// <returns>Task that completes when the script has been executed.</returns>
+    [PublicAPI]
+    Task<ExecResult> ExecScript(string scriptContent);
```
HofmeisterAn: `ExecScriptAsync`

src/DotNet.Testcontainers/Containers/Modules/IScriptExecution.cs
```diff
@@ -0,0 +1,19 @@
...
+  using System.Threading.Tasks;
+  using JetBrains.Annotations;
+
+  /// <summary>
+  /// Define the contract for database modules that has the script execution capability.
+  /// </summary>
+  public interface IScriptExecution
+  {
+    /// <summary>
+    /// Executes a script in the database container.
```
HofmeisterAn: `Executes a database script in the running Testcontainer.`

HofmeisterAn (changes requested):
Shouldn't we add the interface `IScriptExecution` to the base class `TestcontainerDatabase`? If the module does not support database scripts (or it has no implementation yet) we can simply throw `NotImplementedException`.

I suggest something like this:

```cs
// TestcontainerDatabase.cs

public abstract class TestcontainerDatabase : HostedServiceContainer, IDatabaseScript
{
  protected TestcontainerDatabase(ITestcontainersConfiguration configuration, ILogger logger)
    : base(configuration, logger)
  {
  }

  // ...

  public virtual string GetTempScriptFile()
  {
    return Path.Combine("/tmp", Guid.NewGuid().ToString("N"), ".tmp");
  }

  public virtual Task<ExecResult> ExecScriptAsync(string script)
  {
    throw new NotImplementedException();
  }
}
```

Then each supported module can override `ExecScriptAsync`, e. g.:

```cs
// PostgreSqlTestcontainer.cs

public override async Task<ExecResult> ExecScriptAsync(string script)
{
  var tempScriptFile = this.GetTempScriptFile();

  await this.CopyFileAsync(tempScriptFile, Encoding.UTF8.GetBytes(script))
    .ConfigureAwait(false);

  return await this.ExecAsync(new[] { "psql", "-U", this.Username, "-d", this.Database, "-a", "-f", tempScriptFile })
    .ConfigureAwait(false);
}
```

Everything else looks good 👍 . Thanks for the pull request and contribution.

fabiogouw:
> Shouldn't we add the interface `IScriptExecution` to the base class `TestcontainerDatabase`? If the module does not support database scripts (or it has no implementation yet) we can simply throw `NotImplementedException`.
> 
> I suggest something like this:
> 
> ```cs
> // TestcontainerDatabase.cs
> 
> public abstract class TestcontainerDatabase : HostedServiceContainer, IDatabaseScript
> {
>   protected TestcontainerDatabase(ITestcontainersConfiguration configuration, ILogger logger)
>     : base(configuration, logger)
>   {
>   }
> 
>   // ...
> 
>   public virtual string GetTempScriptFile()
>   {
>     return Path.Combine("/tmp", Guid.NewGuid().ToString("N"), ".tmp");
>   }
> 
>   public virtual Task<ExecResult> ExecScriptAsync(string script)
>   {
>     throw new NotImplementedException();
>   }
> }
> ```
> 
> Then each supported module can override `ExecScriptAsync`, e. g.:
> 
> ```cs
> // PostgreSqlTestcontainer.cs
> 
> public override async Task<ExecResult> ExecScriptAsync(string script)
> {
>   var tempScriptFile = this.GetTempScriptFile();
> 
>   await this.CopyFileAsync(tempScriptFile, Encoding.UTF8.GetBytes(script))
>     .ConfigureAwait(false);
> 
>   return await this.ExecAsync(new[] { "psql", "-U", this.Username, "-d", this.Database, "-a", "-f", tempScriptFile })
>     .ConfigureAwait(false);
> }
> ```
> 
> Everything else looks good 👍 . Thanks for the pull request and contribution.

@HofmeisterAn, thanks for your comments, I changed the code so most of them are implemented now, but there's one I'd like to discuss further.

When I was developing this code, my first thought was to change the abstract class TestcontainerDatabase so it could have the ExecScript method, like your suggestion. But I've found out that not all databases have the scripting capability (in the case, CouchDB). So it was strange to me having the CouchDbTestcontainer class with this method, even throwing the NotImplemented exception, so I decided to implement the IScriptExecution interface (now IDatabaseScript, I changed its name) on just the databases that have this feature (most of them, in fact). I think that if a class implements a given interface, it has to implement the feature.

One alternative to this case, for the databases that doesn't have a native script capability, is to let it run generic bash scripts. At least it would work for CouchDB, since it can be manipulated with http call (curl).

What do you think?

HofmeisterAn (commented):
> One alternative to this case, for the databases that doesn't have a native script capability, is to let it run generic bash scripts. At least it would work for CouchDB, since it can be manipulated with http call (curl).

I thought of that, too. If a database does not have native script capabilities `sh` would be helpful. The modules should be moved into their own dependencies anyway. Then they can provide database specific features much easier. As long as this is not the case I think it's easier to maintain from the base class.

> even throwing the NotImplemented exception

I think it's not that bad. If someone needs the script capability then they can always create a pull request. Modules are very simple right now.

fabiogouw:
Sounds ok to me! I'll implement the `sh` capability in the base class and move the reference to the IDatabaseScript interface to the base class.

HofmeisterAn (approved):
👍 Thanks, I'll merge it ~~tomorrow~~.

**Edit**:
Sorry - I can't do it today. Tomorrow, I promise 🤞.

HofmeisterAn:
`1.6.0-beta.2028`

fabiogouw:
Hello, @HofmeisterAn ! I've noticed that you released a new version of the lib (1.6.0). Is this PR included?

HofmeisterAn:
> Hello, @HofmeisterAn ! I've noticed that you released a new version of the lib (1.6.0). Is this PR included?

Yes, it is. I'm sorry — I forgot to add it to the release notes. The issue wasn't assigned to the milestone. I'll update the information later that day.

https://github.com/testcontainers/testcontainers-dotnet/blob/2aeec1e9b88b8580c9914b73bee56583ad005e03/src/DotNet.Testcontainers/Containers/Modules/IDatabaseScript.cs#L19

https://github.com/testcontainers/testcontainers-dotnet/blob/2aeec1e9b88b8580c9914b73bee56583ad005e03/src/DotNet.Testcontainers/Containers/Modules/Databases/MsSqlTestcontainer.cs#L34

fabiogouw:
Great, thanks!
There's also this PR #392, if you could mention it too, the release notes description will be more complete.

PR #413, opened 2021-12-21
Add support for ryuk resource reaper

### LINKED ISSUE #242: Force Clean Containers on build before start

PSanetra:
This PR adds support for the [ryuk resource reaper](https://github.com/testcontainers/moby-ryuk).

Features
* Starts a separate ryuk container and registers a session id label (`dotnet.testcontainers.resource-reaper-session`)
* Ryuk deletes containers, networks, images and volumes with this label automatically after 10 seconds when the TCP connection to the test runner got lost or if the ryuk container is explicitly stopped (e.g. by disposing the `ResourceReaper` instance). The TCP connection is not lost during halting on a break point.
* A default resource reaper is used on containers, images, volumes and networks if no separate resource reaper session id is specified.
* Adds support for AutoRemove (equivalent to the `--rm` flag on `docker run`) on TestContainers. This is needed to remove the ryuk container itself
* Adds support for privileged containers (not sure anymore why this is included in this PR. Was working already a while on this...)

Fixes #242


src/DotNet.Testcontainers/Containers/Configurations/Misc/ResourceReaperContainerConfiguration.cs
```diff
@@ -0,0 +1,77 @@
...
+      this.Image = image;
+      this.Port = port;
+      this.DefaultPort = defaultPort;
+      this.WaitStrategy = new WaitForContainerUnix().UntilPortIsAvailable(defaultPort);
+    }
+
+    /// <summary>
+    /// Gets or sets the reaper container image.
+    /// Must be compatible to ryuk and support pruning on SIGINT and SIGTERM signals.
+    /// As of ryuk 0.3.3 pruning is not supported by the original ryuk image. By default the fork ghcr.io/psanetra/ryuk:2021.12.20 is used.
```
PSanetra: As you can see I am currently using a custom build of ryuk here. I think we should wait with merging this PR for the PR https://github.com/testcontainers/moby-ryuk/pull/22 to be merged, so that we can use the original ryuk image.

HofmeisterAn (changes requested):
Thanks for your work! I'm looking forward to merge this feature, this will help a lot.

- I suggest to change some parts of the `ResourceReaper` implementation. I think `ResourceReaperContainer`, `ResourceReaperContainerConfiguration` and `TestcontainersBuilderResourceReaperExtension` is not necessary. I attached a refactored class (not finished yet).
- To access the logger in `ResourceReaper`, we can use the logger field in `TestcontainersContainer` and add a getter.
- Log messages are now defined in `Logging.cs`.
- We need to keep an eye on Windows containers too (I think Windows containers and WSL will work, but Hyper-V not).

The PR is quite big to review. It would help a lot if you focus on the bare minimum and do not check-in unnessary changes like namespaces, etc.

```cs
public sealed class ResourceReaper : IAsyncDisposable
{
  private const string RyukImage = "ghcr.io/psanetra/ryuk:2021.12.20";

  private const int RyukPort = 3306;

  private static readonly ResourceReaper DefaultResourceReaper = new ResourceReaper();

  private readonly ITestcontainersContainer resourceReaperContainer;

  private ResourceReaper()
    : this(Guid.NewGuid())
  {
  }

  private ResourceReaper(Guid sessionId)
  {
    this.resourceReaperContainer = new TestcontainersBuilder<ITestcontainersContainer>()
      .WithName($"testcontainers-ryuk-{sessionId:D}")
      .WithImage(RyukImage)
      .WithAutoRemove(true)
      .WithCleanUp(false)
      .WithPortBinding(0, RyukPort)
      .WithBindMount("/var/run/docker.sock", "/var/run/docker.sock", AccessMode.ReadOnly)
      .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(RyukPort))
      .Build();

    this.SessionId = sessionId;
  }

  /// <summary>
  /// Gets the resource reaper session id.
  /// </summary>
  public Guid SessionId { get; }

  /// <summary>
  ///
  /// </summary>
  /// <param name="ct">Cancellation token.</param>
  /// <returns>Task that completes when the resource reaper has been started.</returns>
  public static async Task<ResourceReaper> GetAndStartDefaultAsync(CancellationToken ct = default)
  {
    // TODO: Return `DefaultResourceReaper` if resource reaper is up and running.
    await DefaultResourceReaper.StartAsync(ct)
      .ConfigureAwait(false);

    return DefaultResourceReaper;
  }

  /// <summary>
  ///
  /// </summary>
  /// <param name="ct">Cancellation token.</param>
  /// <returns>Task that completes when the resource reaper has been started.</returns>
  public static async Task<ResourceReaper> GetAndStartNewAsync(CancellationToken ct = default)
  {
    var resourceReaper = new ResourceReaper();

    await resourceReaper.StartAsync(ct)
      .ConfigureAwait(false);

    return resourceReaper;
  }

  /// <inheritdoc />
  public ValueTask DisposeAsync()
  {
    return this.resourceReaperContainer.DisposeAsync();
  }

  private Task StartAsync(CancellationToken ct = default)
  {
    return this.resourceReaperContainer.StartAsync(ct);
  }

  private Task StopAsync(CancellationToken ct = default)
  {
    return this.resourceReaperContainer.StopAsync(ct);
  }
}
```

Thanks again.

PSanetra:
@HofmeisterAn I have applied your suggestions and also did some additional improvements. I have added those changes in separate commits. Maybe this is easier to review and can be squashed before merging the PR.

Unfortunately I am not able to test this under windows.

HofmeisterAn:
> @HofmeisterAn I have applied your suggestions and also did some additional improvements. I have added those changes in separate commits. Maybe this is easier to review and can be squashed before merging the PR.
> 

Did you see my pr in your fork?

> Unfortunately I am not able to test this under windows.

I can do that after the holidays.

PSanetra:
@HofmeisterAn Sorry I was not watching the fork and did not receive a notification. I have merged your branch into this PR and resolved the conflicts. I think you should be able to push directly to the branch in my fork as I have enabled the `Allow edits by maintainers` option in this PR.

HofmeisterAn:
@PSanetra I'll start this week with the Windows tests, etc.

src/DotNet.Testcontainers/Services/ResourceReaper.cs
```diff
@@ -0,0 +1,219 @@
...
+
+              var stream = tcpClient.GetStream();
+
+              if (!ryukInitializedTaskSource.Task.IsCompleted)
+              {
+                await this.SendSessionLabel(stream, cancellationToken).ConfigureAwait(false);
+                ryukInitializedTaskSource.SetResult(true);
+              }
+
+              while (!cancellationToken.IsCancellationRequested)
```
HofmeisterAn: @PSanetra is that loop necessary to keep the connection open?

src/DotNet.Testcontainers/Services/ResourceReaper.cs
```diff
@@ -0,0 +1,219 @@
...
+          }
+          catch (OperationCanceledException e)
+          {
+            this.logger.ResourceReaperConnectionLostTrace();
+          }
+          catch (Exception e)
+          {
+            this.logger.ResourceReaperConnectionLostError();
+
+            await Task.Delay(TimeSpan.FromSeconds(1), cancellationToken).ConfigureAwait(false);
```
HofmeisterAn: Why do we need this delay?

tests/DotNet.Testcontainers.Tests/Fixtures/KeepTestcontainersUpAndRunning.cs
```diff
@@ -3,6 +3,14 @@ namespace DotNet.Testcontainers.Tests.Fixtures
   public static class KeepTestcontainersUpAndRunning
   {
     public static string[] Command { get; }
-      = { "/bin/sh", "-c", "tail -f /dev/null" };
+      =
```
HofmeisterAn, line 6: Why did you changed this?

HofmeisterAn: In my tests your change does not work. E. g. alpine exits.

HofmeisterAn:
@PSanetra somehow I can not push on your branch. __I created another pull request in your fork__. It would be nice if you review and merge it again. After that I'll take a look at the Windows containers. Maybe we need to add some minor fixes too, I did not run all unit tests 😬.

tests/DotNet.Testcontainers.Tests/Fixtures/KeepTestcontainersUpAndRunning.cs
PSanetra: Weird. I wanted to make the tests faster this way as `tail -f /dev/null` does not stop on SIGTERM and there is always quite a delay until docker decides to use SIGKILL to stop the process.

Is the following experiment reproducible on your machine with alpine?
```bash
$ docker run -d --rm --name test --entrypoint=/bin/sh alpine:3.15.0 '-c' 'trap "exit" TERM; while true; do sleep 1; done;'            
59a9e9f2cf193e224a6c45b189fe181eaaacce6815709531dbc878b850fb3602
$ docker ps
CONTAINER ID   IMAGE           COMMAND                   CREATED         STATUS         PORTS     NAMES
59a9e9f2cf19   alpine:3.15.0   "/bin/sh -c 'trap \"e…"   6 seconds ago   Up 5 seconds             test
$ docker stop test
test
$ docker ps
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
$
```

src/DotNet.Testcontainers/Services/ResourceReaper.cs
PSanetra: This is just in case if future ryuk decides to send more lines while the connection is open.

src/DotNet.Testcontainers/Services/ResourceReaper.cs
PSanetra: This delay throttles the connection attempts a little bit if there is some kind of endless loop if the connection can not be reestablished. Otherwise I could imagine that such an endless connection attempt loop could also eat some cpu.

PSanetra:
For some reason GitHub does not show my response to your comment `In my tests your change does not work. E. g. alpine exits.` in this conversation overview, so I repost that here:

Weird. I wanted to make the tests faster this way as `tail -f /dev/null` does not stop on SIGTERM and there is always quite a delay until docker decides to use SIGKILL to stop the process.

Is the following experiment reproducible on your machine with alpine?
```bash
$ docker run -d --rm --name test --entrypoint=/bin/sh alpine:3.15.0 '-c' 'trap "exit" TERM; while true; do sleep 1; done;'            
59a9e9f2cf193e224a6c45b189fe181eaaacce6815709531dbc878b850fb3602
$ docker ps
CONTAINER ID   IMAGE           COMMAND                   CREATED         STATUS         PORTS     NAMES
59a9e9f2cf19   alpine:3.15.0   "/bin/sh -c 'trap \"e…"   6 seconds ago   Up 5 seconds             test
$ docker stop test
test
$ docker ps
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
$
```

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,228 @@
...
+    public Guid SessionId { get; }
+
+    /// <summary>
+    /// Starts and returns the default <see cref="ResourceReaper" /> instance.
+    /// </summary>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>Task that completes when the <see cref="ResourceReaper" /> has been started.</returns>
+    public static async Task<ResourceReaper> GetAndStartDefaultAsync(CancellationToken ct = default)
+    {
+      // TODO: Use the State property?
```
PSanetra: What state property do you mean and for what purpose?

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    /// <summary>
+    /// Starts and returns a new <see cref="ResourceReaper" /> instance.
+    /// </summary>
+    /// <param name="sessionId">The session id.</param>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>Task that completes when the <see cref="ResourceReaper" /> has been started.</returns>
+    public static async Task<ResourceReaper> GetAndStartNewAsync(Guid sessionId, CancellationToken ct = default)
+    {
+      // TODO: Do we need to check if the session id is already in use?
```
PSanetra: I would keep it simple and assume that this method is called only once with the same id. Also `resourceReaper.StartAsync(ct)` would throw if there is already a `ResourceReaper` container running with the same id as the container name is already in use in that case. We could add a test for this case to define this behavior as intended.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,228 @@
...
+      await this.resourceReaperContainer.StartAsync(ct)
+        .ConfigureAwait(false);
+
+      _ = this.MaintainRyukConnection(ryukInitializedTaskSource, ct);
+
+      await ryukInitializedTaskSource.Task
+        .ConfigureAwait(false);
+    }
+
+    private async Task MaintainRyukConnection(TaskCompletionSource<bool> ryukInitializedTaskSource, CancellationToken ct = default)
```
PSanetra: We can remove the default value for the `CancellationToken` as this private method is only called with this parameter.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,228 @@
...
+                    .ConfigureAwait(false);
+
+                  await stream.FlushAsync(linkedCancellationTokenSource.Token)
+                    .ConfigureAwait(false);
+
+                  do
+                  {
+                    _ = await stream.ReadAsync(readBytes, 0, readBytes.Length, linkedCancellationTokenSource.Token)
+                      .ConfigureAwait(false);
+                    hasAcknowledge = "ack\n".Equals(Encoding.ASCII.GetString(readBytes), StringComparison.OrdinalIgnoreCase);
```
PSanetra: I think it would be better to keep reading the stream line by line as this would break if ryuk sends some other line message before the ack:
Example:
```
hello\n
ack\n
```

This code would read the stream as `hell`, `o\nac`, `k\n...`. Therefore the `ResourceReaper` would never initialize successfully although ryuk has responded with `ack\n`.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,228 @@
...
+                  }
+                  while (!hasAcknowledge);
+
+                  ryukInitializedTaskSource.SetResult(true);
+                }
+
+                if (!linkedCancellationTokenSource.IsCancellationRequested)
+                {
+                  // Keep the connection to Ryuk up.
+                  _ = await stream.ReadAsync(readBytes, 0, readBytes.Length, default)
```
PSanetra: The `cancellationToken` parameter on this `ReadAsync` should be `linkedCancellationTokenSource.Token`.

Should we also wrap this in a while loop in case future Ryuk containers are more noisy?

src/DotNet.Testcontainers/Containers/TestcontainersContainer.cs
```diff
@@ -194,6 +200,7 @@ public Task<ExecResult> ExecAsync(IList<string> command, CancellationToken ct =
     /// <returns>A task that represents the asynchronous clean up operation of a Testcontainer.</returns>
     public async Task CleanUpAsync(CancellationToken ct = default)
     {
+      // TODO: We need at least one way to clean up Windows containers. ResourceReaper won't work.
```
PSanetra: Do you know why the `ResourceReaper` won't work for Windows containers? Would it be possible to introduce a change into the Ryuk application so that Ryuk is also able to clean up Windows containers? Or would we need to provide Ryuk as a Windows container image? I think it would be nice to be able to also clean up Windows containers via a resource reaper container.

PSanetra (commented):
I have added a few comments to your changes and todos. I don't know why pushing to my PR branch does not work. I have sent you an invitation as collaborator to my fork. I think as collaborator you should be able to push to any branch.

HofmeisterAn:
> Is the following experiment reproducible on your machine with alpine?

Yes, you're right 👍 it works. I'm sorry for that. Probably I messed something else up. I'll add it again.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: `IRunningDockerContainer.State` aka `this.resourceReaperContainer.State`.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: 👍 I would prefere `ReadLineAsync` as well, but there is no overload that includes the `CancellationToken` as far as I know. I'll take a look at it again.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: > The cancellationToken parameter on this ReadAsync should be linkedCancellationTokenSource.Token.

Are you sure? This will cancel `ReadAsync` after 10 seconds (`initTimeoutCancellationTokenSource`).

> Should we also wrap this in a while loop in case future Ryuk containers are more noisy?

👍

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I removed the todo.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I removed the default value.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I changed the part that reads from the stream. It should handle different messages now too, but does that happen at all:
https://github.com/testcontainers/moby-ryuk/blob/a0508338f5a0d2f4f8de0dc5ff7355f32bf0c8ca/main.go#L125?

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I changed that

src/DotNet.Testcontainers/Containers/TestcontainersContainer.cs
HofmeisterAn: Well, if there is no WSL available and the Windows engine is enabled then you can not pull and run Unix images - as far as I know. I'll need to double check that in the next days.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: Oh, you are right. We should use the `ct` parameter here and in the while loops in this method. The `linkedCancellationTokenSource` should only be used in the initialization part of this method.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: In the current implementation of Ryuk this can not happen as far as I known, but I am not 100% sure that this can not change in the future.

You are right about that missing CancellationToken parameter in `ReadLineAsync`. I overlooked that.

src/DotNet.Testcontainers/Containers/TestcontainersContainer.cs
PSanetra: ok, in that case it would probably be a good idea to create a custom Windows container ryuk image.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I removed the todo. I think it's not necessary.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I replaced it with the `ct`.

src/DotNet.Testcontainers/Containers/TestcontainersContainer.cs
HofmeisterAn: Yep, I think that too. In the meantime we can keep the old clean up behaviour (without to purge orphan resources):

> // TODO: Either remove test or fix DisposeAsync (set ContainerListResponse; keep compatibility).

Then Windows containers get at least disposed (removed) if the test host does not crash.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,243 @@
...
+  using System.Threading.Tasks;
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Clients;
+  using DotNet.Testcontainers.Configurations;
+
+  public sealed class ResourceReaper : IAsyncDisposable
+  {
+    public const string ResourceReaperSessionLabel = TestcontainersClient.TestcontainersLabel + ".resource-reaper-session";
+
+    private const string RyukImage = "ghcr.io/psanetra/ryuk:2021.12.20";
```
HofmeisterAn, line 18: Update image to the official version.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,243 @@
...
+
+                var sendBytes = Encoding.ASCII.GetBytes(filter);
+
+                var readBytes = new byte[tcpClient.ReceiveBufferSize];
+
+                if (!ryukInitializedTaskSource.Task.IsCompleted)
+                {
+                  bool hasAcknowledge;
+
+                  var messageBuffer = new MemoryStream();
```
HofmeisterAn: Wrap it in a `using` block.

HofmeisterAn (changes requested):
A section in the README.md that explains the resource reaper / ryuk would be nice.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: Nice :+1: 
What do you think about the token in the loops? I think they should use `ct`, too.

HofmeisterAn: Yea, ofc - I missed that...

PSanetra:
@HofmeisterAn There were still some issues with the MaintainRyukConnection method and the way how the ResourceReaper was started, initialized and disposed. I have fixed those issues and added some tests for those cases.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -147,23 +178,37 @@ private async Task StartAsync(CancellationToken ct = default)
...
 
-      _ = this.MaintainRyukConnection(ryukInitializedTaskSource, ct);
+      this.connectionTask = this.MaintainRyukConnection(ryukInitializedTaskSource);
 
-      await ryukInitializedTaskSource.Task
-        .ConfigureAwait(false);
+      await Task.WhenAny(
+        Task.Delay(Timeout.InfiniteTimeSpan, ct),
+        ryukInitializedTaskSource.Task
+      ).ConfigureAwait(false);
```
PSanetra: I am not sure if there is a more elegant way to wait for the `ryukInitializedTaskSource.Task` either to complete or for `ct` to be cancelled.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -147,23 +178,37 @@ private async Task StartAsync(CancellationToken ct = default)
...
+                 this.cancellationTokenSource.Token,
+                 initTimeoutCancellationTokenSource.Token))
         {
           var endpoint = new IPEndPoint(IPAddress.Loopback, this.resourceReaperContainer.GetMappedPublicPort(RyukPort));
 
-          using (var tcpClient = new TcpClient())
+          while (!this.cancellationTokenSource.Token.IsCancellationRequested && (ryukInitializedTaskSource.Task.IsCompleted || !initTimeoutCancellationTokenSource.IsCancellationRequested))
           {
-            while (!linkedCancellationTokenSource.IsCancellationRequested)
+            using (var tcpClient = new TcpClient())
```
PSanetra: Tests were flaky when we did reuse the TcpClient in each loop pass.

tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/ResourceReaperTest.cs
```diff
@@ -119,14 +124,97 @@ await this.resourceReaper.DisposeAsync()
...
+      // Increase InitTimeout to ensure that GetAndStartNewAsync is stopped via the CancellationToken parameter
+      ResourceReaper.InitTimeout = TimeSpan.FromSeconds(10);
+
+      // Then
+      using (var cancellationTokenSource = new CancellationTokenSource())
+      {
+        var startTask = ResourceReaper.GetAndStartNewAsync(Guid.NewGuid(), cancellationTokenSource.Token);
+
+        // TODO: Is there a better way to ensure that the ryuk container is started MaintainRyukConnection is running?
+        await Task.Delay(TimeSpan.FromSeconds(2));
```
PSanetra: This is quite dirty, but I had no better idea for now.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: This shouldn't be different to:

```cs
await ryukInitializedTaskSource.Task
  .ConfigureAwait(false);
```

There must be a different error. Which test faild?

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// <summary>
+    /// Gets the <see cref="ResourceReaper" /> session id.
+    /// </summary>
+    public Guid SessionId { get; }
+
+    /// <summary>
+    /// Sets image for Ryuk container.
+    /// </summary>
+    public static string RyukImage { get; set; } = "ghcr.io/psanetra/ryuk:2021.12.20";
```
HofmeisterAn: I really don't like the static properties, are they necessary? I think the developer shouldn't take care of the `ResourceReaper` configuration. `ResourceReaper` should run in the background, ideally without any interaction. 

Maybe we can provide an overload method incl. an interface with the configuration for `GetAndStartNewAsync`.

tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/ResourceReaperTest.cs
HofmeisterAn: What about an event? `ResourceReaper` can send a notification if the connection is established.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -0,0 +1,288 @@
...
+    {
+      using (var initTimeoutCancellationTokenSource = new CancellationTokenSource(InitTimeout))
+      {
+        using (var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(
+                 this.cancellationTokenSource.Token,
+                 initTimeoutCancellationTokenSource.Token))
+        {
+          var endpoint = new IPEndPoint(IPAddress.Loopback, this.resourceReaperContainer.GetMappedPublicPort(RyukPort));
+
+          while (!this.cancellationTokenSource.Token.IsCancellationRequested && (ryukInitializedTaskSource.Task.IsCompleted || !initTimeoutCancellationTokenSource.IsCancellationRequested))
```
HofmeisterAn: I think `!linkedCancellationTokenSource.IsCancellationRequested` is enough. In which case would it be not? I even think `if (!ryukInitializedTaskSource.Task.IsCompleted)` L:225 is unnecessary. In which case the while will fail and we set the task result?

tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/ResourceReaperTest.cs
HofmeisterAn: BTW, do we need all those tests for `AutoRemove`, `CleanUp`, `CancellationToken`, etc.? I agree tests are great, but right now they are hard to understand and if the `ResourceReaper` starts properly, cleans up everything, I prefer simpler code. What do you think?

PSanetra: hmm, I would keep those tests. E.g. the `AutoRemove` test failed when status `409` Http errors were not caught. In my mind it is very important to have them otherwise it will be more or less impossible to refactor/change the code in the future. I agree that they are sometimes hard to read.

The event idea is nice 👍

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: The `RyukPort` property was nice to have to write one of the tests. Otherwise it is probably not really useful. I think especially the `RyukImage` property is needed as I expect some users need to point to some internal container registry if they are not allowed to use arbitrary registries. Alternatively we could provide a special environment variable to set a custom image name.

I could rewrite the test to use a different image instead of a port. Then we could get rid of the port property.

The `InitTimeout` is also mostly nice to have for tests. I don't know if someone maybe wants to change it in a real environment. (Maybe downloading the image might be too slow in a cicd pipeline...)

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: `!linkedCancellationTokenSource.IsCancellationRequested` would cancel the `MaintainRyukConnection` loop as soon as the `InitTimeout` has elapsed even if initialization was already successful.

`if (!ryukInitializedTaskSource.Task.IsCompleted)` is probably not necessary. It would only lead to registering the session id label again when the connection to ryuk was accidentally lost after initialization. Registering the label again is not necessary but also no problem.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: In the current code the CancellationToken `ct` would otherwise not cancel the `StartAsync` method. Instead the method would block until the initialization was successful or the `InitTimout` was elapsed.

I think the test `GetAndStartNewAsyncShouldBeACancellableDuringConnectionInitialization` would fail if only `await ryukInitializedTaskSource.Task.ConfigureAwait(false);` was used.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: Then lets overload. What about something like `GetAndStartNewAsync(Guid sessionId, string ryukImage = RyukImage, string ryukPort = RyukPort, CancellationToken ct = default)`?

> The InitTimeout is also mostly nice to have for tests. I don't know if someone maybe wants to change it in a real environment. (Maybe downloading the image might be too slow in a cicd pipeline...)

They still have `ct`.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: But then we can pass `ct` to `MaintainRyukConnection` and set either a result or an exception for `ryukInitializedTaskSource`. I think the code right now is bad, you can not cancel `MaintainRyukConnection` at all.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: ok, should we remove the `InitTimeout` completely and expect the user to provide such a timeout if they want to have one? (I think it should be the default to have one. At least for the Default-ResourceReaper.)

tests/DotNet.Testcontainers.Tests/Unit/Containers/Unix/ResourceReaperTest.cs
HofmeisterAn: Maybe we can split them later into separate classes. Then it's easier to read. But for now, let's focus on getting it stable. We can clean up the tests later.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: That is a good idea. But `ct` should cancel `MaintainRyukConnection` only during connection initialization.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: No, we can keep the timeout if you like. Just add it as an additional parameter (maybe `initTimeoutInMillisecond`), or what do you think? Any better idea? I just think the static configuration is odd.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: > `!linkedCancellationTokenSource.IsCancellationRequested` would cancel the `MaintainRyukConnection` loop as soon as the `InitTimeout` has elapsed even if initialization was already successful.

That's something I already was wondering about. In that case, we are breaking out of the second while loop, right? I checked a couple of times for an `OperationCanceledException`, but I didn't catch anything.

Doesn't the connection to Ryuk have to remain open? Do need to protect the second while loop from breaking?

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: Sorry, I don't get it. Probably I need to take a closer look at it again. As far as I understand it, `ct` can not cancel the connection initialization right now. It gets not passed to `MaintainRyukConnection`.

But it controls the termination of `StartAsync` instead. If someone cancels `ct` from outside, `Task.Delay` will finish. This can be every time. Before, while and after the connection initialization.

`MaintainRyukConnection` knows nothing about the cancellation, the task **continues to run**.

Sorry for the back and forth, but some parts are just too complex 😂.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: Yes the Ryuk connection should remain open, but after disconnect Ryuk waits 10 seconds for a reconnect. If that does not happen it triggers the clean up. I think it is only to be resilient against network issues.

HofmeisterAn: Well, why don't we remove the timeout and replace it with something like a circuit breaker? I think the timeout (the way we're using it right now) is a design flaw. It gets canceled even if the connection is established.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: Done

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: Sorry for the confusion. I have reverted this part again and pass `ct` (now called specifically `initCancellationToken`) to `MaintainRyukConnection` again. I have also moved the the timeout mechanism out of the `MaintainRyukConnection` method into `GetAndStartNewAsync`, so that `MaintainRyukConnection` has only to care about one `initCancellationToken` and one general connection maintenance token.

The `initCancellationToken` is ignored as soon as the connection initialization was successful.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
PSanetra: I am not sure how this would look like. Would you try to implement it? As mentioned in another comment I have moved the timeout logic now into the `GetAndStartNewAsync` method.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: 👍 I can take a look at it on the weekend.

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: Should I try to add the circuit breaker on the weekend too? Or should we keep the timeout?

PSanetra: I would be ok with the timeout. Maybe the circuit breaker could be introduced in a separate PR.

HofmeisterAn (changes requested):
IMO we are ready to go. What do you think? There are still a few small things to do, but nothing important. I suggest creating separate prs for them.

Squash the commits to a single one, add a proper commit message and merge (I rebased already on `develop`, thats why I force pushed)?

Thanks again for your help.

PSanetra:
@HofmeisterAn Yes, sounds good 👍 
It would be nice to wait for a new Ryuk release and replace `ghcr.io/psanetra/ryuk:2021.12.20` before a new dotnet-testcontainers release.

PSanetra:
@HofmeisterAn I have created https://github.com/testcontainers/moby-ryuk/issues/35 to see if it was possible to contribute Windows container support to the official Ryuk project or if it was needed to create a fork for that purpose.

HofmeisterAn:
> @HofmeisterAn Yes, sounds good 👍
> It would be nice to wait for a new Ryuk release and replace `ghcr.io/psanetra/ryuk:2021.12.20` before a new dotnet-testcontainers release.

We don't need to merge the changes straight to master. We can update the image later, if the official release takes some more time. In the meantime, I can take a look at the minor improvements (SonarQube, etc.).

> @HofmeisterAn I have created [testcontainers/moby-ryuk#35](https://github.com/testcontainers/moby-ryuk/issues/35) to see if it was possible to contribute Windows container support to the official Ryuk project or if it was needed to create a fork for that purpose.

👍 Thanks. As far as I know Windows does not support to mount the Docker daemon like Linux. But maybe it will work with WSL or a socket. Perhaps I can do some tests end of this or next week.

HofmeisterAn (approved):

PR #419, opened 2022-02-10
Feature/move resource reaper timeout

HofmeisterAn:


src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -135,27 +134,47 @@ public static Task<ResourceReaper> GetAndStartNewAsync(string ryukImage = RyukIm
...
+
+      initTimeout = Equals(default(TimeSpan), initTimeout) ? TimeSpan.FromSeconds(10) : initTimeout;
+
       try
       {
-        // TODO: initTimeoutCts should start after the container is up and running (this includes Docker pull).
-        using (var initTimeoutCts = new CancellationTokenSource(initTimeout.Equals(default) ? TimeSpan.FromSeconds(10) : initTimeout))
+        StateChanged?.Invoke(null, new ResourceReaperStateEventArgs(resourceReaper, ResourceReaperState.Created));
+
+        await resourceReaper.resourceReaperContainer.StartAsync(ct)
```
HofmeisterAn, line 147: @PSanetra FYI: I moved the container start out of the `initTimeout` scope. This will prevent unexpected errors e.g. if pulling the image for whatever reason takes some time.

PSanetra: Yes, good idea:+1:

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -135,27 +134,47 @@ public static Task<ResourceReaper> GetAndStartNewAsync(string ryukImage = RyukIm
     [PublicAPI]
     public static async Task<ResourceReaper> GetAndStartNewAsync(Guid sessionId, string ryukImage = RyukImage, TimeSpan initTimeout = default, CancellationToken ct = default)
     {
+      var ryukInitializedTaskSource = new TaskCompletionSource<bool>();
+
       var resourceReaper = new ResourceReaper(sessionId, ryukImage);
 
+      if (!TestcontainersSettings.ResourceReaperEnabled)
+      {
+        return resourceReaper;
```
PSanetra: Returning a `ResourceReaper` instance when `ResourceReaper` is not enabled is a little weird. This would be the only case in which `GetAndStartNewAsync` returns a `ResourceReaper` that is actually not initialized.

Maybe `TestcontainersSettings.ResourceReaperEnabled` should only disable the default `ResourceReaper` and `GetAndStartDefaultAsync` should return null if it was not enabled.

src/DotNet.Testcontainers/Configurations/TestcontainersSettings.cs
```diff
@@ -10,6 +11,12 @@
   /// </summary>
   public static class TestcontainersSettings
   {
+    /// <summary>
+    /// Gets or sets a value indicating whether the <see cref="ResourceReaper" /> is enabled or not.
+    /// </summary>
+    public static bool ResourceReaperEnabled { get; set; }
+      = true;
```
PSanetra, line 18: Should `ResourceReaperEnabled = false` also prevent adding the default `ResourceReaper` session id label to the docker resources?

src/DotNet.Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I agree. I'm not happy with it, too. The only reason I added it was: `Initialization.StopAsync`. This will trigger the default Resource Reaper clean up immediately. For Windows, it will fail, because we call `ResourceReaper.GetAndStartDefaultAsync` explicit for the first time and it can't start Ryuk.

src/DotNet.Testcontainers/Configurations/TestcontainersSettings.cs
HofmeisterAn: Oh, that's smart 🥇 I probably can combine this with `Initialization.StopAsync` and remove:

```csharp
if (!TestcontainersSettings.ResourceReaperEnabled)
{
    return resourceReaper;
}
```

PR #426, opened 2022-02-19
Fix ResourceReaper connection to non-localhost

PSanetra:
This PR fixes the ResourceReaper to be able to connect to Ryuk when it is not available on localhost.

Related discussion: https://github.com/HofmeisterAn/dotnet-testcontainers/discussions/423#discussioncomment-2179011

### DIFF
--- a/src/DotNet.Testcontainers/Containers/ResourceReaper.cs
+++ b/src/DotNet.Testcontainers/Containers/ResourceReaper.cs
@@ -223,15 +223,13 @@ namespace DotNet.Testcontainers.Containers
     /// <param name="ct">The cancellation token to cancel the <see cref="ResourceReaper" /> initialization. This will not cancel the maintained connection.</param>
     private async Task MaintainRyukConnection(TaskCompletionSource<bool> ryukInitializedTaskSource, CancellationToken ct)
     {
-      var endpoint = new IPEndPoint(IPAddress.Loopback, this.resourceReaperContainer.GetMappedPublicPort(RyukPort));
-
       while (!this.maintainConnectionCts.IsCancellationRequested && (!ct.IsCancellationRequested || ryukInitializedTaskSource.Task.IsCompleted))
       {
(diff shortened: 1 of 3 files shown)


HofmeisterAn (commented):
I tried that before replying to the discussion #423 too. But this won't work for Docker Desktop:

Hostname → `172.17.0.1` (it's `<gateway_ip>:<mapped_port>`) ❌
`host.docker.internal` → `192.168.65.2` ✔️

Maybe we can check for `host.docker.internal` in `GetContainerGateway` first.

PSanetra:
@HofmeisterAn Yes, checking for `host.docker.internal` in `GetContainerGateway` seems to be a good idea 👍

PSanetra:
@HofmeisterAn I am wondering if the Hostname property should ever return the gateway ip if I can not reach the mapped ports via that ip. I would always expect to be able to reach the mapped ports via the hostname of a container.

HofmeisterAn:
> I would always expect to be able to reach the mapped ports via the hostname of a container.

Yep, that's something I struggle too. I don't understand why, in his example, `<ryuk_container_ip>:<mapped_port>` didn't work. Maybe cause of the sibling?

`-p 51234:8080` is bound to the host (not just a single ip). I think that's better because you can access the container via host from different ips. Does that make sense? Like `localhost:51234`, `192.168.178.30:51234`, `172.30.176.1:51234`.

PSanetra:
@HofmeisterAn I think `<ryuk_container_ip>:<mapped_port>` can not work as docker will only map the port to host ips in the host network Kernel namespace. Container ips and ports are bound directly by the process in the container in a _separate_ network Kernel namespace. So `<container_ip>:<mapped_port>` is not bound even if the host is binding `0.0.0.0:<mapped_port>`.

I think mapping the container port to a host port is the way to go and should work in all Docker distributions (Linux, Windows, Mac).

PR #450, opened 2022-05-01
Add resource reaper diagnostics

PSanetra:
This PR add diagnostics information to ResourceReaperException to debug https://github.com/HofmeisterAn/dotnet-testcontainers/issues/449


HofmeisterAn:
Good idea! I changed the access modifier for `Task<ContainerInspectResponse> InspectContainer` to internal. Until now, the library don't expose anything from `Docker.DotNet.Models`. Of course, this doesn't have to be the case in the future, but I'd like to keep it consistent for now. Lets see what logs we get. Thanks.

Issue #456, opened 2022-05-14
Extend WithDockerEndpoint method with Docker.DotNet.Credentials parameter


MarcinSzyszka:
**Is your feature request related to a problem? Please describe.**
I want to connect to docker with TLS and I want to pass credentials object prepared by my own - exactly the same way as it is done in Docker.DotNet library used in dotnet-testcontainers library

**Describe the solution you'd like**
Allow to pass Docker.DotNet.Credentials object as parameter

**Describe alternatives you've considered**
There is no chance to connect to docker with TLS now

**Additional context**
It's simple change by it would allow user to connect to secured docker api (not only with TLS)
Leave credentials configuration to user exactly as it is done in Docker.DotNet library

I will create PR with changes after submiting this issue to follow contributing rules

HofmeisterAn:
I'll close this in favor of #370. There is already an ongoing discussion. Please notice, that the library does not expose models from `Docker.DotNet`. [Here](https://github.com/HofmeisterAn/dotnet-testcontainers/issues/370#issuecomment-966167220) is a short summary about the tasks.

Issue #370, opened 2021-03-05
Suport Docker TLS endpoint with client certificate authentication


prskr:
**Is your feature request related to a problem? Please describe.**

GitLab CI allows to use Docker-in-Docker (DinD) to execute various Docker related tasks. 
The DinD configuration within the public GitLab CI enforces the usage of TLS to connect to the Docker daemon which is apparently not supported by dotnet-testconainers.

**Describe the solution you'd like**

It would be nice to either have the option to further configure the [`DockerClientConfiguration`](https://github.com/dotnet/Docker.DotNet/blob/a5ee0611f37711ec81fbdb5dc48f83e642da2a2a/src/Docker.DotNet/DockerClientConfiguration.cs#L11) e.g. with a Lambda like it's quite common with ASP.NET Core to be able to set e.g. credentials or to expose a `WithDockerEnvEndpoint` method that takes the common environment variables:

* `DOCKER_HOST`
* `DOCKER_TLS_VERIFY`
* `DOCKER_TLS_CERTDIR`/`DOCKER_CERT_PATH`

into account and configures the Docker client accordingly.

**Describe alternatives you've considered**

Setting the `DockerEndpoint` to something like `https://docker:2376` allows to connect to the API via TLS already but I couldn't manage to use the client certs to authorize myself.
There are probably other options e.g. with a nginx proxy between the client and the API but I don't consider them actually very practical.
It would also be sufficient to be able to pass in a pre-configured Docker client or register a pre-configured client previously.

**Additional context**

The [GitLab CI docs](https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#tls-enabled) might also be interesting for further details, paths, etc.

Of course I'd be glad to create a PR to implement the required changes as soon as you confirmed that this request is valid and the design how to implement it is settled!

HofmeisterAn:
👍 I agree, it makes a lot of sense to support client certificate authentication. The implementation is not that difficult, I guess. I suggest the following changes:

1. Add an overloaded [WithDockerEndpoint](https://github.com/HofmeisterAn/dotnet-testcontainers/blob/1ebb2b7169ea88c2608463781179b08c3ad62df2/src/DotNet.Testcontainers/Containers/Builders/ITestcontainersBuilder.cs#L174) that gets the necessary parameters. Pass them along `TestcontainersClient` to `DockerApiClient` and create the `DockerClient` instance including the credentials.
[...]

PR #457, opened 2022-05-14
[#456] #EXTEND 'assemblyName: DotNet.Testcontainers; function: WithDo…

MarcinSzyszka:
…ckerEndpoint ' {Extended method with Docker.DotNet.Credentials parameter and made required changes to related code to pass credentials to Docker.DotNet.DockerClientConfiguration ctor }

### DIFF
--- a/src/DotNet.Testcontainers/Builders/AbstractBuilder.cs
+++ b/src/DotNet.Testcontainers/Builders/AbstractBuilder.cs
@@ -1,7 +1,8 @@
-﻿namespace DotNet.Testcontainers.Builders
+namespace DotNet.Testcontainers.Builders
 {
   using System;
   using System.Collections.Generic;
+  using Docker.DotNet;
   using DotNet.Testcontainers.Configurations;
   using DotNet.Testcontainers.Containers;
 
@@ -32,6 +33,12 @@
     public TBuilderEntity WithDockerEndpoint(string endpoint)
     {
       return this.MergeNewConfiguration(new DockerResourceConfiguration(endpoint: new Uri(endpoint)));
+    }
+
+    /// <inheritdoc cref="IAbstractBuilder{TBuilderEntity}" />
+    public TBuilderEntity WithDockerEndpoint(string endpoint, Credentials credentials)
+    {
+      return this.MergeNewConfiguration(new DockerResourceConfiguration(endpoint: new Uri(endpoint), credentials: credentials));
     }
 
     /// <inheritdoc cref="IAbstractBuilder{TBuilderEntity}" />
--- a/src/DotNet.Testcontainers/Builders/IAbstractBuilder.cs
+++ b/src/DotNet.Testcontainers/Builders/IAbstractBuilder.cs
@@ -1,6 +1,7 @@
-﻿namespace DotNet.Testcontainers.Builders
+namespace DotNet.Testcontainers.Builders
 {
   using System;
+  using Docker.DotNet;
   using DotNet.Testcontainers.Containers;
   using JetBrains.Annotations;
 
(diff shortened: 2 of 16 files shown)


HofmeisterAn (commented):
Thanks for your contribution. I can do the review at the beginning of next month. Right now, I have limit access to a proper device — sorry for the delay. I think this relates to #370 too.

src/DotNet.Testcontainers/Builders/AbstractBuilder.cs
```diff
@@ -1,7 +1,8 @@
-﻿namespace DotNet.Testcontainers.Builders
+namespace DotNet.Testcontainers.Builders
 {
   using System;
   using System.Collections.Generic;
+  using Docker.DotNet;
```
HofmeisterAn, line 5: The library (builder) does not expose models from this namespace. Please see the discussions in #370 too. I think it makes sense to combine the endpoint and credential in a single interface and offer different implementations specific for the authentication.

src/DotNet.Testcontainers/Builders/AbstractBuilder.cs
```diff
@@ -34,6 +35,12 @@ public TBuilderEntity WithDockerEndpoint(string endpoint)
       return this.MergeNewConfiguration(new DockerResourceConfiguration(endpoint: new Uri(endpoint)));
     }
 
+    /// <inheritdoc cref="IAbstractBuilder{TBuilderEntity}" />
+    public TBuilderEntity WithDockerEndpoint(string endpoint, Credentials credentials)
+    {
+      return this.MergeNewConfiguration(new DockerResourceConfiguration(endpoint: new Uri(endpoint), credentials: credentials));
```
HofmeisterAn, line 41: We should use our own interface. The implementation should consider the common [env variables](https://docs.docker.com/compose/reference/envvars/#docker_host) (`DOCKER_HOST`, `DOCKER_TLS_VERIFY`, `DOCKER_CERT_PATH`) too (at least in the future). Then we can set and apply the default credentials from the constructor based on the machine configuration.

HofmeisterAn (commented):
Please cover your changes with unit tests. I think we can exposes a secure Docker daemon on the Linux agent.

src/DotNet.Testcontainers/Builders/AbstractBuilder.cs
MarcinSzyszka: If I implement our own interface and class for certificate credentials will it be acceptable for now?
I know that perfect soultion would be connecting to dockerd secured with tls automatically as it is done in java lib but it sounds like a lot of work and as we can see in https://github.com/HofmeisterAn/dotnet-testcontainers/issues/370 last comment was in the end of 2k21 ;)
I'd like to use this lib in my projects but without credentials support is useless now.
We could treat this solution as first step to handle connection to secured dockerd - I don't exclude my further contribution on that

HofmeisterAn: > We could treat this solution as first step to handle connection to secured dockerd

I can publish your changes in its own package as soon as I’m back. For develop, I would prefer the mentioned things above.

MarcinSzyszka: Ok, if I want to try to implement solution with docker env variables, should I do that in this PR or create new one?

HofmeisterAn: Yes, a new pull request would be nice. Smaller increments would be appreciated too (first the interface). I'll publish your package later that day.

HofmeisterAn: Your changes are published in `1.6.0-beta.GitHub-457`.

HofmeisterAn:
I'll close this in favor of https://github.com/HofmeisterAn/dotnet-testcontainers/issues/370#issuecomment-1146842098. You can now use the latest version to configure / authenticate the client.

Issue #430, opened 2022-03-04
Get files inside a running container


HofmeisterAn:
**Is your feature request related to a problem? Please describe.**
Get files inside a running container without binding a file share.

**Describe the solution you'd like**
[GetArchiveFromContainerAsync](https://github.com/dotnet/Docker.DotNet/blob/f58748616cc5b679b25496926c5688294c94d850/src/Docker.DotNet/Endpoints/IContainerOperations.cs#L366)

**Describe alternatives you've considered**
\-

**Additional context**
It should be an equivalent to:
https://github.com/HofmeisterAn/dotnet-testcontainers/blob/f25ed8705fa72c9d03eb185c41bd783db0fbb908/src/DotNet.Testcontainers/Containers/IDockerContainer.cs#L141
https://github.com/HofmeisterAn/dotnet-testcontainers/blob/f25ed8705fa72c9d03eb185c41bd783db0fbb908/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs#L147

E.g. something like `ReadFileAsync`. Extract the archive from `GetArchiveFromContainerAsync` to the host.

PR #462, opened 2022-05-28
[#430] #IMPLEMENT 'assemblyName: DotNet.Testcontainers; f…

s-rech:
…unction: CopyFileFromContainerAsync'

{Implemented CopyFileFromContainerAsync to get files from inside a running container.}

### DIFF
--- a/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
+++ b/src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
@@ -193,6 +193,44 @@ namespace DotNet.Testcontainers.Clients
     }
 
     /// <inheritdoc />
+    public async Task<byte[]> CopyFileFromContainerAsync(string id, string filePath, CancellationToken ct = default)
+    {
+      Stream archive;
+      try
+      {
+        archive = await this.containers.GetArchiveFromContainerAsync(id, filePath, ct)
+          .ConfigureAwait(false);
+      }
+      catch (DockerContainerNotFoundException ex)
+      {
+        throw new InvalidOperationException(
+          $"Not possible to get archive with file {filePath} from container: {ex.ResponseBody}",
+          ex);
+      }
+
+      using (var tarInputStream = new TarInputStream(archive, Encoding.Default))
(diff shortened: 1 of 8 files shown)


HofmeisterAn (commented):
Thanks. I'll do the review today or tomorrow. I have a few things to catch up on.

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
```diff
@@ -192,6 +192,44 @@ await this.containers.ExtractArchiveToContainerAsync(id, "/", memStream, ct)
...
+        throw new InvalidOperationException(
+          $"Not possible to get archive with file {filePath} from container: {ex.ResponseBody}",
+          ex);
+      }
+
+      using (var tarInputStream = new TarInputStream(archive, Encoding.Default))
+      {
+        tarInputStream.IsStreamOwner = true;
+        var entry = tarInputStream.GetNextEntry();
+        if (entry == null || entry.IsDirectory || entry.Name != Path.GetFileName(filePath))
```
HofmeisterAn, line 215: I've added some minor improvements. Before I'll merge the pull request, I'd like to ask if this couldn't be simplified to:

```csharp
if (entry.IsDirectory)
{
  throw new InvalidOperationException("Can not read from a directory. Use a file instead.");
}
```

I think we don't need the other conditions. If the file (path) does not exist, `GetArchiveFromContainerAsync` will already fail.

HofmeisterAn (approved):
I've renamed `CopyFileFromContainerAsync` to `ReadFileAsync` (maybe down- and upload is better in the future) and moved the tests into their own class. Other than that, your pull request looks good! Thanks again.

src/DotNet.Testcontainers/Clients/TestcontainersClient.cs
s-rech: Yes, that is correct.

HofmeisterAn:
Published in `1.6.0-beta.2232`, closes #430.

Issue #466, opened 2022-06-01
Add support for image name substitution (prefixing of Docker Hub images)


bohlenc:
**Is your feature request related to a problem? Please describe.**
To handle Docker Hub rate limiting, I want to use a pull through cache to proxy all images pulled from Docker Hub. 
Currently, to make Testcontainers use this pull through cache, developers would have to modify the container setup for all images hosted on Docker Hub and prepend the hostname of the pull through cache to the image name. This is tiresome and error prone.

**Describe the solution you'd like**
I would like to centrally configure a prefix that is prepended to all image names hosted on Docker Hub.

Example: with a configured prefix "my.proxy.com" the image "docker:latest" becomes "my.proxy.com/docker:latest" and the image "my.azurecr.io/docker:latest" would be left as is.

**Describe alternatives you've considered**
- Manually changing all usages of Docker Hub images
- Implementing custom code on top of Testcontainers to centrally configure a prefix

**Additional context**
Testcontainers for Java already has a feature called Image Name Substitution (https://www.testcontainers.org/features/image_name_substitution/).

I already implemented support for a simple prefixing substitution strategy. If you agree with the premise of this issue, I can open a merge request.

HofmeisterAn:
Yes — of course, we can add this functionality. But why do we need so many changes? Couldn't we add this functionality with just a few lines in `DockerImage`?

bohlenc:
I think we could manage without the changes to `TestcontainersBuilder` and move the application of the substitution into `DockerImage`. I initially chose not to, because I see `DockerImage` as a simple primitive which I rather keep clean of any kind of logic or policy. 
I introduced the abstraction of `IImageNameSubstitution` because I see the case for additional implementations, like `testcontainers-java` provides.

If you disagree here, I am certainly willing to simplify and move the functionality into `DockerImage` so we can get this merged.

HofmeisterAn:
I'm just thinking about a simpler implementation. It looks so much code for such a simple task. Maybe it's better to extend the builder with a private method that append the prefix.

bohlenc:
I just think the code is better structured and extensible the way it currently is. Besides, a significant portion of the added LoC is test data and documentation.
But I definitely understand your desire to keep the code base simple. And I agree with you - there **is** a simpler solution to precisely solve the task at hand, and not more.

To bring this to a fruitful end: I will get rid of the abstraction and add the core logic directly in `TestcontainersBuilder` next thing in the morning.

[...]

PR #467, opened 2022-06-01
[#466] #IMPLEMENT 'assemblyName: DotNet.Testcontainers; function: Ima…

bohlenc:
…ge Name Substitution'

### DIFF
--- a/src/DotNet.Testcontainers/Builders/TestcontainersBuilder.cs
+++ b/src/DotNet.Testcontainers/Builders/TestcontainersBuilder.cs
@@ -86,7 +86,8 @@ namespace DotNet.Testcontainers.Builders
     /// <inheritdoc cref="ITestcontainersBuilder{TDockerContainer}" />
     public ITestcontainersBuilder<TDockerContainer> WithImage(IDockerImage image)
     {
-      return this.MergeNewConfiguration(new TestcontainersConfiguration(image: image));
+      var modifiedImage = MaybePrependPrefixToImageName(image);
+      return this.MergeNewConfiguration(new TestcontainersConfiguration(image: modifiedImage));
     }
 
     /// <inheritdoc cref="ITestcontainersBuilder{TDockerContainer}" />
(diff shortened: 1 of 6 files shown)


tests/DotNet.Testcontainers.Tests/Unit/Builders/TestcontainerBuilderTest.cs
```diff
@@ -0,0 +1,68 @@
...
+  [InlineData("my.proxy.com:443", "bar:1.0.0", "my.proxy.com:443/bar:1.0.0")]
+  [InlineData("my.proxy.com", "foo/bar:1.0.0", "my.proxy.com/foo/bar:1.0.0")]
+  [InlineData("my.proxy.com/my-path", "foo/bar:1.0.0", "my.proxy.com/my-path/foo/bar:1.0.0")]
+  [InlineData("my.proxy.com:443", "foo/bar:1.0.0", "my.proxy.com:443/foo/bar:1.0.0")]
+  [InlineData("my.proxy.com:443/my-path", "foo/bar:1.0.0", "my.proxy.com:443/my-path/foo/bar:1.0.0")]
+  [InlineData("my.proxy.com", "myregistry.azurecr.io/foo/bar:1.0.0", "myregistry.azurecr.io/foo/bar:1.0.0")]
+  [InlineData("my.proxy.com", "myregistry.azurecr.io:443/foo/bar:1.0.0", "myregistry.azurecr.io:443/foo/bar:1.0.0")]
+  public void ImageNameFromStringIsPrefixedWhenPrefixIsSet(string prefix, string originalImageName, string expectedName)
+  {
+    TestcontainersSettings.DockerHubImagePrefix = prefix;
```
HofmeisterAn: We just need to make sure, that the tests won't run with other tests. Changing the static property might not be a good idea with parallel tests.

HofmeisterAn (commented):
Thanks. I think thats much cleaner.

tests/DotNet.Testcontainers.Tests/Unit/Builders/TestcontainerBuilderTest.cs
bohlenc: I wasn't aware tests are executed in parallel.

Fixed.

HofmeisterAn: That's not how it works :). You'll need a `CollectionDefinition` first and then set `DisableParallelization`. I can add it. See: https://github.com/xunit/xunit/issues/1999#issuecomment-522635397.

bohlenc: Then maybe the xunit documentation needs fixing: https://xunit.net/docs/running-tests-in-parallel > "Custom Test Collections". Other tests relying on `TestcontainersSettings.DockerHubImagePrefix` then just would have to be in the same collection, no?

Or am I missing something there?

bohlenc: > That's not how it works :). You'll need a `CollectionDefinition` first and then set `DisableParallelization`. I can add it. See: [xunit/xunit#1999 (comment)](https://github.com/xunit/xunit/issues/1999#issuecomment-522635397).

Yep, saw that issue. From what I gather from the docs though that is not required. Maybe I'm wrong.

HofmeisterAn: Each test class is a unique test collection. By default, test collections run parallel. The unit tests in this library are in different test collections. Tests that access the static resource (almost all of them) can't run parallel.

> Other tests relying on `TestcontainersSettings.DockerHubImagePrefix` then just would have to be in the same collection, no?
> 
> Or am I missing something there?

Yes, exactly. Almost all tests rely on that static property. But putting all of them in the same collection is way too complicated. We can just take advantage of the `[CollectionDefinition(DisableParallelization = true)]` behavior:

> Parallel-capable test collections will be run first (in parallel), followed by parallel-disabled test collections (run sequentially).

bohlenc: Ah, got it. Thanks!

bohlenc: I added a simple `ParallelizationDisabled` collection.

HofmeisterAn (approved):
I'll merge it later that day. I'll add a minor change that changes the `IDockerContainer.ImageName` return type to `IDockerImage` and reduces the amount of `InlineData`, we can replace it with `MemberData` (incl. the code smell).

HofmeisterAn:
Published in `1.6.0-beta.2242`, closes #466.

Issue #483, opened 2022-06-17
Customize docker registry for ryuk image when using default resource reaper


harrhp:
**Is your feature request related to a problem? Please describe.**
When default session of resource reaper is [started](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers/Containers/ResourceReaper.cs#L108) it uses image that points to public registry, it is a problem when running tests in CI without internet access. 

**Describe the solution you'd like**
Expose property in `TestcontainersSettings` that allows to specify registry (and tag ?) for ryuk image

**Describe alternatives you've considered**
It's probably possible to disable default resource reaper and manage session manually but it's more code

**Additional context**

HofmeisterAn:
OC, let's add this. What about `IDockerImage TestcontainersSettings.ResourceReaperImage`? We could achieve the same thing with `HubImageNamePrefix`, but the Ryuk image we're using right now is not on Docker Hub.

harrhp:
`IDockerImage TestcontainersSettings.ResourceReaperImage` is good, it would allow to change image name also (potentialy for another existing image or custom built image)

HofmeisterAn:
Published in `2.1.0-beta.2561986470`.

PR #489, opened 2022-06-24
feat(#483): Added TestcontainersSettings.ResourceReaperImage property

vlaskal:
Added TestcontainersSettings.ResourceReaperImage property to customize Resource Reaper image base on #483.
Same implementation as is `DockerEndpointAuthConfig`.

I have just questions if the condition in constructior `.WithImage(ryukImage ?? TestcontainersSettings.ResourceReaperImage.FullName)` should not be covered by any test?

### DIFF
--- a/src/Testcontainers/Configurations/TestcontainersSettings.cs
+++ b/src/Testcontainers/Configurations/TestcontainersSettings.cs
@@ -1,7 +1,8 @@
-﻿namespace DotNet.Testcontainers.Configurations
+namespace DotNet.Testcontainers.Configurations
 {
   using System.Runtime.InteropServices;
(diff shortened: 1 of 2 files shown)


HofmeisterAn (approved):

HofmeisterAn:
OC, if you like, you can add a unit test, but I don't think it's mandatory in this case.

HofmeisterAn:
https://github.com/testcontainers/testcontainers-dotnet/commit/3c129edd8c15fd6f43c8240971390e4971e698dd

vlaskal:
> OC, if you like, you can add a unit test, but I don't think it's mandatory in this case.

Ok let's leave it as is.

PR #490, opened 2022-06-25
fix(#449): Don't let Docker Desktop for Windows choose the random public host port

HofmeisterAn:


HofmeisterAn:
@srollinet We can do something like this I think. What do you think?

src/Testcontainers/Configurations/TestcontainersSettings.cs
```diff
@@ -27,6 +31,19 @@ public static class TestcontainersSettings
...
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="ResourceReaper" /> might not be able to connect to Ryuk on Docker Desktop for Windows.
+    /// Assigning a random port might run into the excluded port range. The container starts, but we can not establish a TCP connection.
+    /// https://github.com/docker/for-win/issues/3171
+    /// https://github.com/docker/for-win/issues/11584
+    /// </remarks>
+    [PublicAPI]
+    public static Func<IDockerEndpointAuthenticationConfiguration, ushort> ResourceReaperPublicHostPort { get; set; }
+      = GetResourceReaperPublicHostPort;
```
srollinet, line 45: I find the name a bit misleading, it doesn't sounds like a function

srollinet (commented):
It looks good and it allows the user to have full control over the public host port

All tests run fine on my computer.

I am also running the tests from `ResourceReaper.Tests` with the function "Repeat until failure" and for now, everything is fine!

srollinet (approved):

src/Testcontainers/Configurations/TestcontainersSettings.cs
HofmeisterAn: I'm not happy with it either. Any other idea?

PR #500, opened 2022-07-01
Better logs for image operations

michal-korniak:
Hi
I noticed that the build process doesn't produce any logs. This code will change it.

### DIFF
--- a/src/Testcontainers/Clients/TraceProgress.cs
+++ b/src/Testcontainers/Clients/TraceProgress.cs
@@ -18,11 +18,7 @@ namespace DotNet.Testcontainers.Clients
       this.LogProgressIfNotNullOrWhiteSpace(value.Stream);
       this.LogProgressIfNotNullOrWhiteSpace(value.ProgressMessage);
       this.LogProgressIfNotNullOrWhiteSpace(value.Status);
-
-      if (value.Error != null)
-      {
-        this.logger.LogError(value.Error.Message.Trim());
-      }
+      this.LogErrorIfNotNullOrWhiteSpace(value.Error.Message);
     }
 
     private void LogProgressIfNotNullOrWhiteSpace(string message)
@@ -30,6 +26,14 @@ namespace DotNet.Testcontainers.Clients
       if (!string.IsNullOrWhiteSpace(message))
       {
         this.logger.Progress(message.Trim());
+      }
+    }
+
+    private void LogErrorIfNotNullOrWhiteSpace(string message)
+    {
+      if (!string.IsNullOrWhiteSpace(message))
+      {
+        this.logger.Error(message.Trim());
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/Testcontainers/Clients/TraceProgress.cs
```diff
@@ -15,7 +15,23 @@ public TraceProgress(ILogger logger)
 
     public void Report(JSONMessage value)
     {
-      this.logger.Progress(value.ProgressMessage);
+      this.LogProgressIfNotNullOrWhiteSpace(value.Stream);
```
HofmeisterAn, line 18: Can't we use a single logging command (pattern) for those three properties?

src/Testcontainers/Clients/TraceProgress.cs
```diff
@@ -15,7 +15,23 @@ public TraceProgress(ILogger logger)
...
+      this.LogProgressIfNotNullOrWhiteSpace(value.ProgressMessage);
+      this.LogProgressIfNotNullOrWhiteSpace(value.Status);
+
+      if (value.Error != null)
+      {
+        this.logger.LogError(value.Error.Message.Trim());
```
HofmeisterAn: Can't you use `value.ErrorMessage`? Please consider the logging concept (https://docs.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging).

HofmeisterAn (changes requested):
> the build process doesn't produce any logs

Did you set the log level to trace?

michal-korniak:
>

michal-korniak:
> > the build process doesn't produce any logs
> 
> Did you set the log level to trace?

No. It's not about log level, but rather on that, that depending on situation we get logs on different properties.

image
image
image

src/Testcontainers/Clients/TraceProgress.cs
michal-korniak: We can, but I don't know if it makes sense. I added few screenshots from the state JSONMessage object bellow. Notice that every event generates exactly one message (only one property is filled). Because of that I'm not sure if it will be good to create such logs: "Stream: [null], ProgressMessage: [null], Status: Lorem Ipsum dadadada"

src/Testcontainers/Clients/TraceProgress.cs
michal-korniak: We can and I use it in few update. These properties are always equal. I also changed LogError to new extension method that I created similar way as Progress().

HofmeisterAn:
> No. It's not about log level

What I meant, is that you get some logs if you set the level to trace.

src/Testcontainers/Clients/TraceProgress.cs
HofmeisterAn: 👍

HofmeisterAn:
Is it always just one property that contains data?

michal-korniak:
> Is it always just one property that contains data?

Yes. I tested it for a while and never noticed that more than one property contains data.

PR #501, opened 2022-07-01
Throwing exception if image doesn't exist after build

michal-korniak:
Hi
I noticed that the framework doesn't give any feedback's if image is not created. This leads to errors in the other lines of code:

"Docker.DotNet.DockerImageNotFoundException : Docker API responded with status code=NotFound, response={"message":"No such image: testcontainers:1656427318745"}"

### DIFF
--- a/src/Testcontainers/Clients/DockerImageOperations.cs
+++ b/src/Testcontainers/Clients/DockerImageOperations.cs
@@ -114,10 +114,10 @@ namespace DotNet.Testcontainers.Clients
         await this.Docker.Images.BuildImageFromDockerfileAsync(buildParameters, dockerFileStream, Array.Empty<AuthConfig>(), new Dictionary<string, string>(), this.traceProgress, ct)
           .ConfigureAwait(false);
 
-        bool isImageCreated = await this.ExistsWithNameAsync(image.FullName, ct).ConfigureAwait(false);
-        if (!isImageCreated)
+        var imageHasBeenCreated = await this.ExistsWithNameAsync(image.FullName, ct).ConfigureAwait(false);
+        if (!imageHasBeenCreated)
         {
-          throw new Exception($"Failed to create image {image.FullName}");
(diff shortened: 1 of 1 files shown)


src/Testcontainers/Clients/DockerImageOperations.cs
```diff
@@ -113,6 +113,12 @@ await this.DeleteAsync(image, ct)
       {
         await this.Docker.Images.BuildImageFromDockerfileAsync(buildParameters, dockerFileStream, Array.Empty<AuthConfig>(), new Dictionary<string, string>(), this.traceProgress, ct)
           .ConfigureAwait(false);
+
+        bool isImageCreated = await this.ExistsWithNameAsync(image.FullName, ct).ConfigureAwait(false);
```
HofmeisterAn: `var`

src/Testcontainers/Clients/DockerImageOperations.cs
```diff
@@ -113,6 +113,12 @@ await this.DeleteAsync(image, ct)
...
+        if (!isImageCreated)
+        {
+          throw new Exception($"Failed to create image {image.FullName}");
```
HofmeisterAn: Something like

```csharp
if (!imageHasBeenCreated)
{
  throw new InvalidOperationException($"Docker image {image.FullName} has not been created.");
}
```

HofmeisterAn (changes requested):

src/Testcontainers/Clients/DockerImageOperations.cs
michal-korniak: changed

src/Testcontainers/Clients/DockerImageOperations.cs
michal-korniak: changed

HofmeisterAn (approved):
I'll merge it together with the other pull requests. Thanks.

PR #503, opened 2022-07-02
Add support for WithCreateContainerParametersModifier

### LINKED ISSUE #481: Support setting ulimits on Docker containers

We are currently trying to spin up an Elastic container as part of our testing infrastructure, to automate integration tests against multiple versions of Elastic. Unfortunately, this requires use of the Docker `ulimits` argument, as exemplified in [their documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#docker-file):

```yaml
es01:
    image: docker.elastic.co/elasticsearch/elasticsearch:${STACK_VERSION}
   ...
    ulimits:
      memlock:
        soft: -1
        hard: -1
```

Setting the ulimits is [supported by Docker run](https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit), for instance:

```bash
docker run --ulimit nofile=1024:1024 debian
```

We would like for this to become available via testcontainers-dotnet. Something like the following would be great:

```c#
new TestcontainersBuilder<MyContainerType>()
    .WithImage(configuration.Image)
    .WithPortBinding(configuration.Port, configuration.DefaultPort)
    .WithUlimit(Ulimit.Memlock, -1, -1)
```

Docker supports the syntax: `<type>=<soft limit>[:<hard limit>]`, so the `.WithUlimit()` command would need to support both the type and a soft limit, and optionally a hard limit as well. If the hard limit is not provided, the soft limit is used for both values.

As of writing this proposal, Docker supports [the following ulimit types](https://github.com/docker/go-units/blob/e682442797b36348f8e1f98defdbf32bac0b6c6f/ulimit.go#L46):

```go
"core"
"cpu"
"data"
"fsize"
"locks"
"memlock"
"msgqueue"
"nice"
"nofile"
"nproc"
"rss"
"rtprio"
"rttime"
"sigpending"
"stack"
```

**Describe alternatives you've considered**

We have tried to set the ulimits globally in our Docker config `/etc/docker/daemon.json`. This is only a possibility in local development environments, and as such does not work in our CI/CD pipelines. It also seems to make our workstations a little unstable when we set all these limits globally. We would like to be able to set it on a container-by-container basis.

**Additional context**

None.

Xitric:
This PR adds a new builder method `WithCreateContainerParametersModifier()` which corresponds to [`withCreateContainerCmdModifier()` from the Java implementation](https://www.testcontainers.org/features/advanced_options/#customizing-the-container).

The purpose of this method is to gain access to low-level configuration on the Docker.DotNet client, for instance to configure Ulimits, without having to expose all of these low-level configurations through the testcontainer builder.

Merging this PR closes testcontainers#481.

### DIFF
--- a/src/Testcontainers/Builders/BuildConfiguration.cs
+++ b/src/Testcontainers/Builders/BuildConfiguration.cs
@@ -22,7 +22,7 @@ namespace DotNet.Testcontainers.Builders
     /// </summary>
     /// <param name="next">Changed configuration.</param>
     /// <param name="previous">Previous configuration.</param>
-    /// <typeparam name="T">Type of <see cref="IReadOnlyDictionary{TKey,TValue}" />.</typeparam>
+    /// <typeparam name="T">Type of <see cref="IEnumerable{T}" />.</typeparam>
     /// <returns>An updated configuration.</returns>
     public static IEnumerable<T> Combine<T>(IEnumerable<T> next, IEnumerable<T> previous)
       where T : class
@@ -35,6 +35,25 @@ namespace DotNet.Testcontainers.Builders
       {
         return next.Concat(previous).ToArray();
       }
+    }
+
+    /// <summary>
+    /// Combines all existing and new configuration changes while preserving the original order of insertion.
+    /// If there are no changes, the previous configurations are returned.
+    /// </summary>
+    /// <param name="next">Changed configuration.</param>
+    /// <param name="previous">Previous configuration.</param>
+    /// <typeparam name="T">Type of <see cref="IReadOnlyList{T}" />.</typeparam>
+    /// <returns>An updated configuration.</returns>
+    public static IReadOnlyList<T> Combine<T>(IReadOnlyList<T> next, IReadOnlyList<T> previous)
+      where T : class
+    {
+      if (next == null || previous == null)
+      {
+        return next ?? previous;
+      }
+
+      return previous.Concat(next).ToArray();
     }
 
     /// <summary>
--- a/src/Testcontainers/Builders/ITestcontainersBuilder.cs
+++ b/src/Testcontainers/Builders/ITestcontainersBuilder.cs
@@ -4,6 +4,7 @@ namespace DotNet.Testcontainers.Builders
   using System.Collections.Generic;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet.Models;
   using DotNet.Testcontainers.Configurations;
   using DotNet.Testcontainers.Containers;
   using DotNet.Testcontainers.Images;
@@ -309,6 +310,16 @@ namespace DotNet.Testcontainers.Builders
     /// <remarks>Multiple wait strategies are executed one after the other.</remarks>
     [PublicAPI]
     ITestcontainersBuilder<TDockerContainer> WithWaitStrategy(IWaitForContainerOS waitStrategy);
+
+    /// <summary>
+    /// Allow low level modifications of <see cref="CreateContainerParameters"/> after the Testcontainer configuration has been applied.
+    /// When adding multiple modifiers, they will be executed in order of insertion.
+    /// </summary>
+    /// <param name="parameterModifier">The action that is invoked for modifying the <see cref="CreateContainerParameters"/> instance.</param>
+    /// <returns>A configured instance of <see cref="ITestcontainersBuilder{TDockerContainer}" />.</returns>
+    /// <remarks>Warning: This exposes the underlying Docker.DotNet API so it might change outside of our control.</remarks>
+    [PublicAPI]
+    ITestcontainersBuilder<TDockerContainer> WithCreateContainerParametersModifier(Action<CreateContainerParameters> parameterModifier);
 
     /// <summary>
     /// Sets the startup callback to invoke after the Testcontainer start.
--- a/src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
+++ b/src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
@@ -4,6 +4,7 @@ namespace DotNet.Testcontainers.Configurations
   using System.Collections.Generic;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet.Models;
   using DotNet.Testcontainers.Containers;
   using DotNet.Testcontainers.Images;
   using DotNet.Testcontainers.Networks;
@@ -97,6 +98,15 @@ namespace DotNet.Testcontainers.Configurations
     /// Gets the wait strategies.
     /// </summary>
     IEnumerable<IWaitUntil> WaitStrategies { get; }
+
+    /// <summary>
+    /// Gets the modifier callbacks for <see cref="CreateContainerParameters"/>.
+    /// </summary>
+    /// <remarks>
+    /// These actions will be executed after the remaining Testcontainer configuration has been applied on the Docker
+    /// client, but just before creating the container.
+    /// </remarks>
+    IReadOnlyList<Action<CreateContainerParameters>> ParameterModifiers { get; }
 
     /// <summary>
     /// Gets the startup callback.
--- a/src/Testcontainers/Configurations/Containers/TestcontainersConfiguration.cs
+++ b/src/Testcontainers/Configurations/Containers/TestcontainersConfiguration.cs
@@ -4,6 +4,7 @@ namespace DotNet.Testcontainers.Configurations
   using System.Collections.Generic;
   using System.Threading;
   using System.Threading.Tasks;
+  using Docker.DotNet.Models;
   using DotNet.Testcontainers.Containers;
   using DotNet.Testcontainers.Images;
   using DotNet.Testcontainers.Networks;
@@ -42,6 +43,7 @@ namespace DotNet.Testcontainers.Configurations
     /// <param name="networkAliases">The container network aliases.</param>
     /// <param name="outputConsumer">The output consumer.</param>
     /// <param name="waitStrategies">The wait strategies.</param>
+    /// <param name="parameterModifiers">The modifier callbacks for <see cref="CreateContainerParameters"/>.</param>
     /// <param name="startupCallback">The startup callback.</param>
     /// <param name="autoRemove">A value indicating whether the Testcontainer is removed by the Docker daemon or not.</param>
     /// <param name="privileged">A value indicating whether the Testcontainer has extended privileges or not.</param>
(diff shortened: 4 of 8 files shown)


HofmeisterAn:
I won't be able to do the review until the beginning of next week - sorry, but thanks already for your pull request and contribution.

src/Testcontainers/Builders/BuildConfiguration.cs
```diff
@@ -37,6 +37,25 @@ public static IEnumerable<T> Combine<T>(IEnumerable<T> next, IEnumerable<T> prev
...
 
+    /// <summary>
+    /// Combines all existing and new configuration changes while preserving the original order of insertion.
+    /// If there are no changes, the previous configurations are returned.
+    /// </summary>
+    /// <param name="next">Changed configuration.</param>
+    /// <param name="previous">Previous configuration.</param>
+    /// <typeparam name="T">Type of <see cref="IList{T}" />.</typeparam>
+    /// <returns>An updated configuration.</returns>
+    public static IList<T> Combine<T>(IList<T> next, IList<T> previous)
```
HofmeisterAn: I think we don't need that. We can change `Combine<T>(IEnumerable<T> next, IEnumerable<T> previous)` to `next.Concat(previous).Reverse().ToArray();`. I think that's what most expect anyway. The change of the order shouldn't be an issue for existing configurations.

src/Testcontainers/Builders/ITestcontainersBuilder.cs
```diff
@@ -319,6 +320,15 @@ public interface ITestcontainersBuilder<out TDockerContainer> : IAbstractBuilder
...
     ITestcontainersBuilder<TDockerContainer> WithStartupCallback(Func<IRunningDockerContainer, CancellationToken, Task> startupCallback);
 
+    /// <summary>
+    /// Allow low level modifications of <see cref="CreateContainerParameters"/> after the Testcontainer configuration has been applied.
+    /// When adding multiple modifiers, they will be executed in order of insertion.
+    /// </summary>
+    /// <param name="parameterModifier">The action that is invoked for modifying the <see cref="CreateContainerParameters"/> instance.</param>
+    /// <returns>A configured instance of <see cref="ITestcontainersBuilder{TDockerContainer}" />.</returns>
+    /// <remarks>Warning: This exposes the underlying Docker.DotNet API so it might change outside of our control.</remarks>
+    ITestcontainersBuilder<TDockerContainer> WithCreateContainerParametersModifier(Action<CreateContainerParameters> parameterModifier);
```
HofmeisterAn, line 322: Add `[PublicAPI]` attribute.

src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
```diff
@@ -105,5 +106,14 @@ public interface ITestcontainersConfiguration : IDockerResourceConfiguration
...
     Func<ITestcontainersContainer, CancellationToken, Task> StartupCallback { get; }
+
+    /// <summary>
+    /// Get the modifier callbacks for <see cref="CreateContainerParameters"/>.
+    /// </summary>
+    /// <remarks>
+    /// These actions will be executed after the remaining Testcontainer configuration has been applied on the Docker
+    /// client, but just before creating the container.
+    /// </remarks>
+    IList<Action<CreateContainerParameters>> ParameterModifiers { get; }
```
HofmeisterAn: `IEnumerable<string>`

src/Testcontainers/Configurations/Containers/TestcontainersConfiguration.cs
```diff
@@ -64,6 +66,7 @@ public TestcontainersConfiguration(IDockerResourceConfiguration dockerResourceCo
       IOutputConsumer outputConsumer = null,
       IEnumerable<IWaitUntil> waitStrategies = null,
       Func<ITestcontainersContainer, CancellationToken, Task> startupCallback = null,
+      IList<Action<CreateContainerParameters>> parameterModifiers = null,
```
HofmeisterAn: If you don't mind, changing the order with `startupCallback` would be nice. This will follow the flow of creating and starting a container.

HofmeisterAn (changes requested):
Thanks for your contribution. I added some minor suggestions and improvements. Could you please cover your changes with a unit test in: `DotNet.Testcontainers.Tests.Unit.Containers.Unix.TestcontainersContainerTest.WithConfiguration`. Something like:

```csharp
...
.WithCreateContainerParametersModifier(parameters = parameters.Name = "Foo")
.WithCreateContainerParametersModifier(parameters = parameters.Name = "Bar")
...

Assert("Bar", ...)
```

src/Testcontainers/Builders/BuildConfiguration.cs
Xitric: True, but the reason I added this version of `Combine` is because there is technically nothing in the contract of the existing method that guarantees ordering of the combined enumerables, because `IEnumerable` does not guarantee preservation of order - `IList` does. So the fact that the existing `Combine` method preserves ordering is just a lucky implementation detail. At any time, this assumed contract could be broken either in the `Concat` method or in the implementation of `Combine`.

So I do not like the idea of treating the sequence of parameter modifiers as an `IEnumerable`, even if I could just convert the return value of `Combine` into a `List` - there is nothing in the contract that guarantees that my elements have not been shuffled around in the meantime.

src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
Xitric: I don't quite understand this comment?

Xitric:
@HofmeisterAn Yes of course I should add a test. I looked around for a short while but I could not determine the right location for it. It has been added now. I just need clarification on a couple of your comments now

src/Testcontainers/Builders/BuildConfiguration.cs
HofmeisterAn: > because `IEnumerable` does not guarantee preservation of order

Yes - that's true, but the implementation does, and we choose the implementation. We can keep `IList` too, I just think it makes sense to stick to one implementation (and I prefer immutable).

src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
HofmeisterAn: That's nonsense, I meant `IEnumerable<Action<CreateContainerParameters>>`

Xitric: Once again, I really don't like the idea of documenting `ParameterModifiers` as an `IEnumerable` when we can (and IMO should) make a stronger guarantee in the contract. We need to promise that the order of `ParameterModifiers` is retained, since it could have really confusing consequences if the order changes over time - and `IEnumerable` is too generic to express such a guarantee. The correct collection type here is `IList` since it expresses exactly the contract intended with `ParameterModifiers`.

But this is your project after all. If you want me to change it to `IEnumerable`, I will, but I also think it is a mistake ;)

src/Testcontainers/Builders/BuildConfiguration.cs
Xitric: I agree with the immutable implementation. The `Combine` method should certainly not mutate the input enumerables. Depending on the other conversation, if we choose to remove `Combine<T>(IList<T> next, IList<T> previous)` and modify `Combine<T>(IEnumerable<T> next, IEnumerable<T> previous)` instead, I will add some unit tests to ensure that the implementation preserves ordering, despite the weaker contract.

src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
HofmeisterAn: > The correct collection type here is `IList` since it expresses exactly the contract intended with `ParameterModifiers`

Immutable is a contract too, and this won't work well with `IList`.

> But this is your project after all.

It's a community project. My personal preferences don't matter 😃. I think both points are valid, there is just no type that covers both in .NET. Usually, I try to reduce changes to a minimal, but let's stick to `IList<T>` and add a comment?

src/Testcontainers/Builders/BuildConfiguration.cs
HofmeisterAn: Let's keep `IList<T>`, see my comment below.

Xitric:
I see your point with immutability. I am not that well versed in the dotnet
collection types, but how about an `IReadOnlyList<T>` or
`IImmutableList<T>`? Or do you think that gets too specific?

man. 4. jul. 2022 18.28 skrev Andre Hofmeister ***@***.***>:

src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
Xitric: Ah, now I see what you were getting at with immutability. This makes perfect sence.

I am not that well versed in the dotnet collection types, but how about an `IReadOnlyList<T>` or `IImmutableList<T>`? Or do you think that gets too specific?

Edit: While `IReadOnlyList<T>` technically does not guarantee immutability, neither does IEnumerable. But they both express read-only views of the underlying collection, so `IReadOnlyList<T>` should be sufficient here.

HofmeisterAn:
> I see your point with immutability. I am not that well versed in the dotnet collection types, but how about an `IReadOnlyList<T>` or `IImmutableList<T>`?

Me neither. I checked `IImmutableList<T>` too, but that's not part of `netstandard` and `IReadOnlyCollection<T>` is just an `IEnumerable<T>`. But I missed `IReadOnlyList<T>`. That sounds right 👏.

src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
HofmeisterAn: > I see your point with immutability. I am not that well versed in the dotnet collection types, but how about an `IReadOnlyList<T>` or `IImmutableList<T>`?

Me neither. I checked `IImmutableList<T>` too, but that's not part of `netstandard` and `IReadOnlyCollection<T>` is just an `IEnumerable<T>`. But I missed `IReadOnlyList<T>`. That sounds right 👏.

Xitric: Perfect, I will update the PR later tonight to use `IReadOnlyList<T>` instead.

Xitric:
@HofmeisterAn There you go. Also added some unit tests for `BuildConfiguration` to clarify expectations of the different method implementations. That way we can catch it if someone breaks contract.

I have never used XUnit before, so pardon me if the class structure is not in line with best practices. Let me know if I should use a combination of the static outer class and sealed inner classes (I have seen that in other tests), and I can refactor it.

HofmeisterAn (approved):
If you like, you can squash the commits. I'll merge the pull request later that day, tomorrow at the latest.

Xitric:
@HofmeisterAn There we go - squashed into a single commit. Thank you for the collaboration so far 🚀

HofmeisterAn:
Closes #481.

PR #511, opened 2022-07-15
Removing temp files after build process

michal-korniak:
Current implementation doesn't clean temp resources after image build process. It leads to the situation when temp directory becomes very large (especially if you use windows containers).

Zrzut ekranu 2022-07-15 134042


HofmeisterAn (approved):

HofmeisterAn:
https://github.com/testcontainers/testcontainers-dotnet/commit/b9d28e465559e036623ca7d5138fd809e2df1d14

PR #512, opened 2022-07-15
DockerImageOperations.BuildAsync - Throwing meaningful exception if docker image name is not lower case

michal-korniak:
Currently this scenario isn't properly handled which leads to the exception with low level message that doesn't give any clues what went wrong.

image

I wanted to add unit tests to this, but it will require more time than I can afford, The BuildAsync method invokes methods from static classes, so it will be necessary to create additional abstractions to make it mockable.

### DIFF
--- a/src/Testcontainers/Clients/DockerImageOperations.cs
+++ b/src/Testcontainers/Clients/DockerImageOperations.cs
@@ -89,6 +89,11 @@ namespace DotNet.Testcontainers.Clients
 
     public async Task<string> BuildAsync(IImageFromDockerfileConfiguration configuration, CancellationToken ct = default)
     {
+      if (configuration.Image.Name.Any(x => char.IsUpper(x)))
+      {
+        throw new ArgumentException("Repository name must be lowercase");
+      }
+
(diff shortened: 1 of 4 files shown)


HofmeisterAn (changes requested):
Can we add that as a guard methode to: https://github.com/testcontainers/testcontainers-dotnet/blob/39671c01d057567ae0ad505925236be27550982c/src/Testcontainers/Images/DockerImage.cs#L42? Then it's easier to test too. What do think?

michal-korniak:
Much easier :) I moved the logic to the class you mentioned and I added some tests to check if it work properly.

HofmeisterAn:
> easier to test too. What do th

Cool, thanks. I'll do the review and merge later that day 👏.

HofmeisterAn (approved):

HofmeisterAn:
I fixed the unit test, thanks (https://github.com/testcontainers/testcontainers-dotnet/commit/dcd1858046059478bb3cd1bca8acfb6a48d69854).

Issue #509, opened 2022-07-10
Check if the Docker endpoint is available


HofmeisterAn:
**Is your feature request related to a problem? Please describe.**
Testcontainers does not check if the Docker endpoint `IDockerEndpointAuthenticationProvider` is available. The strategy may detect a wrong configuration.

**Describe the solution you'd like**
Use the Docker client to ping the endpoint first:
https://github.com/testcontainers/testcontainers-dotnet/blob/95b647e3e895b4d7f0dfc852fc4dba480430b904/src/Testcontainers/Builders/DockerEndpointAuthenticationProvider.cs#L18

**Describe alternatives you've considered**
\-

**Additional context**
\-

PR #514, opened 2022-07-18
feat(#509): Check if the authentication provider can establish a Docker endpoint connection

HofmeisterAn:


HofmeisterAn:
Closes #509.

PR #521, opened 2022-07-26
Add configuration of Mssql server

aygalinc:
Allow usage of custom user & database catalog

### DIFF
--- a/src/Testcontainers/Configurations/Modules/Databases/MsSqlTestcontainerConfiguration.cs
+++ b/src/Testcontainers/Configurations/Modules/Databases/MsSqlTestcontainerConfiguration.cs
@@ -33,22 +33,22 @@ namespace DotNet.Testcontainers.Configurations
     /// <inheritdoc />
     public override string Database
     {
-      get => "master";
-      set => throw new NotImplementedException();
+      get => this.Environments["MSSQL_DATABASE"];
+      set => this.Environments["MSSQL_DATABASE"] = value;
     }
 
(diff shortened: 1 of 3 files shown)


HofmeisterAn (changes requested):
The unit tests (`MsSqlTestcontainerTest`) can't connect to the MsSql instance anymore. The wait strategy does not finish: https://github.com/testcontainers/testcontainers-dotnet/blob/1550fe64a711d1c39a74a10521600fda773ebdc9/src/Testcontainers/Configurations/Modules/Databases/MsSqlTestcontainerConfiguration.cs#L55

aygalinc:
How test Can pass on Windows & fail on unix ? There is a custom setup Somewhere ? I think that all test relies on the same test fixture - and pswd + credential seems ok in the log no ?

HofmeisterAn:
> How test Can pass on Windows & fail on unix ?

Windows and Linux run different tests. Not all images are available on both operating systems. `MsSqlTestcontainerTest` runs on Linux.

> There is a custom setup Somewhere ? I think that all test relies on the same test fixture - and pswd + credential seems ok in the log no ?

Sorry, I don't understand what you mean by that.

enginexon:
@aygalinc 
Hi!

What is your implementation based on? I see you use environment variables

- MSSQL_DATABASE
- MSSQL_USER
- MSSQL_PASSWORD

but I can not find them in Microsoft documentation about running MS SQL Server containers. 

Regards,
Eugene

Issue #528, opened 2022-07-29
DockerEndpointAuthConfig can't initialize when specifying remote host through code


BenasB:
**Describe the bug**
The static class [`TestcontainersSettings`](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers/Configurations/TestcontainersSettings.cs) fails to initialize when using only `WithDockerEndpoint` or `TestcontainersSettings.OS` with endpoint (this is what I call "through code" here as opposed to environment variables or poperties files). The culprit is the private property [`DockerEndpointAuthConfig`](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers/Configurations/TestcontainersSettings.cs#L27). If I understand correctly, none of the AuthenticationProviders are matched during `Where()`s and we end up calling a `First()` on an empty collection.

This is the error:
```
  Error Message:
   System.TypeInitializationException : The type initializer for 'DotNet.Testcontainers.Configurations.TestcontainersSettings' threw an exception.
---- System.InvalidOperationException : Sequence contains no elements
  Stack Trace:
     at DotNet.Testcontainers.Configurations.TestcontainersSettings.set_ResourceReaperEnabled(Boolean value)
   at Bets.Testcontainers.Test.UnitTest1.Test1() in D:\temp-test-containers\UnitTest1.cs:line 13
--- End of stack trace from previous location ---
----- Inner Stack Trace -----
   at System.Linq.ParallelEnumerable.GetOneWithPossibleDefault[TSource](QueryOperator`1 queryOp, Boolean throwIfTwo, Boolean defaultIfEmpty)
   at System.Linq.ParallelEnumerable.First[TSource](ParallelQuery`1 source)
   at DotNet.Testcontainers.Configurations.TestcontainersSettings..cctor()
```

The error is a bit misleading as at first I thought there was a problem with `ResourceReaperEnabled`, but it turns out just the static class constructor fails and `ResourceReaperEnabled` is only there because it was my first call to `TestcontainersSettings`.

I was running this in a **machine without docker installed**. When running it in a machine with docker installed, the problem was not there because I think it detected my local docker host (even though I specified a remote host with `WithDockerEndpoint`) and it failed elsewhere (no connection setup from that machine to the remote docker host, which of course fails)

**To Reproduce**
1. Perform steps on a machine without docker
2. Write a simple test class with a `WithDockerEndpoint` when configuring the test container:
```csharp
[...]

Issue #529, opened 2022-07-29
More descriptive error message when failing to resolve the Docker daemon


Xitric:
**Describe the bug**

A few of my colleagues have an odd setup of Docker, which led to a minor configuration mistake that caused testcontainers to fail resolving the Docker daemon. This has uncovered a small shortcoming in the `TestcontainersSettings` implementation when resolving `DockerEndpointAuthConfig`:

https://github.com/testcontainers/testcontainers-dotnet/blob/ff76e0fb4ff00bc81b657fda982ab8c1608f6fd3/src/Testcontainers/Configurations/TestcontainersSettings.cs#L27-L33

If there is no daemon at `npipe://./pipe/docker_engine` or `unix:/var/run/docker.sock`, only the `DOCKER_HOST` environment variable is left to consider. If that environment variable is missing or incorrect, the resolution fails with a very unhelpful error message:

```
OneTimeSetUp: System.TypeInitializationException : The type initializer for 'DotNet.Testcontainers.Configurations.TestcontainersSettings' threw an exception.
  ----> System.InvalidOperationException : Sequence contains no elements
  Exception doesn't have a stacktrace
```

**To Reproduce**

1. Ensure that the Docker daemon is not available at `npipe://./pipe/docker_engine` or `unix:/var/run/docker.sock`
2. Configure a DOCKER_HOST environment variable that does not point at a proper Docker daemon
3. Create a small test calling `new TestcontainersBuilder()`
4. Run the test and observe the exception during the constructor call above

**Expected behavior**

What we would like is for the code to handle the case of a missing Docker daemon gracefully, and present the user with a meaningful error (and certainly not an empty stacktrace - this might be due to running the code with `.AsParallel()`). This would have helped us to identify the error (an outdated `DOCKER_HOST` environment variable) much faster.

**Versions**

- Any version of Docker will do
- Testcontainers version 2.1.0

**Additional context**

None.

HofmeisterAn:
> A few of my colleagues have an odd setup of Docker, which led to a minor configuration mistake that caused testcontainers to fail resolving the Docker daemon.

It looks like not just your colleagues use this odd setup, see #528 😀. What do you think? Should we use `FirstOrDefault` instead of `First` and check while building the configuration if the endpoint is set? E.g. here:

[...]

PR #545, opened 2022-08-01
feat(#528): Do not require the Docker host configuration on TestcontainersSettings initialization

HofmeisterAn:
@Xitric @BenasB what do you think about this?

### DIFF
--- a/src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
+++ b/src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
@@ -75,6 +75,9 @@ namespace DotNet.Testcontainers.Builders
     /// <inheritdoc />
     public Task<string> Build()
     {
+      _ = Guard.Argument(this.DockerResourceConfiguration.DockerEndpointAuthConfig, nameof(IDockerResourceConfiguration.DockerEndpointAuthConfig))
+        .NotNull();
+
       ITestcontainersClient client = new TestcontainersClient(this.DockerResourceConfiguration.SessionId, this.DockerResourceConfiguration.DockerEndpointAuthConfig, TestcontainersSettings.Logger);
       return client.BuildAsync(this.DockerResourceConfiguration);
     }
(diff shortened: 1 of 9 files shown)


src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
```diff
@@ -75,6 +75,9 @@ public IImageFromDockerfileBuilder WithBuildArgument(string name, string value)
     /// <inheritdoc />
     public Task<string> Build()
     {
+      _ = Guard.Argument(this.DockerResourceConfiguration.DockerEndpointAuthConfig, nameof(IDockerResourceConfiguration.DockerEndpointAuthConfig))
```
BenasB, line 78: One small question: why discard assignment here and not just calling the function alone? that is why `_ =` at the front?

BenasB (approved):
It seems good. If I understand correctly, we allow `DockerEndpointAuthConfig` to become null at the start, but then enforce it to have a value in the end when building (it has to come from somewhere) - this makes sense. I don't think I can test this on my actual setup because we have limited access to nuget packages (until it is released officially).

P.S. I haven't had the chance to contribute to open source projects yet, so I'm not much of an expert.

src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
vlaskal: If I undersand correctly it is new best practice to show that the result is useless for reader or even for compiler. Zou can find out more here: https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards 
I would like to know if there is some real compiler optimiyation or is not.

HofmeisterAn: As far as I know IL is the same (no compiler optimization).

HofmeisterAn:
ab7fbb4171c1a51e69697c0d9ea39e09529469ac, closes #528, #529.

PR #548, opened 2022-08-02
feat(#370) Added mTLS support to docker endpoint

vlaskal:
This PR add two options for docker endpoint authentication.
1. it is support of simple secured endpoint by TLS
2. it is support of client authentication via mTLS

PR contains 2 commits on for each option.

This PR is draft only to review approach of implementation.

### DIFF
--- a/src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
+++ b/src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
@@ -1,0 +1,95 @@
+namespace DotNet.Testcontainers.Builders
+{
+  using System;
+  using System.IO;
+  using System.Linq;
+  using System.Security.Cryptography.X509Certificates;
+  using Docker.DotNet.X509;
+  using DotNet.Testcontainers.Configurations;
+  using Org.BouncyCastle.Crypto;
+  using Org.BouncyCastle.OpenSsl;
+  using Org.BouncyCastle.Pkcs;
+  using Org.BouncyCastle.Security;
+  using Org.BouncyCastle.X509;
+
+  /// <inheritdoc cref="IDockerRegistryAuthenticationProvider" />
+  internal sealed class MTlsEndpointAuthenticationProvider : TlsEndpointAuthenticationProvider
+  {
+    private static readonly X509CertificateParser CertificateParser = new X509CertificateParser();
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="MTlsEndpointAuthenticationProvider" /> class.
+    /// </summary>
+    public MTlsEndpointAuthenticationProvider()
+      : this(PropertiesFileConfiguration.Instance, EnvironmentConfiguration.Instance)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="MTlsEndpointAuthenticationProvider" /> class.
+    /// </summary>
+    /// <param name="customConfigurations">A list of custom configurations.</param>
+    public MTlsEndpointAuthenticationProvider(params ICustomConfiguration[] customConfigurations)
+      : base(customConfigurations)
+    {
+    }
+
+    /// <inheritdoc />
+    public override bool IsApplicable()
+    {
+      var certificatesFiles = new[] { ClientCertificateFileName, ClientCertificateKeyFileName };
+      return this.TlsEnabled && this.TlsVerifyEnabled && certificatesFiles.Select(fileName => Path.Combine(this.CertificatesDirectoryPath, fileName)).All(File.Exists);
+    }
+
+    /// <inheritdoc />
+    public override IDockerEndpointAuthenticationConfiguration GetAuthConfig()
+    {
+      var credentials = new CertificateCredentials(this.GetClientCertificate());
+      credentials.ServerCertificateValidationCallback = this.ServerCertificateValidationCallback;
+      return new DockerEndpointAuthenticationConfiguration(this.DockerEngine, credentials);
+    }
+
+    /// <inheritdoc />
+    protected override X509Certificate2 GetClientCertificate()
+    {
+      var clientCertificateFilePath = Path.Combine(this.CertificatesDirectoryPath, ClientCertificateFileName);
+      var clientCertificateKeyFilePath = Path.Combine(this.CertificatesDirectoryPath, ClientCertificateKeyFileName);
+      return CreateFromPemFile(clientCertificateFilePath, clientCertificateKeyFilePath);
+    }
+
+    private static X509Certificate2 CreateFromPemFile(string certPemFilePath, string keyPemFilePath)
+    {
+      if (!File.Exists(certPemFilePath))
+      {
+        throw new FileNotFoundException(certPemFilePath);
+      }
+
+      if (!File.Exists(keyPemFilePath))
+      {
+        throw new FileNotFoundException(keyPemFilePath);
+      }
+
+      using (var keyPairStream = new StreamReader(keyPemFilePath))
+      {
+        var store = new Pkcs12StoreBuilder().Build();
+
+        var certificate = CertificateParser.ReadCertificate(File.ReadAllBytes(certPemFilePath));
+
+        var password = Guid.NewGuid().ToString("D");
+
+        var keyPair = (AsymmetricCipherKeyPair)new PemReader(keyPairStream).ReadObject();
+
+        var certificateEntry = new X509CertificateEntry(certificate);
+
+        var keyEntry = new AsymmetricKeyEntry(keyPair.Private);
+        store.SetKeyEntry(certificate.SubjectDN + "_key", keyEntry, new[] { certificateEntry });
+
+        using (var certificateStream = new MemoryStream())
+        {
+          store.Save(certificateStream, password.ToCharArray(), new SecureRandom());
+          return new X509Certificate2(Pkcs12Utilities.ConvertToDefiniteLength(certificateStream.ToArray()), password);
+        }
+      }
+    }
+  }
+}
--- a/src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
+++ b/src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
@@ -1,0 +1,115 @@
+namespace DotNet.Testcontainers.Builders
+{
+  using System;
+  using System.IO;
+  using System.Linq;
+  using System.Net;
+  using System.Net.Security;
+  using System.Security.Cryptography.X509Certificates;
+  using DotNet.Testcontainers.Configurations;
+
+  /// <inheritdoc cref="IDockerRegistryAuthenticationProvider" />
+  internal class TlsEndpointAuthenticationProvider : DockerEndpointAuthenticationProvider
+  {
+    protected const string CaCertificateFileName = "ca.pem";
+
+    protected const string ClientCertificateFileName = "cert.pem";
+
+    protected const string ClientCertificateKeyFileName = "key.pem";
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TlsEndpointAuthenticationProvider" /> class.
+    /// </summary>
+    public TlsEndpointAuthenticationProvider()
+      : this(PropertiesFileConfiguration.Instance, EnvironmentConfiguration.Instance)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TlsEndpointAuthenticationProvider" /> class.
+    /// </summary>
+    /// <param name="customConfigurations">A list of custom configurations.</param>
+    public TlsEndpointAuthenticationProvider(params ICustomConfiguration[] customConfigurations)
+      : this(customConfigurations
+        .OrderByDescending(item => item.GetDockerTlsVerify())
+        .ThenByDescending(item => item.GetDockerTls())
+        .DefaultIfEmpty(new PropertiesFileConfiguration(Array.Empty<string>()))
+        .First())
+    {
+    }
+
+    private TlsEndpointAuthenticationProvider(ICustomConfiguration customConfiguration)
+    {
+      this.TlsEnabled = customConfiguration.GetDockerTls() || customConfiguration.GetDockerTlsVerify();
+      this.TlsVerifyEnabled = customConfiguration.GetDockerTlsVerify();
+      this.CertificatesDirectoryPath = customConfiguration.GetDockerCertPath() ?? Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".docker");
+      this.DockerEngine = customConfiguration.GetDockerHost() ?? new Uri("tcp://localhost:2376");
+    }
+
+    protected bool TlsEnabled { get; }
+
+    protected bool TlsVerifyEnabled { get; }
+
+    protected string CertificatesDirectoryPath { get; }
+
+    protected Uri DockerEngine { get; }
+
+    /// <inheritdoc />
+    public override bool IsApplicable()
+    {
+      return this.TlsEnabled;
+    }
+
+    /// <inheritdoc />
+    public override IDockerEndpointAuthenticationConfiguration GetAuthConfig()
+    {
+      var credentials = new TlsCredentials();
+      credentials.ServerCertificateValidationCallback = this.ServerCertificateValidationCallback;
+      return new DockerEndpointAuthenticationConfiguration(this.DockerEngine, credentials);
+    }
+
+    /// <summary>
+    /// Gets the root certificate authority (CA).
+    /// </summary>
+    /// <returns>The root certificate authority (CA).</returns>
+    protected virtual X509Certificate2 GetCaCertificate()
+    {
+      return new X509Certificate2(Path.Combine(this.CertificatesDirectoryPath, CaCertificateFileName));
+    }
+
+    /// <summary>
+    /// Gets the client certificate.
+    /// </summary>
+    /// <returns>The client certificate.</returns>
+    protected virtual X509Certificate2 GetClientCertificate()
+    {
+      return null;
+    }
+
+    /// <inheritdoc cref="ServicePointManager.ServerCertificateValidationCallback" />
+    protected virtual bool ServerCertificateValidationCallback(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)
+    {
+      switch (sslPolicyErrors)
+      {
+        case SslPolicyErrors.None:
+          return true;
+        case SslPolicyErrors.RemoteCertificateNameMismatch:
+        case SslPolicyErrors.RemoteCertificateNotAvailable:
+          return false;
+        case SslPolicyErrors.RemoteCertificateChainErrors:
+        default:
+          using (var caCertificate = this.GetCaCertificate())
+          {
+            var validationChain = new X509Chain();
+            validationChain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
+            validationChain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
+            validationChain.ChainPolicy.ExtraStore.Add(caCertificate);
+            validationChain.ChainPolicy.ExtraStore.AddRange(chain.ChainElements.OfType<X509ChainElement>().Select(element => element.Certificate).ToArray());
+            var isVerified = validationChain.Build(new X509Certificate2(certificate));
+            var isSignedByExpectedRoot = validationChain.ChainElements[validationChain.ChainElements.Count - 1].Certificate.RawData.SequenceEqual(caCertificate.RawData);
+            return isVerified && isSignedByExpectedRoot;
+          }
+      }
+    }
+  }
+}
--- a/src/Testcontainers/Configurations/TestcontainersSettings.cs
+++ b/src/Testcontainers/Configurations/TestcontainersSettings.cs
@@ -21,7 +21,14 @@ namespace DotNet.Testcontainers.Configurations
     private static readonly IDockerImage RyukContainerImage = new DockerImage("testcontainers/ryuk:0.3.4");
 
     private static readonly IDockerEndpointAuthenticationConfiguration DockerEndpointAuthConfig =
-      new IDockerEndpointAuthenticationProvider[] { new EnvironmentEndpointAuthenticationProvider(), new NpipeEndpointAuthenticationProvider(), new UnixEndpointAuthenticationProvider() }
+      new IDockerEndpointAuthenticationProvider[]
+        {
+          new MTlsEndpointAuthenticationProvider(),
+          new TlsEndpointAuthenticationProvider(),
+          new EnvironmentEndpointAuthenticationProvider(),
+          new NpipeEndpointAuthenticationProvider(),
+          new UnixEndpointAuthenticationProvider(),
+        }
         .AsParallel()
         .Where(authProvider => authProvider.IsApplicable())
         .Where(authProvider => authProvider.IsAvailable())
--- a/tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
+++ b/tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
@@ -1,0 +1,27 @@
+﻿namespace DotNet.Testcontainers.Tests.Fixtures
+{
+  using System.Collections.Generic;
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Containers;
+  using JetBrains.Annotations;
+
+  [UsedImplicitly]
+  public sealed class DockerMTlsFixture : ProtectDockerDaemonSocket
+  {
+    public DockerMTlsFixture()
+      : base(new TestcontainersBuilder<TestcontainersContainer>())
+    {
+    }
+
+    public override IList<string> CustomProperties
+    {
+      get
+      {
+        var customProperties = base.CustomProperties;
+        customProperties.Add("docker.tls=false");
+        customProperties.Add("docker.tls.verify=true");
+        return customProperties;
+      }
+    }
+  }
+}
--- a/tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerTlsFixture.cs
+++ b/tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerTlsFixture.cs
@@ -1,0 +1,28 @@
+﻿namespace DotNet.Testcontainers.Tests.Fixtures
+{
+  using System.Collections.Generic;
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Containers;
+  using JetBrains.Annotations;
+
+  [UsedImplicitly]
+  public sealed class DockerTlsFixture : ProtectDockerDaemonSocket
+  {
+    public DockerTlsFixture()
+      : base(new TestcontainersBuilder<TestcontainersContainer>()
+        .WithCommand("--tlsverify=false"))
+    {
+    }
+
+    public override IList<string> CustomProperties
+    {
+      get
+      {
+        var customProperties = base.CustomProperties;
+        customProperties.Add("docker.tls=true");
+        customProperties.Add("docker.tls.verify=false");
+        return customProperties;
+      }
+    }
+  }
+}
--- a/tests/Testcontainers.Tests/Fixtures/Containers/Unix/ProtectDockerDaemonSocket.cs
+++ b/tests/Testcontainers.Tests/Fixtures/Containers/Unix/ProtectDockerDaemonSocket.cs
@@ -1,0 +1,71 @@
+﻿namespace DotNet.Testcontainers.Tests.Fixtures
+{
+  using System;
+  using System.Collections.Generic;
+  using System.IO;
+  using System.Threading.Tasks;
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Containers;
+  using DotNet.Testcontainers.Images;
+  using Xunit;
+
+  public abstract class ProtectDockerDaemonSocket : IAsyncLifetime
+  {
+    public const string DockerVersion = "20.10.18";
+
+    private const string CertsDirectoryName = "certs";
+
+    private const ushort TlsPort = 2376;
+
+    private readonly string hostCertsDirectoryPath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("D"), CertsDirectoryName);
+
+    private readonly string containerCertsDirectoryPath = Path.Combine("/", CertsDirectoryName);
+
+    private readonly IDockerImage image = new DockerImage(string.Empty, "docker", DockerVersion + "-dind");
+
+    private readonly ITestcontainersContainer container;
+
+    protected ProtectDockerDaemonSocket(ITestcontainersBuilder<TestcontainersContainer> containerConfiguration)
+    {
+      this.container = containerConfiguration
+        .WithImage(this.image)
+        .WithPrivileged(true)
+        .WithExposedPort(TlsPort)
+        .WithPortBinding(TlsPort, true)
+        .WithBindMount(this.hostCertsDirectoryPath, this.containerCertsDirectoryPath, AccessMode.ReadWrite)
+        .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(TlsPort))
+        .Build();
+    }
+
+    public virtual IList<string> CustomProperties
+    {
+      get
+      {
+        var customProperties = new List<string>();
+        customProperties.Add($"docker.host={this.TcpEndpoint}");
+        customProperties.Add($"docker.cert.path={Path.Combine(this.hostCertsDirectoryPath, "client")}");
+        return customProperties;
+      }
+    }
+
+    private Uri TcpEndpoint
+    {
+      get
+      {
+        return new UriBuilder("tcp", this.container.Hostname, this.container.GetMappedPublicPort(TlsPort)).Uri;
+      }
+    }
+
+    public Task InitializeAsync()
+    {
+      _ = Directory.CreateDirectory(this.hostCertsDirectoryPath);
+      return this.container.StartAsync();
+    }
+
+    public Task DisposeAsync()
+    {
+      return this.container.DisposeAsync().AsTask();
+    }
+  }
+}
--- a/tests/Testcontainers.Tests/Unit/Configurations/DockerEndpointAuthenticationProviderTest.cs
+++ b/tests/Testcontainers.Tests/Unit/Configurations/DockerEndpointAuthenticationProviderTest.cs
@@ -2,6 +2,7 @@ namespace DotNet.Testcontainers.Tests.Unit
 {
   using System;
   using System.Collections.Generic;
+  using System.IO;
   using System.Runtime.InteropServices;
   using DotNet.Testcontainers.Builders;
   using DotNet.Testcontainers.Configurations;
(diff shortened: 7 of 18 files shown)


vlaskal:
I am not sure why few tests failed when locally run fine. I think that my code is out of test scope.

What this PR miss are:

- [x] Test strategy for new `TlsEndpointAuthenticationProvider` and `MTlsEndpointAuthenticationProvider`
- [x] Update build agents to be able to test these providers
- [x] Cleanup based on last PRs
- [x] Review of supportability in .NET Standard 2.1 (Do we test both .NET Standards?)
- [x] Used Custom configurations

src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,164 @@
+namespace DotNet.Testcontainers.Builders
```
vlaskal, line 1: This handles MTLS where is client verified on server side

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,73 @@
+namespace DotNet.Testcontainers.Builders
```
vlaskal, line 1: This file handles just secured connection to the server side.

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+      return new DockerEndpointAuthenticationConfiguration(this.dockerEngine, new TlsCredentials(this.ServerCertificateValidationCallback));
+    }
+
+    private bool ServerCertificateValidationCallback(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)
+    {
+      switch (sslPolicyErrors)
+      {
+        case SslPolicyErrors.None:
+          return true;
```
vlaskal, line 95: In this case CA certificate is used from system certificate storage and does not need to be in extra file.

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+        case SslPolicyErrors.RemoteCertificateNotAvailable:
+        case SslPolicyErrors.RemoteCertificateNameMismatch:
+          return false;
+        case SslPolicyErrors.RemoteCertificateChainErrors:
+        default:
+          var validationChain = new X509Chain();
```
vlaskal: This case means that validation failed and CA needs to be loaded from ca.pem file in defined storage.

src/Testcontainers/Configurations/Credentials/TlsCredentials.cs
```diff
@@ -0,0 +1,31 @@
...
+    public TlsCredentials([CanBeNull] RemoteCertificateValidationCallback serverCertificateValidationCallback)
+    {
+      this.ServerCertificateValidationCallback = serverCertificateValidationCallback;
+    }
+
+    public RemoteCertificateValidationCallback ServerCertificateValidationCallback { get; set; }
+
+    public override bool IsTlsCredentials()
+    {
+      return true;
```
vlaskal: This class is here only because of IsTlsCredentials needs to be true and there is no other way how to set it up.

src/Testcontainers/Configurations/CustomConfiguration.cs
```diff
@@ -18,6 +19,52 @@ protected Uri GetDockerHost(string propertyName)
       return this.properties.TryGetValue(propertyName, out var propertyValue) && Uri.TryCreate(propertyValue, UriKind.RelativeOrAbsolute, out var dockerHost) ? dockerHost : null;
     }
 
+    protected bool? GetDockerTls(string propertyName)
+    {
```
vlaskal: Question: Do we care if the value was explicitly set on certain level in case of bool value? I used Nullable<bool> because  I do not know if it was set up in properties files or in environment etc. 
What do you think?

src/Testcontainers/Configurations/CustomConfiguration.cs
```diff
@@ -18,6 +19,52 @@ protected Uri GetDockerHost(string propertyName)
...
+      {
+        return null;
+      }
+
+      return propertyValue;
+    }
+
     protected bool GetRyukDisabled(string propertyName)
     {
       return this.properties.TryGetValue(propertyName, out var propertyValue) && bool.TryParse(propertyValue, out var ryukDisabled) && ryukDisabled;
```
vlaskal: In case we care if the value was explictly set, then this implementation is not correct. What do you think. Do we care?

src/Testcontainers/Configurations/ICustomConfiguration.cs
```diff
@@ -17,6 +17,28 @@ internal interface ICustomConfiguration
     [CanBeNull]
     Uri GetDockerHost();
 
+    /// <summary>
+    /// Gets the Docker uses TLS.
+    /// </summary>
+    /// <returns>The Docker uses TLS.</returns>
+    /// <remarks>https://www.testcontainers.org/features/configuration/#customizing-docker-host-detection.</remarks>
+    bool? GetDockerTls();
```
vlaskal: I am not sure but most probably it should be rather a property than method. It looks like Java pattern of get / set methods. What do you think?

src/Testcontainers/Configurations/TestcontainersSettings.cs
```diff
@@ -25,7 +25,14 @@ public static class TestcontainersSettings
     private static readonly ICustomConfiguration EnvironmentConfiguration = new EnvironmentConfiguration();
 
     private static readonly IDockerEndpointAuthenticationConfiguration DockerEndpointAuthConfig =
-      new IDockerEndpointAuthenticationProvider[] { new EnvironmentEndpointAuthenticationProvider(), new NpipeEndpointAuthenticationProvider(), new UnixEndpointAuthenticationProvider() }
+      new IDockerEndpointAuthenticationProvider[]
```
vlaskal, line 24: This solution brings an idea to me. Do we want to open it for 3rd party customization? Like to add other providers from users. 
Maybe same approach can be used for custom configuration.

tests/Testcontainers.Tests/Unit/Configurations/CustomConfigurationTest.cs
```diff
@@ -32,6 +35,49 @@ public void GetDockerHostCustomConfiguration(string propertyName, string propert
...
+      [InlineData("DOCKER_TLS", "", null)]
+      [InlineData("DOCKER_TLS", "0", false)]
+      [InlineData("DOCKER_TLS", "false", false)]
+      [InlineData("DOCKER_TLS", "FALSE", false)]
+      [InlineData("DOCKER_TLS", "1", true)]
+      [InlineData("DOCKER_TLS", "TRUE", true)]
+      [InlineData("DOCKER_TLS", "true", true)]
+      public void GetDockerTlsCustomConfiguration(string propertyName, string propertyValue, bool? expected)
+      {
+        SetEnvironmentVariable(propertyName, propertyValue);
```
vlaskal: This is something what I want to discuss with you. Becasue in case that something is set up via ENV variables then this  and other tests can break. Which approach is best to use?

src/Testcontainers/Configurations/CustomConfiguration.cs
```diff
@@ -14,6 +15,37 @@ protected CustomConfiguration(IReadOnlyDictionary<string, string> properties)
...
+      }
+      catch (Exception)
+      {
+        return null;
+      }
+    }
+
+    protected string GetDockerCertPath(string propertyName)
+    {
+      _ = this.properties.TryGetValue(propertyName, out var propertyValue);
```
HofmeisterAn:
```csharp
_ = this.properties.TryGetValue(propertyName, out var propertyValue);
return propertyValue;
```

src/Testcontainers/Configurations/CustomConfiguration.cs
```diff
@@ -25,23 +57,38 @@ protected Uri GetDockerHost(string propertyName)
       return this.properties.TryGetValue(propertyName, out var propertyValue) && Uri.TryCreate(propertyValue, UriKind.RelativeOrAbsolute, out var dockerHost) ? dockerHost : null;
     }
 
-    protected JsonDocument GetDockerAuthConfig(string propertyName)
+    protected bool? GetDockerTls(string propertyName)
     {
-      _ = this.properties.TryGetValue(propertyName, out var propertyValue);
-
-      if (string.IsNullOrEmpty(propertyValue))
+      if (!this.properties.TryGetValue(propertyName, out var propertyValue) || string.IsNullOrWhiteSpace(propertyValue))
```
HofmeisterAn:
```csharp
_ = this.properties.TryGetValue(propertyName, out var propertyValue);
return "1".Equals(propertyValue, StringComparison.OrdinalIgnoreCase) || (bool.TryParse(propertyValue, out var tlsEnabled) && tlsEnabled);
```

src/Testcontainers/Configurations/CustomConfiguration.cs
```diff
@@ -25,23 +57,38 @@ protected Uri GetDockerHost(string propertyName)
...
+        return dockerTls;
       }
-      catch (Exception)
+
+      return propertyValue == "1";
+    }
+
+    protected bool? GetDockerTlsVerify(string propertyName)
+    {
+      if (!this.properties.TryGetValue(propertyName, out var propertyValue) || string.IsNullOrWhiteSpace(propertyValue))
```
HofmeisterAn:
```csharp
_ = this.properties.TryGetValue(propertyName, out var propertyValue);
return "1".Equals(propertyValue, StringComparison.OrdinalIgnoreCase) || (bool.TryParse(propertyValue, out var tlsVerifyEnabled) && tlsVerifyEnabled);
```

src/Testcontainers/Configurations/ICustomConfiguration.cs
```diff
@@ -27,12 +43,18 @@ internal interface ICustomConfiguration
...
-    /// Gets the Docker registry authentication custom configuration.
+    /// Gets the Docker uses TLS.
     /// </summary>
-    /// <returns>The Docker authentication custom configuration.</returns>
-    /// <remarks>https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#access-an-image-from-a-private-container-registry.</remarks>
-    [CanBeNull]
-    JsonDocument GetDockerAuthConfig();
+    /// <returns>The Docker uses TLS.</returns>
+    /// <remarks>https://www.testcontainers.org/features/configuration/#customizing-docker-host-detection.</remarks>
+    bool? GetDockerTls();
```
HofmeisterAn: I would like to avoid null values (`bool GetDockerTls()`). This requires to update the unit tests.

src/Testcontainers/Configurations/ICustomConfiguration.cs
```diff
@@ -27,12 +43,18 @@ internal interface ICustomConfiguration
...
+
+    /// <summary>
+    /// Gets the Docker uses TLS and verifies the remote.
+    /// </summary>
+    /// <returns>The Docker uses TLS and verifies the remote.</returns>
+    /// <remarks>https://www.testcontainers.org/features/configuration/#customizing-docker-host-detection.</remarks>
+    bool? GetDockerTlsVerify();
```
HofmeisterAn: I would like to avoid null values (`bool GetDockerTlsVerify()`). This requires to update the unit tests.

HofmeisterAn (changes requested):
I would like to split this pull request into two smaller chunks. The first one should just contain the changes regarding `ICustomConfiguration` (reading the custom configuration values incl. tests). The second one the TLS implementation. It makes the review easier and we can focus just on the TLS part next. What do you think? In the meantime I'll setup the TLS test environment. Thanks again.

src/Testcontainers/Configurations/CustomConfiguration.cs
HofmeisterAn: I would like to avoid nullable and the additional checks. I think `||` is fine (`GetRyukDisabled` use a similar pattern).

src/Testcontainers/Configurations/ICustomConfiguration.cs
HofmeisterAn: A method is fine. It does a bit more than just holding or returning a value.

src/Testcontainers/Configurations/TestcontainersSettings.cs
HofmeisterAn: Yep, we can do that in the future. Until now, other devs can set the provider via `TestcontainersSettings.OS.DockerEndpointAuthConfig` too.

tests/Testcontainers.Tests/Unit/Configurations/CustomConfigurationTest.cs
HofmeisterAn: This shouldn't be a problem. We can even improve this, by adding the target to `SetEnvironmentVariable`: `Environment.SetEnvironmentVariable(propertyName, propertyValue, EnvironmentVariableTarget.Process)`. The tests are executed sequentially at the end (no other test will run).

src/Testcontainers/Configurations/CustomConfiguration.cs
HofmeisterAn: I don't think we need to care about that. Either someone set the value to `true` or in all other cases the value is `false`.

vlaskal:
> I would like to split this pull request into two smaller chunks. The first one should just contain the changes regarding `ICustomConfiguration` (reading the custom configuration values incl. tests). The second one the TLS implementation. It makes the review easier and we can focus just on the TLS part next. What do you think? In the meantime I'll setup the TLS test environment. Thanks again.

Ok, will try to prepare it soon.

HofmeisterAn:
We can use something like this to create a test instance:

```csharp
public sealed class GitHub : IAsyncLifetime
{
  private const string CertsDirectoryName = "certs";

  private static readonly string ContainerCertDirectoryPath = Path.Combine("/", CertsDirectoryName);

  private static readonly string HostCertDirectoryPath = Path.Combine(Path.GetTempPath(), CertsDirectoryName);

  private readonly ITestcontainersContainer tlsContainer = new TestcontainersBuilder<TestcontainersContainer>()
    .WithImage("docker:20.10-dind")
    .WithPrivileged(true)
    .WithEnvironment("DOCKER_CERT_PATH", ContainerCertDirectoryPath)
    .WithEnvironment("DOCKER_TLS_CERTDIR", ContainerCertDirectoryPath)
    .WithEnvironment("DOCKER_TLS", "1")
    .WithEnvironment("DOCKER_TLS_VERIFY", "1")
    .WithBindMount(HostCertDirectoryPath, ContainerCertDirectoryPath, AccessMode.ReadWrite)
    .Build();

  [Fact]
  public Task PullRequest548()
  {
    return Task.CompletedTask;
  }

  public Task InitializeAsync()
  {
    return this.tlsContainer.StartAsync();
  }

  public Task DisposeAsync()
  {
    return this.tlsContainer.DisposeAsync().AsTask();
  }
}
```

This generates the certificates too.

vlaskal:
Ok will try to use it. But I will update PR to latest master then will split it to two smaller chunks with tests.
Then we can deal with one feature in a time.

vlaskal:
@HofmeisterAn  I updated mTls implementation based on implementation of Tls in PR #597  including inheritance to not repeat code. Uou can look at implementation and chak it. Also here is a test which tests that mTls work.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerTlsFixture.cs
```diff
@@ -0,0 +1,61 @@
...
+  using System.IO;
+  using System.Threading.Tasks;
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Containers;
+  using JetBrains.Annotations;
+  using Xunit;
+
+  [UsedImplicitly]
+  public sealed class DockerTlsFixture : IAsyncLifetime
```
HofmeisterAn: Is `DockerTlsFixture` the right name? How can the tests TLS only.

vlaskal: I can add Tls test but it does not work for me. I will add Skip reason but just to see how it cann look like

HofmeisterAn: It is not necessary to push it, I just update the property and test it tomorrow.

vlaskal: I found out working test for simple TLS. So I added mTLS and TLS tests. Maybe soe inheritance would be good to add.

vlaskal: I added some inheritance to the tests so it looks better now.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/ProtectDockerDaemonSocket.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    private const string CertsDirectoryName = "certs";
+
+    private const ushort TlsPort = 2376;
+
+    private readonly ITestcontainersContainer container;
+
+    private readonly IDockerImage image = new DockerImage(string.Empty, "docker", DockerVersion + "-dind");
+
+    public ProtectDockerDaemonSocket(ITestcontainersBuilder<TestcontainersContainer> containerConfiguration)
```
HofmeisterAn: I split the fixtures into two classes `DockerMTlsFixture` and `DockerTlsFixture`. We can configure the build individual for each of them. I noticed that the env variables do not have any effect on the container configuration / the Docker daemon, so I removed them.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace DotNet.Testcontainers.Tests.Fixtures
+{
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Containers;
+  using JetBrains.Annotations;
+
+  [UsedImplicitly]
+  public sealed class DockerMTlsFixture : ProtectDockerDaemonSocket
+  {
+    public DockerMTlsFixture()
```
HofmeisterAn, line 11: Based on the Docker [docs](https://docs.docker.com/engine/security/protect-access/#other-modes) I would assume `DockerMTlsFixture` should use the following configuration: `tlsverify`, `tlscacert`, `tlscert`, `tlskey`. It would be nice to set the configuration explicit (even if it is the default configuration I believe)

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerTlsFixture.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace DotNet.Testcontainers.Tests.Fixtures
+{
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Containers;
+  using JetBrains.Annotations;
+
+  [UsedImplicitly]
+  public sealed class DockerTlsFixture : ProtectDockerDaemonSocket
```
HofmeisterAn, line 9: Based on the Docker [docs](https://docs.docker.com/engine/security/protect-access/#other-modes) I would assume `DockerTlsFixture` should use following configuration: `tls`, `tlscert`, `tlskey`. The same applies here as well as for `DockerTlsFixture`. I am not sure if `--tlsverify=false` corresponds to that configuration.

tests/Testcontainers.Tests/Unit/Containers/Unix/ProtectDockerDaemonSocketTest.cs
```diff
@@ -0,0 +1,84 @@
...
+  using DotNet.Testcontainers.Builders;
+  using DotNet.Testcontainers.Clients;
+  using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Tests.Fixtures;
+  using Microsoft.Extensions.Logging.Abstractions;
+  using Xunit;
+
+  public static class ProtectDockerDaemonSocketTest
+  {
+    private static ICustomConfiguration GetCustomConfiguration(ProtectDockerDaemonSocket protectDockerDaemonSocket, bool tlsEnabled, bool tlsVerifyEnabled)
```
HofmeisterAn: No matter which combination I use for `tlsEnabled` and `tlsVerifyEnabled` the tests always succeed. I would expect a different behaviour here.

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+  using System.Linq;
+  using System.Net.Security;
+  using System.Security.Cryptography.X509Certificates;
+  using DotNet.Testcontainers.Configurations;
+
+  /// <inheritdoc cref="IDockerRegistryAuthenticationProvider" />
+  internal class TlsEndpointAuthenticationProvider : DockerEndpointAuthenticationProvider
+  {
+    private const string DefaultUserDockerFolderName = ".docker";
+    private const string DefaultCaCertFileName = "ca.pem";
```
HofmeisterAn: Is `ca.pem` correct here? If I look into the Docker [docs](https://docs.docker.com/engine/security/protect-access/#other-modes) the client mode does not mention `tlscacert` for the TLS configuration.

HofmeisterAn (changes requested):
I think the pull request is on the right track. I would like to make sure that the configurations for the fixtures are correct, afterwards we can finalize the providers. Setting the fixtures explicit to `tlsverify` or `tls` should be the next step (dunno if that is possible via containers command args). Maybe we need to generate certificates by ourself and check them in to set the configuration explicit.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/ProtectDockerDaemonSocket.cs
vlaskal: Ok

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
vlaskal: The docker used for dockerd use a certain `entry-script.sh` script. So to set and turn of does not follow exactly parameters from docker documentation. I tested whole command line setup for tls only and even that does not work to me. I was happy to find out possibility to turn off `tlsverify`. This was mentioned by the script when you run the docker image without cert generation. I found accidentaly because my cert folder was somehow old and entry-point.sh script could not generate certs so started without tls at all.
Combination used:

```
      .WithCommand("dockerd",
        "--host=unix:///var/run/docker.sock",
        "--host=tcp://0.0.0.0:2376",
        "--tls",
        "--tlscacert=/certs/server/ca.pem",
        "--tlscert=/certs/server/cert.pem",
        "--tlskey=/certs/server/key.pem")
```
this run stil in tlsverify mode

```
      .WithCommand("dockerd",
        "--host=unix:///var/run/docker.sock",
        "--host=tcp://0.0.0.0:2376",
        "--tls",
        "--tlscert=/certs/server/cert.pem",
        "--tlskey=/certs/server/key.pem")
```

this does not start and complains that `--tlscacert` is missing.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerTlsFixture.cs
vlaskal: No it does not corresponds to docker configuration. It corresponds to image used to run run dockerd.

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
vlaskal: You are right - I run a lot of test with Docker CLI and CA is required for validation of the server certificate. I used it for current implementation.
1. try to validate cert against system CAs store
2. used ca.pem folder as CLI does (even docs said something else)

tests/Testcontainers.Tests/Unit/Containers/Unix/ProtectDockerDaemonSocketTest.cs
vlaskal: I will check but be carefull what the entrypoint.sh script of the docker image really does. I failed several times until I found that --tlsverify=false switch

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
HofmeisterAn: I double-checked the configuration. I tried it explicit as well. Probably the docs are incomprehensible (at least for me). I came to the same conclusion. I think the fixtures look good 👏.

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
HofmeisterAn: Ok, I will take a look at it that again. `ca.pem` sounds reasonable (that is what a client need), but reading the docs I would expect something different...

tests/Testcontainers.Tests/Unit/Containers/Unix/ProtectDockerDaemonSocketTest.cs
vlaskal: Haha, that was nice catch. I removed these flags because they are useless in commit 2e11eae.
These flags are used only in method `bool IDockerEndpointAuthenticationProvider.IsApplicable()` to ensure if this provider should be used when is registered in collection `TestcontainersSettings.DockerEndpointAuthConfig`. This collection is not used by tests.

HofmeisterAn: Makes sense 👍. I think we can use it to test `IsApplicable` too in the future. Something like:

```csharp
new[] { MTlsEndpointAuthenticationProvider(customConfiguration), TlsEndpointAuthenticationProvider(customConfiguration) }.First(...)
Assert.IsType<MTlsEndpointAuthenticationProvider>(...)
```

Then only the provider part is left. I will take a look at it then we can merge it I think. Or do you think anything else is left?

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
vlaskal: Based on [documentation](https://docs.docker.com/engine/security/protect-access/#other-modes) where client combinations are mentioned:

> Client modes
> tls: Authenticate server based on public/default CA pool
> tlsverify, tlscacert: Authenticate server based on given CA
> tls, tlscert, tlskey: Authenticate with client certificate, do not authenticate server based on given CA
> tlsverify, tlscacert, tlscert, tlskey: Authenticate with client certificate and authenticate server based on given CA

Then I understand that documentation this way:
* tls - try look for certificate in system trusted CA certificates storage ("based on public") and then try to use `ca.pem` file ("default CA") but to find one valid CA is enough to pass
* tlsverify, tlscacert - server is still set with `tls=true` only and in this case client verifies server certifcate against one specific CA stored in specified file or in default file which is `ca.pem`
* tls, tlscert, tlskey - tls is validated as is described in first bullet point, and client send cert for authetication because server requires it
* tlsverify, tlscacert, tlscert, tlskey - tlsverify is validated as is described in second bullet point, and client send cert for authetication because server requires it

So if system trusted root CAs storage does not contain any valid ceritificate and `ca.pem` file is not available then the docke CLI client complains that ca.pem file si not available.
This is my experience with Docke CLI.

tests/Testcontainers.Tests/Unit/Containers/Unix/ProtectDockerDaemonSocketTest.cs
vlaskal: IsApplicable is already tested by test `DockerEndpointAuthenticationProviderTest.AuthProviderShouldBeApplicable`. I think it's all.

src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+
+      using (var keyPairStream = new StreamReader(keyPemFilePath))
+      {
+        var store = new Pkcs12StoreBuilder().Build();
+
+        var certificate = CertificateParser.ReadCertificate(File.ReadAllBytes(certPemFilePath));
+
+        var password = Guid.NewGuid().ToString("D");
+
+        var keyPair = (AsymmetricCipherKeyPair)new PemReader(keyPairStream).ReadObject();
```
HofmeisterAn, line 80: I updated to this implementation. This makes the RegEx obsolete.

tests/Testcontainers.Tests/Unit/Configurations/DockerEndpointAuthenticationProviderTest.cs
```diff
@@ -35,7 +37,17 @@ public AuthProviderTestData()
       {
         var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
         var defaultConfiguration = new PropertiesFileConfiguration(Array.Empty<string>());
-        var dockerHostConfiguration = new PropertiesFileConfiguration(new[] { "docker.host=" + DockerHost });
+        var dockerHostConfiguration = new PropertiesFileConfiguration(new[] { $"docker.host={DockerHost}" });
+        var dockerTlsConfiguration = new PropertiesFileConfiguration(new[] { "docker.tls=true" });
+        var dockerMTlsConfiguration = new PropertiesFileConfiguration(new[] { "docker.tls.verify=true" });
+        this.Add(new object[] { new MTlsEndpointAuthenticationProvider(defaultConfiguration), false });
+        // this.Add(new object[] { new MTlsEndpointAuthenticationProvider(dockerMTlsConfiguration), true });
```
HofmeisterAn: I need to fix those 2 tests. They require the cert and key file. I think we can generate a temp dummy file for that.

HofmeisterAn (commented):
@vlaskal Would you take a look at my recent changes? I think I just need to fix 2 tests, then we are good to merge. What do you think?

src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
vlaskal: Yes, I think it is ok. Regex was there because there can be 2 tzpes of keys and each needs to be handled different way. But so far we did not hit that case.

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    protected const string ClientCertificateKeyFileName = "key.pem";
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TlsEndpointAuthenticationProvider" /> class.
+    /// </summary>
+    /// <param name="customConfigurations">A list of custom configurations.</param>
+    public TlsEndpointAuthenticationProvider(params ICustomConfiguration[] customConfigurations)
+      : this(customConfigurations
+        .OrderByDescending(item => item.GetDockerTls() && item.GetDockerHost() != null)
```
vlaskal: Why do you need ordering based on Tls. This does not reflect TlsVerify which is need just for mTls. I think it goes little bit  against. Tls is set to true and TlsVerify can be false. Then TlsVerify can be true but Tls can be false, but it does not matter becuase client take it as true of of the box.

vlaskal: Also I think that we should never change order and follow overriding rule which is ,I guess, Environment variable ->Testcontainer setting file -> (maybe in future) command line arguments -> settings set by code -> finall settings.
Probably for now it is ok, not sure if tlsVerify will not be affected then.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
```diff
@@ -0,0 +1,27 @@
...
+      : base(new TestcontainersBuilder<TestcontainersContainer>())
+    {
+    }
+
+    public override IList<string> CustomProperties
+    {
+      get
+      {
+        var customProperties = base.CustomProperties;
+        customProperties.Add("docker.tls=true");
```
vlaskal: I would suggest to set `docker.tls=false` because client should have automatic handling this to `true` based on `TlsVerify` only (based on docs).

src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+    public MTlsEndpointAuthenticationProvider(params ICustomConfiguration[] customConfigurations)
+      : base(customConfigurations)
+    {
+    }
+
+    /// <inheritdoc />
+    public override bool IsApplicable()
+    {
+      var certificatesFiles = new[] { ClientCertificateFileName, ClientCertificateKeyFileName };
+      return this.TlsEnabled && this.TlsVerifyEnabled && certificatesFiles.Select(fileName => Path.Combine(this.CertificatesDirectoryPath, fileName)).All(File.Exists);
```
vlaskal, line 41: This is not correct. `this.TlsEnabled` can be `false`.
```suggestion
      return this.TlsVerifyEnabled && certificatesFiles.Select(fileName => Path.Combine(this.CertificatesDirectoryPath, fileName)).All(File.Exists);
```

tests/Testcontainers.Tests/Unit/Configurations/DockerEndpointAuthenticationProviderTest.cs
vlaskal: Yep sounds good. I had that specific code like youhave right now but I removed it.

vlaskal:
> @vlaskal Would you take a look at my recent changes? I think I just need to fix 2 tests, then we are good to merge. What do you think?

Done. I think we are close.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
HofmeisterAn: Thats a good point. Then we can change it to `customConfiguration.GetDockerTls() || customConfiguration.GetDockerTlsVerify()`

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
HofmeisterAn: Yep, you are right, I'll update it tomorrow (incl. die others).

HofmeisterAn: I think the order does not matter here (at least for now). If `GetDockerTls()` or `GetDockerTlsVerify()` is not set, it is always false. Hence, we cannot override it. We pick the first configuration that enables it. To support to override the properties we need them nullable. Same for

https://github.com/testcontainers/testcontainers-dotnet/blob/78cf9a3b910ef73088142ebcc267d8127b54f68f/src/Testcontainers/Configurations/TestcontainersSettings.cs#L40

We can add that feature in a new PR (we need to switch the order in `TestcontainersSettings` too). But you are right, the ordering above is wrong.

----

If we make the properties nullable we can take the last one that is not null.

HofmeisterAn: I've updated the ordering. Before we release `2.2.0` we can refactor the custom properties and make them nullable. That's probably better anyway. We need to switch props and env configurations too.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/DockerMTlsFixture.cs
HofmeisterAn: It is now `false`

HofmeisterAn: I've changed it to the above suggestion.

src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
HofmeisterAn: TLS verify sets TLS to `true` too.

src/Testcontainers/Builders/TlsEndpointAuthenticationProvider.cs
vlaskal: I think we can leave it after that release. and do more regarding settings.

src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
vlaskal: Ok, I did not catch that.

vlaskal (commented):
It looks as final.

src/Testcontainers/Builders/MTlsEndpointAuthenticationProvider.cs
HofmeisterAn: You where right at the beginning. I changed that after your comment. Thanks.

HofmeisterAn (approved):

HofmeisterAn:
Closes #370.

PR #549, opened 2022-08-02
Testcontainer for CosmosDB Emulator

Yeseh:
I seem to be on the right track, so I'm opening a PR for some feedback.
General solution direction is to use a custom HttpClientHandler to query the database using the REST API instead of using the CosmosDB client SDK.

Here's my current view of the TODOs:
- [x] Setup basic framework for testcontainer
- [x] Implement HttpClient for use with the CosmosDB REST API
- [x] Create a database in the container
- [x] Find a better WaitStrategy (especially with high partition counts)
- [ ] Execute an arbitrary query
- [ ] Allow for multi-port binding in testcontainer to expose other CosmosDB endpoints (also needed for storage emulator)


src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+            "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==";
+
+        private const string CosmosDbDefaultAccountName = 
+            "localhost:8081";
+
+        public CosmosDbTestcontainerConfiguration()
+            : this(CosmosDbImage) { }
+
+        public CosmosDbTestcontainerConfiguration(string image)
+            : base(image, CosmosDbPort, CosmosDbPort)
```
HofmeisterAn: I recommend `base(image, CosmosDbPort)`. It is good practice to use a random assigned host port to avoid port clashes.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+    [PublicAPI]
+    public class CosmosDbTestcontainerConfiguration : TestcontainerDatabaseConfiguration
+    {
+        private const string CosmosDbImage = 
+            "mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator";
+
+        private const int CosmosDbPort = 8081;
+
+        private const string CosmosDbDefaultKey =
+            "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==";
```
HofmeisterAn: What kind of secret is that?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        private const string CosmosDbDefaultAccountName = 
+            "localhost:8081";
```
HofmeisterAn: If there is no user, I suggest to throw an exception (getter and setter `InvalidOperationException`). This looks like a connection string and should be part of `CosmosDbTestcontainer`.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+            get => CosmosDbDefaultKey;
+        }
+
+        public override string Database
+        {
+            get => this.Environments["AZURE_COSMOS_EMULATOR_DATABASE"];
+            set => this.Environments["AZURE_COSMOS_EMULATOR_DATABASE"] = value;
+        }
+
+        public string PartitionCount
```
HofmeisterAn: Maybe as integer?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            get => this.Environments["AZURE_COSMOS_EMULATOR_PARTITION_COUNT"];
+            set => this.Environments["AZURE_COSMOS_EMULATOR_PARTITION_COUNT"] = value;
+        }
+
+        public string EnableDataPersistence
```
HofmeisterAn: Maybe as boolean?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public string IPAddressOverride
+        {
+            get => this.Environments["AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE"];
+            set => this.Environments["AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE"] = value;
+        }
+
+        public override IWaitForContainerOS WaitStrategy => Wait.ForUnixContainer()
+            .UntilCommandIsCompleted("ls");
+            //.UntilMessageIsLogged(this.OutputConsumer.Stdout, "Started");
```
HofmeisterAn: You probably need to read `stdout`, here is an example: https://github.com/testcontainers/testcontainers-dotnet/discussions/438#discussion-3949974.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,110 @@
...
+  public sealed class CosmosDbTestcontainer : TestcontainerDatabase
+  {
+    private string CosmosUrl;
+
+    private HttpClient HttpClient;
+
+    internal CosmosDbTestcontainer(ITestcontainersConfiguration configuration, ILogger logger)
+        : base(configuration, logger)
+    {
+        CosmosUrl = $"https://{this.Username}.documents.azure.com/dbs";
```
HofmeisterAn: `this.Username` --> `this.Hostname` (it is not `localhost` for every Docker environment).

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,110 @@
...
+  using DotNet.Testcontainers.Configurations;
+  using Microsoft.Extensions.Logging;
+  using System.Text.Json;
+  using System;
+  using System.Threading;
+  using System.Text;
+
+  public sealed class CosmosDbTestcontainer : TestcontainerDatabase
+  {
+    private string CosmosUrl;
```
HofmeisterAn: `readonly`

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,110 @@
...
+    private HttpClient HttpClient;
+
+    internal CosmosDbTestcontainer(ITestcontainersConfiguration configuration, ILogger logger)
+        : base(configuration, logger)
+    {
+        CosmosUrl = $"https://{this.Username}.documents.azure.com/dbs";
+    }
+
+    public override string ConnectionString
+        => $"AccountEndpoint=https://{this.Username}:{this.Port}.documents.azure.com:443/;AccountKey={this.Password}";
```
HofmeisterAn: `this.Username` --> `this.Hostname` (it is not `localhost` for every Docker environment).

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,110 @@
...
+        string queryString, IEnumerable<KeyValuePair<string, string>> parameters = default)
+    {
+        Console.WriteLine("Executing query...");
+        var client = GetHttpClient();
+        var parJsonStr =JsonSerializer.Serialize(parameters);
+        var body = new { Query = queryString, Parameters = parJsonStr };
+        var reqBodyStr = JsonSerializer.Serialize(body);
+        var content = new StringContent(reqBodyStr);
+
+        var response = await client.PostAsync(CosmosUrl, content);
```
HofmeisterAn: Should this not use the `ConnectionString`? Then `CosmosUrl` is obsolete.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,110 @@
...
+  using System.Text.Json;
+  using System;
+  using System.Threading;
+  using System.Text;
+
+  public sealed class CosmosDbTestcontainer : TestcontainerDatabase
+  {
+    private string CosmosUrl;
+
+    private HttpClient HttpClient;
```
HofmeisterAn: `readonly`

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        return response;
+    }        
+
+    private HttpClient GetHttpClient() 
+    {
+      if (HttpClient != null) return HttpClient;
+
+      var handler = new CosmosDbHttpHandler(this.Password);
+      var client = new HttpClient(handler);
```
HofmeisterAn: I think this can be done in the constructor.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
HofmeisterAn: `await` is not necessary. Just `return client.PostAsync(CosmosUrl, content)`, remove `async`.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        return response;
+    }
+    // TODO: Call this implicitly on initialize
+    public async Task<HttpResponseMessage> CreateDatabaseAsync()
+    {
+        Console.WriteLine("Attempting to create database...");
+        var url = $"https://localhost:8081/dbs";
+        var client = GetHttpClient();
+        var jsonData = JsonSerializer.Serialize(new { id = string.IsNullOrEmpty(this.Database) ? "testdb" : this.Database });
```
HofmeisterAn: Why hardcoded `testdb`. I would put the default vaule in `CosmosDbTestcontainerConfiguration` and throw an exception if it is null or empty.

HofmeisterAn (changes requested):
Thanks for your contribution and pull request 🙏.

I have added some suggestions (or questions). I did not review `CosmosDbHttpHandler` incl. the tests. I will do that at the end. Please consider to fix the code smells (warnings) too, thanks.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
Yeseh: It's the default key for the cosmosDB emulator, it's in the docs so not leaking anything here.

HofmeisterAn: https://docs.microsoft.com/de-de/azure/cosmos-db/local-emulator?tabs=ssl-netstd21#sql-api

Yeseh:
Thanks for the thourough review! Will get back to this later this week hopefully.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
Yeseh: Hmm, I can't get this to work. The stdout wait strategy deadlocks for me, debugging in the wait strategy itself shows output remains empty. So I'm guessing the output stream is not coming through. Might be because I'm running linux containers on docker for windows atm?

Any ideas?

Yeseh:
Still WIP but improved a bit to model it after the recent work on the Azurite container. I came to the conclusion that it's kinda nuts to implement the client stuff in the container, can just let the user take care of that :).

ktjn:
@Yeseh You are not connecting the output consumer. Add:   .WithOutputConsumer(configuration.OutputConsumer). Also add a this.OutputConsumer = Consume.RedirectStdoutAndStderrToStream(new MemoryStream(), new MemoryStream());

If you want I can help out with this.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
ktjn: See below

Yeseh:
> @Yeseh You are not connecting the output consumer. Add: .WithOutputConsumer(configuration.OutputConsumer). Also add a this.OutputConsumer = Consume.RedirectStdoutAndStderrToStream(new MemoryStream(), new MemoryStream());
> 
> If you want I can help out with this.

Ah that makes a lot of sense. Be my guest with helping, i'm massively short on time atm :)

ktjn:
Ok. I currently have a working CosmosDbFixture which can be used to fix this. One thing to note is that the host is not always "localhost" depending on where docker is running. I can fork your repo or you can give me access to your witchevey you prefer.

ktjn:
@Yeseh Did some quick changes to make the test work. I think the mongo api should be separated into some subclass. Messy with all properties.

Yeseh:
@ktjn Thanks a lot! 

Definitely agree on the mongo part. Think the best route is a separate testcontainer altogether, because of all the differences. What do you think?

ktjn:
> @ktjn Thanks a lot!
> 
> Definitely agree on the mongo part. Think the best route is a separate testcontainer altogether, because of all the differences. What do you think?

Yes, at least do the sql mode first. Btw, I saw the build was stuck. It might have to do with "localhost" and running on linux.

I guess we should try to remove the cosmosdb client dependency and go for a pure http based test as well?

ktjn:
It's getting stuck at:
2022-09-06 20:51:19.295 +02:00 [INF] DotNet.Testcontainers.Tests.Initialized Attach "DotNet.Testcontainers.Configurations.RedirectStdoutAndStderrToStream" at Docker container "c9021df372bcf28a38a87946b07b6c077c3f22b0659e7ad7192a3639899a6860"
2022-09-06 20:51:19.297 +02:00 [INF] DotNet.Testcontainers.Tests.Initialized Start Docker container "c9021df372bcf28a38a87946b07b6c077c3f22b0659e7ad7192a3639899a6860"

I works fine on windows and wsl. I'll setup a linux vm and will try it there.

ktjn:
The test still works on a linux vm. Added a 5 min timeout so it doesn't run forever. Might need to add some logging to understand the problem.

HofmeisterAn:
Do you need any help? Let me know if I should take a look at something.

ktjn:
> Do you need any help? Let me know if I should take a look at something.

Thanks. No, just wanted to understand why the container didn't start correctly. But when adding some debug everything ran ok. I want a clean working baseline before I help @Yeseh clean this up.

ktjn:
@HofmeisterAn Can you please stop the build? my task.delay hack did fail the test but the wait strategy is still running? Can I set a timeout for the tests until this problem is found?

HofmeisterAn:
Sorry, just read your message now. BTW, you should assign a random host port `base(image, DefaultSqlApiPort, DefaultSqlApiPort)` should be `base(image, DefaultSqlApiPort)`.

ktjn:
> Sorry, just read your message now. BTW, you should assign a random host port `base(image, DefaultSqlApiPort, DefaultSqlApiPort)` should be `base(image, DefaultSqlApiPort)`.

Yes, will fix. Just wanted to build to be stable first. I set a 15 min timeout on ci build now. A max timeout value for wait strategy might be a good idea as well.

ktjn:
@HofmeisterAn @Yeseh I think we pretty much got it now. It is possible to remove the cosmos client dependency by doing raw rest calls to cosmos.

We might have the intermittent test failure when the container is shutting down for some reason.

What do you think?

ktjn:
Ok, seems to be a known error: 
https://github.com/microsoft/azure-pipelines-tasks/issues/15865

Maybe try a different image when running the tests?

ktjn:
Hmm.... cake didn't like ubuntu 22.

ktjn:
Ok, might be working now. Azurite test failed once as well.

ktjn:
This seems to be the real issue: https://github.com/Azure/azure-cosmos-db-emulator-docker/issues/45

ktjn:
@HofmeisterAn A suggestion. Just run the test on windows until the actual emulator bug is fixed. What do you think?

HofmeisterAn:
> @HofmeisterAn A suggestion. Just run the test on windows until the actual emulator bug is fixed. What do you think?

OC, sounds good. I assume the configuration is the same. Just move the test to the Windows namespace. BTW, I'll update to windows2022 and ubuntu-22.04 soon as Cake 2.3.0 gets released (in case that's important for the CosmosDB configuration).

ktjn:
> > @HofmeisterAn A suggestion. Just run the test on windows until the actual emulator bug is fixed. What do you think?
> 
> OC, sounds good. I assume the configuration is the same. Just move the test to the Windows namespace. BTW, I'll update to windows2022 and ubuntu-22.04 soon as Cake 2.3.0 gets released (in case that's important for the CosmosDB configuration).

It looks like more of an cpu architecture problem. 22.04 might help. I'll fix the test.

ktjn:
@HofmeisterAn I created a "SkipOnLinuxOS" fact attribute similar to SkipOnLinuxEngineAttribute(neat idea btw) and used that for the cosmos tests. It works now but the build is a bit fragile due to other tests.

ktjn:
@Yeseh anything else you want to add to this?

HofmeisterAn:
> @HofmeisterAn I created a "SkipOnLinuxOS" fact attribute similar to SkipOnLinuxEngineAttribute(neat idea btw) and used that for the cosmos tests. It works now but the build is a bit fragile due to other tests.

Hm, can't you use the `SkipOnLinuxEngine` attribute?

> It works now but the build is a bit fragile due to other tests.

What part? Haven't had any problems in the past (then I can take a look at it).

ktjn:
> > @HofmeisterAn I created a "SkipOnLinuxOS" fact attribute similar to SkipOnLinuxEngineAttribute(neat idea btw) and used that for the cosmos tests. It works now but the build is a bit fragile due to other tests.
> 
> Hm, can't you use the `SkipOnLinuxEngine` attribute?

The test works fine as long as docker host is not running linux(and specific intel arch). The container can still be linux based. 
I though SkipOnLinuxEngine was the acctual docker image?

> > It works now but the build is a bit fragile due to other tests.
> 
> What part? Haven't had any problems in the past (then I can take a look at it).

The azurite test failed once and 'UntilPortIsAvailable' failed once. Just check the actions.

There is one problem. Even though the tests are skipped the fixture is still created. Any suggestion how to fix that?

HofmeisterAn:
> The test works fine as long as docker host is not running linux(and specific intel arch). The container can still be linux based.

I hope I did not misunderstand something, but the idea was to run the tests on the Windows host as long as the issues are not fixed (`ubuntu-20.04` using an incompatible Intel architecture, `ubuntu-22.04` failing with Cake 2.2.0), right?

As far as I know, the Azure DevOps and GitHub Windows build agents can't run Linux images (LCOW is not supported): https://github.com/actions/runner-images/issues/252#issuecomment-610370237.

I think your tests are not executed at all:

1. They get skipped due to `SkipOnLinuxOS` (on Linux)
2. They get skipped due to not being in the `DotNet.Testcontainers.Tests.Unit.Containers.Windows` namespace (on Windows):

https://github.com/testcontainers/testcontainers-dotnet/blob/f7c0fc6bed67d706ad03c2d89ca19767b323ed92/.github/workflows/cicd.yml#L50

Is the Cosmos DB Emulator a multi-arch image?


> I though SkipOnLinuxEngine was the acctual docker image?

I use this attribute to disable the Windows specific tests on Linux. All tests run on Linux expect those with the `SkipOnLinuxEngine` attribute. Since the tests have grown so much, we need to refactor and put them in dedicated test assemblies. This will help a lot in the future. The first step will be the dedicated modules.

> There is one problem. Even though the tests are skipped the fixture is still created. Any suggestion how to fix that?

That's due to the xUnit.net collection.

ktjn:
> I hope I did not misunderstand something, but the idea was to run the tests on the Windows host as long as the issues are not fixed (`ubuntu-20.04` using an incompatible Intel architecture, `ubuntu-22.04` failing with Cake 2.2.0), right?
> 
> As far as I know, the Azure DevOps and GitHub Windows build agents can't run Linux images (LCOW is not supported): [actions/runner-images#252 (comment)](https://github.com/actions/runner-images/issues/252#issuecomment-610370237).

Yes, you are correct. This works fine on my machine, but not on the build agents. No test are never run.

> 
> Is the Cosmos DB Emulator a multi-arch image?

No. This is a linux image AFAIK.

The tests were never run on windows(host/image).

It feels like this solution wasn't good enough. I'll revert the changes.

Let's see what happens when you switch to 22.04 / 2022.

HofmeisterAn:
> Yes, you are correct. This works fine on my machine, but not on the build agents. No test are never run.

Ok, then I understood it wrong. I thought we were running the tests on the hosted Windows agents (multi-arch image). But that is not a problem. We can enable them later. Let's focus on an MVP first.

Can you reduce the changes to the neccsary once? Then I can start with the review.

Thanks for your contribution 🙏.

ktjn:
> > Yes, you are correct. This works fine on my machine, but not on the build agents. No test are never run.
> 
> Ok, then I understood it wrong. I thought we were running the tests on the hosted Windows agents (multi-arch image). But that is not a problem. We can enable them later. Let's focus on an MVP first.
> 
> Can you reduce the changes to the neccsary once? Then I can start with the review.
> 
> Thanks for your contribution 🙏.

Do you think I should disable the tests? I might be able to run get the cpuinfo from inside the container and see if the tests will work. Otherwise the only option is to set the tests to manual/or skip.

I'll clean everything up then.

Yeseh:
@ktjn Man thanks for picking up my slack! Don't have anything to add at the moment. I have time again so can I help you with anything concrete?

ktjn:
> @ktjn Man thanks for picking up my slack! Don't have anything to add at the moment. I have time again so can I help you with anything concrete?

Thanks. I'm trying to find a good middleground without disabling the test. Since there is a random chance of getting a build agent where the emulator will fail to start. I want to detect if the container isn't starting and set the test to inconclusive or something.

So far I can timeout the start but it wont dispose correctly. 

public override async Task InitializeAsync()
    {
      var containerTask = this.Container.StartAsync();
      var task = await Task.WhenAny(new[] { containerTask, Task.Delay(10 * 1000) });
      if (task != containerTask)
      {
        await this.DisposeAsync();
      }
    }

ktjn:
So I added a waittimeout. @HofmeisterAn What do you think about this? Could be a neat feature. Otherwise waits will run forever.

Not really sure what happens when the container timeouts. Let's see :-)

HofmeisterAn:
> Do you think I should disable the tests?

I would suggest disabling them with a comment mentioning the problem. Maybe we can use a build agent from the Testcontainers organization until the issue is solved with the once from Microsoft.

> So I added a waittimeout. @HofmeisterAn What do you think about this? Could be a neat feature. Otherwise waits will run forever.

I do not think we need this. You can cancel the tasks anytime with a cancellation token ([CancellationTokenSource](https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.-ctor?view=net-6.0#system-threading-cancellationtokensource-ctor(system-timespan))).

ktjn:
> I would suggest disabling them with a comment mentioning the problem. Maybe we can use a build agent from the Testcontainers organization until the issue is solved with the once from Microsoft.

I disabled the tests with a comment. It looks like ubuntu18 is the only working version which is pretty obsolete.


> I do not think we need this. You can cancel the tasks anytime with a cancellation token ([CancellationTokenSource]

Totally missed the cancellationtoken. Just saw the timeout which was set to -1 in WaitStrategy.WaitUntil and exposed it. Updated now to use the cancellationtoken. I set it to max 5 min. I don't want to disable the fixture.

Yeseh:
Running into a lot of trouble with other unrelated tests not passing in my environment, so not making progress atm :/. Feel a little bit out of my depth here, so ok if I leave this to you for now @ktjn ?

ktjn:
> Running into a lot of trouble with other unrelated tests not passing in my environment, so not making progress atm :/. Feel a little bit out of my depth here, so ok if I leave this to you for now @ktjn ?

No worries. I can take it from here 😄

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,50 @@
...
+        this.portNumber = portNumber;
+      }
+
+      protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+      {
+        request.RequestUri = new Uri($"https://{this.host}:{this.portNumber}{request.RequestUri?.PathAndQuery}");
+        var response = await base.SendAsync(request, cancellationToken);
+        return response;
+      }
+    }
```
JonasBenz: This looks like a workaround. If we would set the IP of the Docker host as environment variable AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE, then we could get rid of the `UrlRewriter`.

Unfortunately it is not that straightforward to get the IP of the docker host. I tried to set it to `172.17.0.1` as described here: https://stackoverflow.com/questions/48546124/what-is-linux-equivalent-of-host-docker-internal/48547074#48547074. This worked on our Gitlab runner, which uses Socket Bindings like described here: https://docs.gitlab.com/ee/ci/docker/using_docker_build.html, but it failed on my local machine running Docker Desktop 4.11.1 on Windows.

Maybe @HofmeisterAn has an idea how to get the IP of the docker host before the container is started?

ktjn: > This looks like a workaround. If we would set the IP of the Docker host as environment variable AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE, then we could get rid of the `UrlRewriter`.

Yes, I know.  But it wasn't possible to set the override as the ip is unavailable at this point. Since this is only a test support lib I think the workaround is ok.

JonasBenz: The problem is, when your production code which you want to test still uses the [DocumentClient](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.client.documentclient?view=azure-dotnet), then you are not able to set another `HttpClient`. You can only set the `HttpMessageHandler`.

Of course it would be better to change the production code to use [CosmosClient](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.cosmosclient?view=azure-dotnet), but you want some integration tests using testconainers-dotnet to cover that, before you do that change i guess.

So maybe it would be better to at least provide a property of type `HttpMessageHandler`, instead of `HttpClient`? This would also be compatible with  [HttpClientBuilderExtensions.ConfigurePrimaryHttpMessageHandler](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.httpclientbuilderextensions.configureprimaryhttpmessagehandler?view=dotnet-plat-ext-6.0)

Anyway if we could set AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE to the correct IP this would be the best solution in my opinion.

ktjn: > So maybe it would be better to at least provide a property of type `HttpMessageHandler`, instead of `HttpClient`? This would also be compatible with [HttpClientBuilderExtensions.ConfigurePrimaryHttpMessageHandler](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.httpclientbuilderextensions.configureprimaryhttpmessagehandler?view=dotnet-plat-ext-6.0)
> 
> Anyway if we could set AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE to the correct IP this would be the best solution in my opinion.

This would be the best solution I think. @HofmeisterAn any idea how to get hold of the hosts ip? before configuration.

HofmeisterAn: Sorry, I have been on a trip the past days. I think Kafka uses a similar approach. Might be the startup callback useful?

https://github.com/testcontainers/testcontainers-dotnet/blob/997c019428132ebc94dc981b454324e1e7a6f302/src/Testcontainers/Configurations/Modules/MessageBrokers/KafkaTestcontainerConfiguration.cs#L78

ktjn: > Sorry, I have been on a trip the past days. I think Kafka uses a similar approach. Might be the startup callback useful?
> 
 
Yes. This looks promising. I'll try something similar.

ktjn:
Ok, so I tried setting the ip of the host. Took me a while to figure out that the values wren't actually used(fix this now at least). Still the get accountinfo works but the database operation times out.

Any idea @JonasBenz?

ktjn:
Did some more investigation. The url rewriter is by far the most efficeient way to get the cosmos emulator running in docker.

JonasBenz:
> Did some more investigation. The url rewriter is by far the most efficeient way to get the cosmos emulator running in docker.

So setting the IP of the docker host as environment variable AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE like @HofmeisterAn suggested here did not work?: https://github.com/testcontainers/testcontainers-dotnet/pull/549#discussion_r979285675

If the `UrlRewriter` is the better solution, I would prefer to provide a property of type `HttpMessageHandler`, instead of `HttpClient` because of the reasons described here: https://github.com/testcontainers/testcontainers-dotnet/pull/549#discussion_r974226170

Or providing both `HttpMessageHandler` and `HttpClient` would also be an option.

ktjn:
> Or providing both `HttpMessageHandler` and `HttpClient` would also be an option.

Done.

ktjn:
@HofmeisterAn I guess this PR is ready for review.

HofmeisterAn:
> @HofmeisterAn I guess this PR is ready for review.

👏 I will do it in the next days. I hope it is not urgent.

src/Testcontainers/Builders/TestcontainersBuilderCosmosDbExtension.cs
```diff
@@ -0,0 +1,29 @@
+namespace DotNet.Testcontainers.Builders
+{
+  using System.Linq;
+  using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Containers;
+  using JetBrains.Annotations;
+
+  [PublicAPI]
+  public static class TestcontainersBuilderCosmosDbExtension
```
HofmeisterAn: All other classes inherit from `TestcontainerDatabase` (or equivalent), why do we not use `TestcontainersBuilderDatabaseExtension`?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,66 @@
...
+  {
+    public const string DefaultCosmosDbApiImage =
+      "mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator";
+
+    private const int DefaultCosmosPort = 8081;
+
+    public CosmosDbTestcontainerConfiguration()
+      : this(DefaultCosmosDbApiImage)
+    {
+      this.OutputConsumer = Consume.RedirectStdoutAndStderrToStream(new MemoryStream(), new MemoryStream());
```
HofmeisterAn: Move the `OutputConsumer` initialization to the constructor below. Otherwise, it is not always initialized. I think we can make the class `sealed` and use following to initialize the properties:

```csharp
this.PartitionCount = 1;
this.IpAddressOverride = "127.0.0.1";
this.OutputConsumer = Consume.RedirectStdoutAndStderrToStream(new MemoryStream(), new MemoryStream());
this.WaitStrategy = Wait.ForUnixContainer().UntilMessageIsLogged(this.OutputConsumer.Stdout, "Started|Shutting");
```

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    public override string Database
+    {
+      get => this.Environments["AZURE_COSMOS_EMULATOR_DATABASE"];
+      set => this.Environments["AZURE_COSMOS_EMULATOR_DATABASE"] = value;
+    }
+
+    public int PartitionCount
+    {
+      get => int.Parse(this.Environments["AZURE_COSMOS_EMULATOR_PARTITION_COUNT"], CultureInfo.InvariantCulture);
```
HofmeisterAn: `int.TryParse(this.Environments["AZURE_COSMOS_EMULATOR_PARTITION_COUNT"], NumberStyles.Integer, CultureInfo.InvariantCulture, out var partitionCount) ? partitionCount : 1;`?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,66 @@
...
+      set => this.Environments["AZURE_COSMOS_EMULATOR_PARTITION_COUNT"] = value.ToString(CultureInfo.InvariantCulture);
+    }
+
+    public bool EnableDataPersistence
+    {
+      get => this.Environments["AZURE_COSMOS_EMULATOR_ENABLE_DATA_PERSISTENCE"] == "true";
```
HofmeisterAn: `bool.TryParse(this.Environments["AZURE_COSMOS_EMULATOR_ENABLE_DATA_PERSISTENCE"], out var enableDataPersistence) && enableDataPersistence;`?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,66 @@
...
+      set => this.Environments["AZURE_COSMOS_EMULATOR_ENABLE_DATA_PERSISTENCE"] = value.ToString().ToLower(CultureInfo.InvariantCulture);
```
HofmeisterAn: `ToLowerInvariant()`

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,53 @@
...
+  using Microsoft.Extensions.Logging;
+
+  public sealed class CosmosDbTestcontainer : TestcontainerDatabase
+  {
+    internal CosmosDbTestcontainer(ITestcontainersConfiguration configuration, ILogger logger)
+      : base(configuration, logger)
+    {
+    }
+
+    public HttpMessageHandler HttpMessageHandler => new UrlRewriter(this.Hostname, this.Port, new HttpClientHandler { ServerCertificateCustomValidationCallback = (message, cert, chain, errors) => true, });
```
HofmeisterAn: We do not need this. See the suggestion below.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public HttpClient HttpClient
+    {
+      get
+      {
+        return new HttpClient(this.HttpMessageHandler);
```
HofmeisterAn: `return new HttpClient(new UriRewriter(this.Hostname, this.Port));`

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,53 @@
...
+      }
+    }
+
+    public override string ConnectionString =>
+      $"AccountEndpoint=https://{this.Hostname}:{this.Port};AccountKey={this.Password}";
+
+    private class UrlRewriter : DelegatingHandler
```
HofmeisterAn:
```csharp
private sealed class UriRewriter : DelegatingHandler
{
  private readonly string hostname;

  private readonly int port;

  public UriRewriter(string hostname, int port)
    : base(new HttpClientHandler { ServerCertificateCustomValidationCallback = (message, cert, chain, errors) => true })
  {
    this.hostname = hostname;
    this.port = port;
  }

  protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
  {
    request.RequestUri = new UriBuilder("https", this.hostname, this.port, request.RequestUri.PathAndQuery).Uri;
    return base.SendAsync(request, cancellationToken);
  }
}
```

tests/Testcontainers.Tests/Fixtures/Containers/Unix/Modules/Databases/CosmosDbFixture.cs
```diff
@@ -0,0 +1,58 @@
...
+        .WithCosmosDb(configuration)
+        .Build();
+    }
+
+    public CosmosDbTestcontainerConfiguration Configuration { get; set; }
+
+    public override async Task InitializeAsync()
+    {
+      // workaround for broken cosmosdb emulator
+      var maxWait = TimeSpan.FromSeconds(5 * 1000);
```
HofmeisterAn: You do not need all that. `CancellationTokenSource` does that already:

```csharp
[UsedImplicitly]
public sealed class CosmosDbFixture : DatabaseFixture<CosmosDbTestcontainer, CosmosClient>
{
  private readonly CosmosDbTestcontainerConfiguration configuration = new CosmosDbTestcontainerConfiguration();

  private readonly CancellationTokenSource cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));

  public CosmosDbFixture()
  {
    this.Container = new TestcontainersBuilder<CosmosDbTestcontainer>()
      .WithCosmosDb(this.configuration)
      .Build();
  }

  public override Task InitializeAsync()
  {
    return this.Container.StartAsync(this.cts.Token);
  }

  public override Task DisposeAsync()
  {
    return this.Container.DisposeAsync().AsTask();
  }

  public override void Dispose()
  {
    this.configuration.Dispose();
  }
}
```

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CosmosDbTestcontainerTest.cs
```diff
@@ -0,0 +1,42 @@
...
+    public sealed class ConnectionTests : IClassFixture<CosmosDbFixture>
+    {
+      private readonly CosmosDbFixture fixture;
+
+      public ConnectionTests(CosmosDbFixture fixture)
+      {
+        this.fixture = fixture;
+      }
+
+      private CosmosClientOptions Options => new CosmosClientOptions { HttpClientFactory = () => this.fixture.Container.HttpClient, ConnectionMode = ConnectionMode.Gateway, };
```
HofmeisterAn: In general, I would prefer to create the client in the fixture, but that is a bit cumbersome due to the HTTP client. What about this:

```csharp
[Collection(nameof(Testcontainers))]
public sealed class ConnectionTests : IClassFixture<CosmosDbFixture>, IDisposable
{
  private readonly HttpClient httpClient;

  private readonly CosmosClient cosmosClient;

  public ConnectionTests(CosmosDbFixture cosmosDbFixture)
    : this(cosmosDbFixture.Container.HttpClient, cosmosDbFixture.Container.ConnectionString)
  {
  }

  private ConnectionTests(HttpClient httpClient, string connectionString)
  {
    var cosmosClientOptions = new CosmosClientOptions();
    cosmosClientOptions.ConnectionMode = ConnectionMode.Gateway;
    cosmosClientOptions.HttpClientFactory = () => httpClient;
    this.httpClient = httpClient;
    this.cosmosClient = new CosmosClient(connectionString, cosmosClientOptions);
  }

  [Fact]
  public async Task ShouldEstablishConnection()
  {
    var accountProperties = await this.cosmosClient.ReadAccountAsync()
      .ConfigureAwait(false);

    Assert.Equal("localhost", accountProperties.Id);
  }

  [Fact]
  public async Task CreateDatabaseTest()
  {
    var databaseResponse = await this.cosmosClient.CreateDatabaseIfNotExistsAsync("db")
      .ConfigureAwait(false);

    Assert.Equal(HttpStatusCode.Created, databaseResponse.StatusCode);
  }

  public void Dispose()
  {
    this.cosmosClient.Dispose();
    this.httpClient.Dispose();
  }
}
```

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,53 @@
...
+  using System.Linq;
+  using System.Net;
+  using System.Net.Http;
+  using System.Net.Sockets;
+  using System.Threading;
+  using System.Threading.Tasks;
+  using DotNet.Testcontainers.Configurations;
+  using Microsoft.Extensions.Logging;
+
+  public sealed class CosmosDbTestcontainer : TestcontainerDatabase
```
HofmeisterAn, line 12: `[PublicAPI]`

HofmeisterAn (changes requested):

src/Testcontainers/Builders/TestcontainersBuilderCosmosDbExtension.cs
```diff
@@ -0,0 +1,29 @@
...
+  {
+    public static ITestcontainersBuilder<CosmosDbTestcontainer> WithCosmosDb(
+      this ITestcontainersBuilder<CosmosDbTestcontainer> builder, CosmosDbTestcontainerConfiguration configuration)
+    {
+      builder = configuration.Environments.Aggregate(builder, (current, environment)
+        => current.WithEnvironment(environment.Key, environment.Value));
+
+      return builder.WithImage(configuration.Image)
+        .WithPortBinding(configuration.DefaultPort, true)
```
HofmeisterAn: If you would like to keep the extension, change this to: `WithPortBinding(configuration.Port, configuration.DefaultPort)`.

src/Testcontainers/Builders/TestcontainersBuilderCosmosDbExtension.cs
ktjn: Yes, I agree. Had to set the database to something. Maybe this should be a constructor parameter?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
ktjn: Good point. WaitStrategy is not settable so I kept the override.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
ktjn: Fixed

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
ktjn: Fixed

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
ktjn: Much better. Fixed.

tests/Testcontainers.Tests/Fixtures/Containers/Unix/Modules/Databases/CosmosDbFixture.cs
ktjn: Didn't know that. You always learn something new :-)

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CosmosDbTestcontainerTest.cs
ktjn: Yes, looks good.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
ktjn: Fixed

src/Testcontainers/Builders/TestcontainersBuilderCosmosDbExtension.cs
ktjn: Removed this file.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
HofmeisterAn: Yes - you need to override `WaitStrategy`, but you can still set the value in the constructor (read-only; just to create the instance once). Equivalent to `OutputConsumer`.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+  using System.Globalization;
+  using System.IO;
+  using DotNet.Testcontainers.Builders;
+  using JetBrains.Annotations;
+
+  [PublicAPI]
+  public sealed class CosmosDbTestcontainerConfiguration : TestcontainerDatabaseConfiguration
+  {
+    public const string DefaultCosmosDbApiImage =
+      "mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator";
```
HofmeisterAn, line 13: Can we pin the version?

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
ktjn: done

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
```diff
@@ -0,0 +1,52 @@
...
+
+  [PublicAPI]
+  public sealed class CosmosDbTestcontainer : TestcontainerDatabase
+  {
+    internal CosmosDbTestcontainer(ITestcontainersConfiguration configuration, ILogger logger)
+      : base(configuration, logger)
+    {
+    }
+
+    public HttpMessageHandler HttpMessageHandler => new UriRewriter(this.Hostname, this.Port);
```
HofmeisterAn, line 19: Do we need this property?

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CosmosDbTestcontainerTest.cs
```diff
@@ -0,0 +1,57 @@
...
+      }
+
+      private ConnectionTests(HttpClient httpClient, string connectionString)
+      {
+        var cosmosClientOptions = new CosmosClientOptions { ConnectionMode = ConnectionMode.Gateway, HttpClientFactory = () => httpClient };
+        this.httpClient = httpClient;
+        this.cosmosClient = new CosmosClient(connectionString, cosmosClientOptions);
+      }
+
+      [Fact(Skip = "Waiting for a working cosmosdb emulator")]
```
HofmeisterAn: Can we add the GitHub issue, why we skip those tests?

HofmeisterAn (changes requested):
Looks good. I have just a few more small suggestions / improvements.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
ktjn: This was requested by @JonasBenz so, I guess it has to do with .NET FW and cosmosclient.

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/CosmosDbTestcontainerTest.cs
ktjn: Good idea, done.

src/Testcontainers/Configurations/Modules/Databases/CosmosDbTestcontainerConfiguration.cs
ktjn: I think that only "lastest" exists.

ktjn: https://mcr.microsoft.com/v2/cosmosdb/linux/azure-cosmos-emulator/tags/list

"2.11.11-5e1f77c6",
"2.11.11-787c2a71",
"2.11.12-efd4ee75",
"2.11.13-1ef6349a",
"2.11.13-36cf4abc",
"2.11.13-a2eeffac",
"2.11.13-c5c131e3",
"2.11.13-sas",
"2.14.5.1",
"2.14.5.2",
"2.14.5.internal",
"latest",
"mongodb",
"mongodb-debug"

"lastest" should be the safest choice.

HofmeisterAn: Ok, thanks for the info.

src/Testcontainers/Containers/Modules/Databases/CosmosDbTestcontainer.cs
HofmeisterAn: Ok, but then I would use the property at `HttpClient` too. With the dedicated modules we can move it into a separate class in the future.

ktjn: Done

HofmeisterAn (approved):

HofmeisterAn:
07c6786946f57233f027e938f64285ce9c0fded0, does it close #421? Or is anything major left? Does it make sense to create a new issue?

ktjn:
Maybe for fixing the tests when the emulator is stable.

HofmeisterAn:
> Maybe for fixing the tests when the emulator is stable.

Then I will create a separate issue. Other than that the Cosmos DB Emulator is done?

ktjn:
> 
> Then I will create a separate issue. Other than that the Cosmos DB Emulator is done?

Yes!

HofmeisterAn:
07c6786946f57233f027e938f64285ce9c0fded0, closes #421.

PR #552, opened 2022-08-05
feat: Add http wait strategy

sumihiran:
Adds external http wait strategy as an extension of `IWaitForContainerOS`.  In doing so, http wait strategy can be chained with OS specific ones. I am opening this PR for feedback.

**Todo**

- [X] Wait for status code
- [ ] TLS 
- [ ] Request headers
- [ ] Basic authentication
- [ ] Response predicate

**Example**

```csharp
var waitRequestBuilder = HttpWaitRequest
    .ForPort(8024)
    .ForPath("/actuator/health")
    .ForStatusCode(200)
    .WithReadTimeout(TimeSpan.FromSeconds(5));

Wait.ForUnixContainer()
  .UntilPortIsAvailable(8024)
  .UntilHttpRequestIsCompleted(waitRequestBuilder.build())
```

### DIFF
--- a/src/Testcontainers/Builders/HttpWaitRequest.cs
+++ b/src/Testcontainers/Builders/HttpWaitRequest.cs
@@ -1,0 +1,137 @@
+namespace DotNet.Testcontainers.Builders
+{
+  using System;
+  using System.Collections.Generic;
+  using System.Diagnostics.CodeAnalysis;
+  using System.Net.Http;
+  using JetBrains.Annotations;
+
+  /// <summary>
+  /// A http wait request.
+  /// </summary>
+  [PublicAPI]
+  public class HttpWaitRequest
+  {
+    /// <summary>
+    /// Initializes a new instance of the <see cref="HttpWaitRequest" /> class.
+    /// </summary>
+    /// <param name="port">The private container port.</param>
+    /// <param name="statusCodes">The expected status codes.</param>
+    /// <param name="path">The absolute path of the request uri.</param>
+    /// <param name="method">The http method.</param>
+    /// <param name="readTimeout">the http connection read timeout.</param>
+    private HttpWaitRequest(int port, ISet<int> statusCodes, string path, HttpMethod method, TimeSpan readTimeout)
+    {
+      this.Port = port;
+      this.StatusCodes = statusCodes;
+      this.Path = path;
+      this.Method = method;
+      this.ReadTimeout = readTimeout;
+    }
+
+    public int Port { get; }
+
+    public string Path { get; }
+
+    public HttpMethod Method { get; }
+
+    public ISet<int> StatusCodes { get; }
+
+    public TimeSpan ReadTimeout { get; }
+
+    /// <summary>
+    /// Returns a new instance of wait request builder for a given port
+    /// </summary>
+    /// <param name="port">The private container port.</param>
+    /// <returns>Instance of <see cref="HttpWaitRequest.Builder" />.</returns>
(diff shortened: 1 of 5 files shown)


src/Testcontainers/Configurations/WaitStrategies/ExternalWaitStrategyExtension.cs
```diff
@@ -0,0 +1,24 @@
+namespace DotNet.Testcontainers.Configurations
+{
+  using DotNet.Testcontainers.Builders;
+  using JetBrains.Annotations;
+
+  public static class ExternalWaitStrategyExtension
```
HofmeisterAn, line 6: Any reason why you choose an extension? I think we can put this into `WaitForContainerOS`.

src/Testcontainers/Configurations/WaitStrategies/UntilHttpRequestIsCompleted.cs
```diff
@@ -0,0 +1,94 @@
...
+
+      await WaitStrategy.WaitUntil(
+        async () =>
+        {
+          try
+          {
+            var response = await httpClient.SendAsync(
+              new HttpRequestMessage(
+                this.request.Method,
+                this.BuildRequestUri(testcontainers.Hostname, testcontainers.GetMappedPublicPort(this.request.Port))));
```
HofmeisterAn, line 47: I think you can use the [UriBuilder](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder?view=net-6.0) here. If you create `HttpRequestMessage` in the `HttpWaitRequest` you probably don't need it here at all.

HofmeisterAn, line 47: Maybe you can pass the `Hostname` to something like `HttpRequestMessage Build(string hostname, int mappedPublicPort)`. Or `WithHostname(...)`. That the builder creates the `HttpRequestMessage`. Then you probably don't need the nested `Builder` class too.

src/Testcontainers/Builders/HttpWaitRequest.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// <summary>
+    /// A fluent wait request builder.
+    /// </summary>
+    [SuppressMessage("ReSharper", "ParameterHidesMember", Justification = "Fluent builder")]
+    [PublicAPI]
+    public class Builder
+    {
+      private const string DefaultPath = "/";
+
+      private readonly int port;
```
HofmeisterAn, line 62: I think you can use the [UriBuilder](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder?view=net-6.0) here and store the values in the builder fields.

src/Testcontainers/Builders/HttpWaitRequest.cs
```diff
@@ -0,0 +1,137 @@
...
+      this.ReadTimeout = readTimeout;
+    }
+
+    public int Port { get; }
+
+    public string Path { get; }
+
+    public HttpMethod Method { get; }
+
+    public ISet<int> StatusCodes { get; }
```
HofmeisterAn, line 38: `ISet<HttpStatusCode>`?

HofmeisterAn (changes requested):
Looks pretty good, thanks! Just a few minor comments.

HofmeisterAn:
I will close this in favor of #717. I've added you as a co-author. Thanks for your contribution.

PR #555, opened 2022-08-07
feat(#421) Added AzuriteTestcontainer

vlaskal:
Hi @HofmeisterAn ,

I've created this draft PR with `AzuriteTestcontainer`. Currently there is only basic support of options.
What is done:
- Default image
- Custom image
- Host Blob port bind
- Host Queue port bind
- Host Table port bind
- Connection string
- Wait strategy
- Default output consumer
- onlyBlob
- onlyQueue
- onlyTable
- azurite.blobHost
- azurite.queueHost
- azurite.tableHost
- azurite.location
- azurite.silent
- azurite.debug
- azurite.loose
- azurite.skipApiVersionCheck
- azurite.disableProductStyleUrl
- azurite.containerBlobPort
- azurite.containerQueuePort
- azurite.containerTablePort

What is not done:
- azurite.containerLocation
- azurite.cert
- azurite.key
- azurite.pwd
- azurite.oauth
- Account name
- Account Key

I need to also discuss how to fit it into current approach, which is inheritance from `TestcontainerDatabaseConfiguration`, `TestcontainerDatabase` and `DatabaseFixture<TDockerContainer, TDatabaseConnection>`. The main problem what I have is that I use just few prepared fields or properties and most of them are useless for Azurite. I look at eg. `KafkaTestcontainer` implementation and I have one suggestion:
I think class `HostedServiceContainer` should be more generic. It means I would like to remove `Username` and `Password` proeprty and moved them to specific classes. In this case to `TestcontainerDatabase` also if it make sense to `TestcontainerMessageBroker` and leave only `Port` setup there. Then I can use eg. `BlobPort` for this default `Port` implementation.
Same think applies to configuration classes `HostedServiceConfiguration`, `TestcontainerDatabaseConfiguration` and `TestcontainerMessageBrokerConfiguration` .
Then I can inherit `AzuriteTestcontainer` directly from `HostedServiceTestcontainer`.

What do you think?


vlaskal:
Main options, switches and settings are implemented. I think more than 90% of use cases should be supported. I would wait if someone will request them.
I will move this PR to ready state.

Packages.props
```diff
@@ -4,6 +4,9 @@
     <GlobalPackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0" />
     <GlobalPackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
     <GlobalPackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" />
+    <PackageReference Update="Azure.Storage.Blobs" Version="12.13.0" />
```
HofmeisterAn: I would put them to the `3rd party client dependencies` block.

Testcontainers.sln.DotSettings
```diff
@@ -0,0 +1,6 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
HofmeisterAn: Please only add changes that belongs to the pull request / feature.

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
```diff
@@ -0,0 +1,139 @@
...
+      {
+        builder = builder.WithPortBinding(configuration.TablePort, AzuriteTestcontainerConfiguration.DefaultTablePort);
+      }
+
+      if (!string.IsNullOrWhiteSpace(configuration.Location))
+      {
+        builder = builder.WithBindMount(configuration.Location, AzuriteTestcontainerConfiguration.DefaultLocation);
+      }
+
+      builder = builder
```
HofmeisterAn: Immediately return.

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
```diff
@@ -0,0 +1,139 @@
...
+        .WithCommand(GetDisableProductStyleUrl(configuration))
+        .WithCommand(GetDebugLog(configuration));
+
+      return builder;
+    }
+
+    private static string[] GetDebugLog(AzuriteTestcontainerConfiguration configuration)
+    {
+      var debugLogFilePath = Path.Combine(AzuriteTestcontainerConfiguration.DefaultLocation, "debug.log");
+      return configuration.DisableProductStyleUrl ? new[] { "--debug", debugLogFilePath } : null;
```
HofmeisterAn: Shouldn't this be `configuration.Debug`?.

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
```diff
@@ -0,0 +1,139 @@
...
+        args.Add(AzuriteTestcontainerConfiguration.DefaultQueueEndpoint);
+      }
+
+      if (configuration.RunTableOnly || configuration.RunAllServices)
+      {
+        args.Add("--tableHost");
+        args.Add(AzuriteTestcontainerConfiguration.DefaultTableEndpoint);
+      }
+
+      return args.Count > 0 ? args.ToArray() : null;
```
HofmeisterAn: I think you can just return `args.ToArray()`.

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,279 @@
+namespace DotNet.Testcontainers.Configurations.Modules.Databases
```
HofmeisterAn: It's `DotNet.Testcontainers.Configurations`

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,279 @@
...
+{
+  using System;
+  using System.Collections.Generic;
+  using DotNet.Testcontainers.Builders;
+  using JetBrains.Annotations;
+
+  public class AzuriteTestcontainerConfiguration : IDisposable
```
HofmeisterAn: `[PublicAPI]`

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,279 @@
...
+    public bool RunAllServices => !this.RunBlobOnly && !this.RunQueueOnly && !this.RunTableOnly;
+
+    /// <summary>
+    /// Gets or sets workspace location path.
+    /// </summary>
+    /// <remarks>
+    /// Corresponds to the default workspace location of the hosted service.
+    /// </remarks>
+    [PublicAPI]
+    public string Location
```
HofmeisterAn: `[NotNull]`

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,279 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AzuriteTestcontainerConfiguration" /> class.
+    /// </summary>
+    /// <param name="image">The Docker image.</param>
+    protected AzuriteTestcontainerConfiguration(string image)
+    {
+      this.Image = image;
+      this.Environments = new Dictionary<string, string>();
+      this.OutputConsumer = Consume.DoNotConsumeStdoutAndStderr();
```
HofmeisterAn: You don't need (and use) that, that's the default value. Developers can still set the output consumer with their builder configuration.

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AzuriteTestcontainerConfiguration" /> class.
+    /// </summary>
+    /// <param name="image">The Docker image.</param>
+    protected AzuriteTestcontainerConfiguration(string image)
+    {
+      this.Image = image;
+      this.Environments = new Dictionary<string, string>();
```
HofmeisterAn: You don't need (and use) that. Developers can still set the environment variables with their builder configuration.

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,279 @@
...
+    public AzuriteTestcontainerConfiguration()
+      : this(DefaultAzuriteImage)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AzuriteTestcontainerConfiguration" /> class.
+    /// </summary>
+    /// <param name="image">The Docker image.</param>
+    protected AzuriteTestcontainerConfiguration(string image)
```
HofmeisterAn: `public`

HofmeisterAn (changes requested):
I did the review of the extension and configuration class. Beside of some minor suggestions, I have two general one:

- I'd suggest to use a more common naming pattern (in. NET) for the boolean properties. Instead of `Debug`, `DisableProductStyleUrl`, `SkipApiVersionCheck` or `RunTableOnly`, I would use something like `DebugEnabled`, `TableServiceEnabled`, `IsSilent`, etc.
- The `WithPortBinding` configuration looks static. Please use random assigned host ports (`GetMappedPublicPort`). We cannot rely on static port configurations. Ports may be already in use.

vlaskal:
> * I'd suggest to use a more common naming pattern (in. NET) for the boolean properties. Instead of `Debug`, `DisableProductStyleUrl`, `SkipApiVersionCheck` or `RunTableOnly`, I would use something like `DebugEnabled`, `TableServiceEnabled`, `IsSilent`, etc.

Original intention was to make it same as are command options and documentation. But you are right. I updated names as you suggested.

> * The `WithPortBinding` configuration looks static. Please use random assigned host ports (`GetMappedPublicPort`). We cannot rely on static port configurations. Ports may be already in use.

PortBindings is not static. It is same implementation as is used in other containers. Only there is not option to set port inside of container. HostPort value is by default 0 which means it will use random port and will be bind to static ContainerPort. Do you think ContainerPort should be also configurable? I tried to make them configurable as is done in other contianers.

Packages.props
vlaskal: Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

Testcontainers.sln.DotSettings
vlaskal: Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
vlaskal: Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
vlaskal: Good catch. Should be each option tested?
Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
vlaskal: Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Sorry, autogenerated vaule.
Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: `[NotNull]` means that returned value cannot be null which is not true in this case because default value is null. I change implementation of this and add `[CanBeNull]` in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Removed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Removed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Fixed in [bddb8be](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/bddb8be4dc91b51a89a704799bbe006d784e4f8c)

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
HofmeisterAn: Not necessarily, but if we can cover more or all it would be great. Usually, I try to cover all.

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
HofmeisterAn: I think my idea was to use `DefaultLocation` instead. I try to avoid `null` as much as I can, but I don't remember anymore.

HofmeisterAn:
> PortBindings is not static. It is same implementation as is used in other containers. Only there is not option to set port inside of container. HostPort value is by default 0 which means it will use random port and will be bind to static ContainerPort.

Ok, then maybe I understood something wrong. I'll take a look at it again (incl. the other classes). Thanks already!

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,233 @@
...
+    public int BlobContainerPort { get; set; } = DefaultBlobPort;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether Blob service should run standalone.
+    /// </summary>
+    /// <remarks>
+    /// Default value is false.
+    /// </remarks>
+    [PublicAPI]
+    public bool BlobServiceOnlyEnabled
```
HofmeisterAn, line 76: I think an enum could improve this a bit. What do you think about something like:

```csharp
public sealed class AzuriteTestcontainerConfiguration
{
  private AzuriteServices enabledServices = AzuriteServices.Blob | AzuriteServices.Queue | AzuriteServices.Table;

  public bool AllServicesEnabled
  {
    get
    {
      return this.enabledServices.HasFlag(AzuriteServices.Blob | AzuriteServices.Queue | AzuriteServices.Table);
    }
  }

  public bool BlobServiceEnabled
  {
    get
    {
      // Either (this requires some changes in GetMainCommand):
      return this.enabledServices.HasFlag(AzuriteServices.Blob);

      // or (see comment in setter):
      return AzuriteServices.Blob.Equals(this.enabledServices)
    }

    set
    {
      // Either, this allows us to enable or disable certain combinations of services:
      this.enabledServices &= value ? AzuriteServices.Blob : ~AzuriteServices.Blob;

      // or set a service exclusive (that's probably what you want, either one or all):
      this.enabledServices = value ? AzuriteServices.Blob : AzuriteServices.Blob | AzuriteServices.Queue | AzuriteServices.Table;
    }
  }

  [Flags]
  private enum AzuriteServices
  {
    /// <summary>
    ///
    /// </summary>
    Blob = 0x01,

    /// <summary>
    ///
    /// </summary>
    Queue = 0x02,

    /// <summary>
    ///
    /// </summary>
    Table = 0x04,
  }
}
```

This make checks like `configuration.BlobServiceOnlyEnabled || configuration.AllServicesEnabled` obsolete (you don't need to check `AllServicesEnabled` anymore).

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,233 @@
...
+    // azurite.oauth OAuth authentication level. Candidate level values: basic.
+    // AZURITE_ACCOUNTS environment variable to set account names and keys.
+
+    /// <summary>
+    /// Default Workspace location folder path. Default is /data.
+    /// </summary>
+    [PublicAPI]
+    public const string DefaultLocation = "/data";
+
+    internal const string DefaultBlobEndpoint = "0.0.0.0";
```
HofmeisterAn: I don't think we need those endpoints in the configuration class. What do you think about moving them to `GetServiceEndpointArgs`? Something like:

```csharp
private static string[] GetServiceEndpointArgs(AzuriteTestcontainerConfiguration configuration)
{
  const string allowRemoteRequests = "0.0.0.0";

  ...

  if (configuration.BlobServiceOnlyEnabled || configuration.AllServicesEnabled)
  {
    args.Add("--blobHost");
    args.Add(allowRemoteRequests);
    args.Add("--blobPort");
    args.Add(configuration.BlobContainerPort.ToString());
  }

  ...
}
```

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
```diff
@@ -0,0 +1,146 @@
...
+    private static string[] GetServiceEndpointArgs(AzuriteTestcontainerConfiguration configuration)
+    {
+      var args = new List<string>();
+
+      if (configuration.BlobServiceOnlyEnabled || configuration.AllServicesEnabled)
+      {
+        args.Add("--blobHost");
+        args.Add(AzuriteTestcontainerConfiguration.DefaultBlobEndpoint);
+        args.Add("--blobPort");
+        args.Add(configuration.BlobContainerPort.ToString());
```
HofmeisterAn: `.ToString(CultureInfo.InvariantCulture)`, respectively for the others too.

src/Testcontainers/Containers/Modules/Databases/AzuriteTestcontainer.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </summary>
+    [PublicAPI]
+    public int ContainerTablePort { get; set; }
+
+    /// <summary>
+    /// Gets the Storage connection string.
+    /// </summary>
+    [PublicAPI]
+    public string ConnectionString =>
+      $"DefaultEndpointsProtocol=http;" +
```
HofmeisterAn: What about a `StringBuilder` or `string.Join(";", ...)`, similar to:
https://github.com/testcontainers/testcontainers-dotnet/blob/fff8c08385f0617c79963d569a7e5bb244133b0d/src/Testcontainers/Configurations/Modules/MessageBrokers/KafkaTestcontainerConfiguration.cs#L63

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
```diff
@@ -0,0 +1,233 @@
+namespace DotNet.Testcontainers.Configurations
+{
+  using DotNet.Testcontainers.Builders;
+  using JetBrains.Annotations;
+
+  [PublicAPI]
+  public class AzuriteTestcontainerConfiguration
+  {
+    // TODO: Add support for these options based on request
```
HofmeisterAn: Can we track open todos in an issue? We can add them as further tasks to #421.

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/AzuriteTestcontainerTest.cs
```diff
@@ -0,0 +1,224 @@
...
+      Assert.Contains(files, file => file.EndsWith(QueueServiceDataFileName, StringComparison.InvariantCultureIgnoreCase));
+      Assert.Contains(files, file => file.EndsWith(TableServiceDataFileName, StringComparison.InvariantCultureIgnoreCase));
+    }
+
+    private static string GetDataFilename(string service)
+    {
+      return $"__azurite_db_{service}__.json";
+    }
+
+    private async Task StartAzuriteContainer(Action<AzuriteTestcontainerConfiguration> configure = null)
```
HofmeisterAn: Can we use dedicated fixtures?

```csharp
public static class AzuriteFixture
{
  private static readonly string TempDir = Environment.GetEnvironmentVariable("AGENT_TEMPDIRECTORY") ?? Directory.GetCurrentDirectory();

  public sealed class AzuriteBlobOnlyFixture : IAsyncLifetime, IDisposable
  {
    private readonly DirectoryInfo dataDirectoryPath = new DirectoryInfo(Path.Combine(TempDir, Guid.NewGuid().ToString("N")));

    private readonly AzuriteTestcontainer container = new TestcontainersBuilder<AzuriteTestcontainer>()
      .WithAzurite(new AzuriteTestcontainerConfiguration())
      .Build();

    public Task InitializeAsync()
    {
      this.dataDirectoryPath.Create();
      return this.container.StartAsync();
    }

    public Task DisposeAsync()
    {
      return this.container.DisposeAsync().AsTask();
    }

    public void Dispose()
    {
      this.dataDirectoryPath.Delete(true);
    }
  }

  ...
}
```

Then pass it via `IClassFixture<T>`to the test.

src/Testcontainers/Containers/Modules/Databases/AzuriteTestcontainer.cs
```diff
@@ -0,0 +1,63 @@
...
+      $"AccountName=devstoreaccount1;" +
+      $"AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;" +
+      (ContainerBlobPort != 0 ? $"BlobEndpoint=http://{this.Hostname}:{this.BlobPort}/devstoreaccount1;" : string.Empty) +
```
HofmeisterAn: Fix Sonar finding (use `this`).

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
```diff
@@ -0,0 +1,146 @@
...
+      }
+
+      if (configuration.TableServiceOnlyEnabled || configuration.AllServicesEnabled)
+      {
+        builder = builder
+          .WithExposedPort(configuration.TableContainerPort)
+          .WithPortBinding(configuration.TablePort, configuration.TableContainerPort);
+      }
+
+      if (!string.IsNullOrWhiteSpace(configuration.Location))
```
HofmeisterAn: Maybe `Directory.Exists(configuration.Location)`?

HofmeisterAn (commented):
Looks good! I have some small suggestions or improvements. They are not necessary, but will still improve it. What do you think about them?

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: I am not sure. I created properties with "ONLY" word because azurite runs with all services or just only one based on documentation https://github.com/Azure/Azurite#run-azurite-v3-docker-image. I am not sure if there is possibility to run any service combination therefore I had strict implementation. 
I will try to look at code and documenation of Azurite if container can run any combination of services. Based on result will try to update.

HofmeisterAn: Yep, that's why I added both options. If only one or all services are a valid configuration the 2nd option convers it.

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Sure, we can. I think will remove these todos because I think we can wait for request. It is not used so often I guess.

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/AzuriteTestcontainerTest.cs
vlaskal: I had fixture, I think it is visible from history. but each use different container with different configuration. so How to do it? Do we want to use 4 fixtures?

HofmeisterAn: > Do we want to use 4 fixtures

Yes. I think nested classes are fine here.

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Make sense

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/AzuriteTestcontainerTest.cs
vlaskal: Is there any sample to see what you exactly mean?

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
vlaskal: I am not sure if I follow you. This condition checks if user wants bind custom location. In case he provides incorrect folder path then I expeect that it will fail later somewhere. But I should set that folder because user expects it. Should I validate path before I will set bind?

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)

src/Testcontainers/Containers/Modules/Databases/AzuriteTestcontainer.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)

src/Testcontainers/Configurations/Modules/Databases/AzuriteTestcontainerConfiguration.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/AzuriteTestcontainerTest.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)

src/Testcontainers/Containers/Modules/Databases/AzuriteTestcontainer.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)

src/Testcontainers/Builders/TestcontainersBuilderAzuriteExtension.cs
vlaskal: Fixed in [31aaf9c](https://github.com/testcontainers/testcontainers-dotnet/pull/555/commits/31aaf9ca336fbf76c035468a3e64378c9087ccbb)
Please check if I understand correctly to your comment.

kaylumah:
Very nice that you are working on this :D 

Came looking because I did not get it working with Azurite and found this pr :)

Would this also work with the Azure SDK for .NET?

I have certifcates created as per their guide and am able to use the docker via

```sh
docker run -p 10000:10000 -p 10001:10001 -p 10002:10002 -v /absolute-path-on-my-machine/azurite:/workspace  mcr.microsoft.com/azure-storage/azurite azurite --blobHost 0.0.0.0  --queueHost 0.0.0.0 --tableHost 0.0.0.0 --oauth basic --cert /workspace/127.0.0.1.pem --key /workspace/127.0.0.1-key.pem
```

and consume via

```csharp
var blobServiceClient = new BlobServiceClient(new Uri("https://127.0.0.1:10000/devstoreaccount1"), new DefaultAzureCredential());
        var containerName = "quickstartblobsbe2dbb6d-efe9-40e6-a8ad-303cffacb9eb";
        var containerClient = blobServiceClient.GetBlobContainerClient(containerName);
        await containerClient.CreateIfNotExistsAsync();
```

---

But I am having trouble with doing the same in TestContainers
The following works (but without azure credential)

```csharp
var testcontainersBuilder =
            new TestcontainersBuilder<TestcontainersContainer>()
                .WithImage(image)
                .WithPortBinding(10000)
                .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(10000));
        var testContainer = testcontainersBuilder.Build();
        await testContainer.StartAsync();

var connectionString = "UseDevelopmentStorage=true;"; 
```

I tried the following

```csharp
var testcontainersBuilder =
            new TestcontainersBuilder<TestcontainersContainer>()
                .WithImage(image)
                .WithBindMount("/Users/maxhamulyak/Development/BlogTopics/AzureStorage/azurite", "/workspace")
                .WithCommand("azurite", "--blobHost 0.0.0.0", "--oauth basic", "--cert /workspace/127.0.0.1.pem", " -key /workspace/127.0.0.1-key.pem")
                .WithPortBinding(10000)
                .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(10000));
```


But that results in container is not running

Any hints how to enable DotnetTestcontainers with ssl and --oauth basic would be great.

Thanks
Max

HofmeisterAn:
> Any hints how to enable DotnetTestcontainers with ssl and --oauth basic would be great.

In your example you are using `" -key /workspace/127.0.0.1-key.pem"` instead of `--key ...`. Did you try something like:
`.WithCommand("azurite", "--blobHost", "0.0.0.0", "--oauth", "basic", "--cert", "/workspace/127.0.0.1.pem", "--key", "/workspace/127.0.0.1-key.pem")`? What error do you get?

kaylumah:
> > Any hints how to enable DotnetTestcontainers with ssl and --oauth basic would be great.
> 
> In your example you are using `" -key /workspace/127.0.0.1-key.pem"` instead of `--key ...`. Did you try something like: `.WithCommand("azurite", "--blobHost", "0.0.0.0", "--oauth", "basic", "--cert", "/workspace/127.0.0.1.pem", "--key", "/workspace/127.0.0.1-key.pem")`? What error do you get?

Ah you spotted my mistake :)
I tried various syntax and apparently in the last one I forgot one dash before key.

```csharp
var testcontainersBuilder =
            new TestcontainersBuilder<TestcontainersContainer>()
                .WithImage(image)
                .WithBindMount("/absolute-local-path/azurite", "/workspace")
                .WithCommand("azurite", "--blobHost", "0.0.0.0", "--oauth", "basic", "--cert", "/workspace/127.0.0.1.pem", "--key", "/workspace/127.0.0.1-key.pem")
                .WithPortBinding(10000)
                .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(10000));

        var testContainer = testcontainersBuilder.Build();
        await testContainer.StartAsync();
        

        var blobServiceClient = new BlobServiceClient(new Uri("https://127.0.0.1:10000/devstoreaccount1"), new DefaultAzureCredential());
        var containerName = "quickstartblobsbe2dbb6d-efe9-40e6-a8ad-303cffacb9eb";
        var containerClient = blobServiceClient.GetBlobContainerClient(containerName);
        await containerClient.CreateIfNotExistsAsync();
        
        var blobFileName = Guid.NewGuid().ToString();
        var blobClient1 = containerClient.GetBlobClient(blobFileName);
        await blobClient1.UploadAsync(BinaryData.FromString("Hello World!"));

        var blobClient2 = containerClient.GetBlobClient(blobFileName);
        using var ms = new MemoryStream();
        await blobClient2.DownloadToAsync(ms);
        ms.Close();
        var text = new UTF8Encoding().GetString(ms.ToArray());
```

Where `/absolute-local-path/azurite` contains the cert files created my mkcert

Thanks again 
Max

HofmeisterAn:
@kaylumah is it working or do you still get an error?

kaylumah:
> @kaylumah is it working or do you still get an error?

With the snippet I shared it is working, including SSL and OAUTH needed for the .NET Azure SDK see (https://github.com/Azure/Azurite#usage-with-azure-storage-sdks-or-tools)

HofmeisterAn (approved):

HofmeisterAn:
c03a7fd27bb5e1b515cee716ba4d54418b25d838, 596b14e74169d174529742da8bbadf01be4d1bd5 thanks for all of your help.

vlaskal:
> [c03a7fd](https://github.com/testcontainers/testcontainers-dotnet/commit/c03a7fd27bb5e1b515cee716ba4d54418b25d838), [596b14e](https://github.com/testcontainers/testcontainers-dotnet/commit/596b14e74169d174529742da8bbadf01be4d1bd5) thanks for all of your help.

Any time

vlaskal:
> > @kaylumah is it working or do you still get an error?
> 
> With the snippet I shared it is working, including SSL and OAUTH needed for the .NET Azure SDK see (https://github.com/Azure/Azurite#usage-with-azure-storage-sdks-or-tools)

@kaylumah Could you describe to me why do you need use SSL? I thought is not used to often and I would like to know if I should add support for it or it can wait. Thanks a lot

kaylumah:
@vlaskal see https://github.com/Azure/Azurite#azure-sdks if you want to use stuff like oauth authentication like `AzureDefaultCredential` you need the https. I believe its best practice as opposed to putting the shared access key in config

vlaskal:
@kaylumah thank you for pointing this to me. I will try to create anothe PR and add support gor this. Will this help you?

kaylumah:
@vlaskal yeah this will help me :)

Before I found your PR I was doing it like

```sharp
var testcontainersBuilder =
            new TestcontainersBuilder<TestcontainersContainer>()
                .WithImage(image)
                .WithBindMount("/absolute-local-path/azurite", "/workspace")
                .WithCommand("azurite", "--blobHost", "0.0.0.0", "--oauth", "basic", "--cert", "/workspace/127.0.0.1.pem", "--key", "/workspace/127.0.0.1-key.pem")
                .WithPortBinding(10000)
                .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(10000));
```

Which does the trick, but doing it typed like you added in this PR would be a big step up

vlaskal:
OK, will work on next PR.

Issue #560, opened 2022-08-11
Fixing 589 Warnings on Rider


mason-chase:
**Fixing 589 Warnings on Rider**
Currently
 
**To Reproduce**
1. Download [Jetbrains Rider ](https://www.jetbrains.com/rider/)
2. Click on problems
3. Check to enable "Warnings"
4. See Warnings

![image](https://user-images.githubusercontent.com/17309021/184215751-b95ca707-abd0-45f6-b875-e79fa9e260b2.png)

**Desktop (please complete the following information):**

**Additional context**

PR #562, opened 2022-08-11
chore: Fixing 589 warnings on JetBrains Rider

mason-chase:
**Fixing 589 Warnings on Rider**
Currently
 
**To Reproduce**
1. Download [Jetbrains Rider ](https://www.jetbrains.com/rider/)
2. Click on problems
3. Check to enable "Warnings"
4. See Warnings

image

After:

image

### DIFF
--- a/Testcontainers.sln.DotSettings
+++ b/Testcontainers.sln.DotSettings
@@ -1,0 +1,8 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/CodeStyle/CSharpUsing/QualifiedUsingAtNestedScope/@EntryValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=CAPI/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=memopt/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=RAMSIZE/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=ryuk/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Testcontainer/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Testcontainers/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/Testcontainers.sln.DotSettings
+++ b/Testcontainers.sln.DotSettings
@@ -1,4 +1,6 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CheckNamespace/@EntryIndexedValue">HINT</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=RedundantNameQualifier/@EntryIndexedValue">HINT</s:String>
 	<s:Boolean x:Key="/Default/CodeStyle/CSharpUsing/QualifiedUsingAtNestedScope/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=CAPI/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=memopt/@EntryIndexedValue">True</s:Boolean>
--- a/src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
+++ b/src/Testcontainers/Configurations/Containers/ITestcontainersConfiguration.cs
@@ -1,3 +1,4 @@
+// ReSharper disable once CheckNamespace to avoid introducing breaking changes
 namespace DotNet.Testcontainers.Configurations
 {
   using System;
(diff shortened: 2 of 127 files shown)


HofmeisterAn (commented):
I don't agree with JetBrains (Rider or ReSharper) here. IMO, namespaces don't necessarily need to be equal to the directory structure. I even think the current namespaces are more convenient for devs (combines similar namespaces together) and contributes (allows organizing with subdirectories).

I am fine configuring things like that in a `.editorconfig` file (central configuration), but I don't want to add configurations for all kinds of static code analyzers distributed in the repository. There are so many of them in .NET and all of them work differently. I am just tired of having all those configurations in the repository without much value.

Having a configuration what works well within in the IDE is great, but I don't like to add IDE specific configurations to the source code.

vlaskal:
I created team Resharper configuration file for myself as well to be honest. But I putting rules based on style. I think it is good idea to have that one which sets rules for Rider, Resharper and Visual Studio. There is a lot of text undersocred and it will helps for orientation. Also generators setting should be set. Eg. fully defined namespaces in inner usings etc.
The question is how to set it up.

HofmeisterAn:
> I think it is good idea to have that one which sets rules for Rider, Resharper and Visual Studio.

From my experience there is no configuration that covers all IDEs well. There are rules that are not available in one of the other. Sometimes rules even conflict with each other. The IDEs suggest different recommendations for the same rule. Keeping the configurations in sync is another topic...

> There is a lot of text undersocred and it will helps for orientation.

Hm, I think it's just the namespaces and some words or abbreviations Rider or ReSharper are not aware of. We have no build warnings or something line that.

> The question is how to set it up.

Having the same rules (configurations) across all common IDEs would be great 👍. IMO it's necessary to support other (most common) IDEs too. Everyone should be able to contribute with the tools they like.

Is it possible to share the same configuration between Visual Studio (w/o ReSharper) and Rider for example?

mason-chase:
I understand that Jetbrains Rider may not be to your taste, however, the rule is that namespace must match folder and if not it should indicate to developer that there is a puporse for this convension, specially if you are using a popular IDE that is throwing unsolved warnings, it makes it difficult to collebate with other developers when you still have warnings that are pending.

It would open your code for a broader community of developers to contribute to code.

mason-chase:
I must add that I find this project very interesting and I would to invite my development teams to use, that also means that we will actively contribute to the project to find bugs we can report or address with a PR

HofmeisterAn:
> I understand that Jetbrains Rider may not be to your taste

I use Rider as well. I don't have any issues with those warnings. I just care about the build output and Sonar (same results across all development environments).

> the rule is that namespace must match folder

As I said, I don't agree with that rule. If you look into the namespaces you're using in a test, the current approach is much cleaner.

> specially if you are using a popular IDE that is throwing unsolved warnings

That's the point. It's just the (one) IDE. If you're using a different IDE, you won't benefit from those rules. You won't notice any warnings. If someone creates a pull request, I don't know if all Rider or ReSharper rules are considered. This might work fine for companies that use the same development environment.

Improving the current configuration is good, but I don't like to pollute all classes with IDE specific rules. Especially if they are not considered by other IDEs.

I think a proper `.editorconfig` and `DotSettings` file that cover Rider, ReSharper, Visual Studio and the different Roslyn analyzers would be a better approach.

mason-chase:
> > I understand that Jetbrains Rider may not be to your taste
> 
> I use Rider as well. I don't have any issues with those warnings. I just care about the build output and Sonar (same results across all development environments).
> 
> > the rule is that namespace must match folder
> 
> As I said, I don't agree with that rule. If you look into the namespaces you're using in a test, the current approach is much cleaner.
> 
> > specially if you are using a popular IDE that is throwing unsolved warnings
> 
> That's the point. It's just the (one) IDE. If you're using a different IDE, you won't benefit from those rules. You won't notice any warnings. If someone creates a pull request, I don't know if all Rider or ReSharper rules are considered. This might work fine for companies that use the same development environment.
> 
> Improving the current configuration is good, but I don't like to pollute all classes with IDE specific rules. Especially if they are not considered by other IDEs.
> 
> I think a proper `.editorconfig` and `DotSettings` file that cover Rider, ReSharper, Visual Studio and the different Roslyn analyzers would be a better approach.

You have a good point and valid point Andre
I will try to find an alternative and fix the PR


Cheers
Mason

mason-chase:
Hi @HofmeisterAn  

I have adjusted the PR to ensure the code is minimal within the code base, 
Can you please review it again?

Cheers
Mason

Testcontainers.sln.DotSettings
```diff
@@ -0,0 +1,10 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CheckNamespace/@EntryIndexedValue">HINT</s:String>
```
HofmeisterAn, line 2: Can we move the Rider / ReSharper related settings from the `.editorconfig` to this file  or the other way around? It would be nice to have all settings in one configuration file:
https://github.com/testcontainers/testcontainers-dotnet/blob/49c876d02260721cc654ee1c5dce9963ce18924d/.editorconfig#L462-L467

Testcontainers.sln.DotSettings
```diff
@@ -0,0 +1,10 @@
...
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=RedundantNameQualifier/@EntryIndexedValue">HINT</s:String>
+	<s:Boolean x:Key="/Default/CodeStyle/CSharpUsing/QualifiedUsingAtNestedScope/@EntryValue">True</s:Boolean>
```
HofmeisterAn, line 4: Do you know how the [custom dictionaries](https://www.jetbrains.com/help/rider/Settings_Spelling.html) work? I think they cover unknown abbreviations too. This would make `SuppressMessage(...)` obsolete.

HofmeisterAn:
> I have adjusted the PR to ensure the code is minimal within the code base,
> Can you please review it again?

Thanks for the changes 🙏.

Testcontainers.sln.DotSettings
HofmeisterAn: Ok, looks like the dictionary does not work with abbreviations. But we can add this to the settings file:

```xml
<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/Abbreviations/=OS/@EntryIndexedValue">OS</s:String>
```

HofmeisterAn:
https://github.com/testcontainers/testcontainers-dotnet/pull/564/commits/96c72633858d7e2f64d457035faaa4803c4b0f21, closes #562, #560. I think this covers most. If there is anything left, just let me know.

mason-chase:
Should I improve this to be merged?

Testcontainers.sln.DotSettings
mason-chase: I will move this

HofmeisterAn:
> Should I improve this to be merged?

Is anything particular missing? Did you check the latest configuration in `develop`?

PR #565, opened 2022-08-21
Fix ExecScriptAsync on MongoDbTestcontainer

ptimothee:
MongoDbTestcontainer ExecScriptAsync should execute mongosh scripts

### DIFF
--- a/src/Testcontainers/Containers/Modules/Databases/MongoDbTestcontainer.cs
+++ b/src/Testcontainers/Containers/Modules/Databases/MongoDbTestcontainer.cs
@@ -1,5 +1,9 @@
-﻿namespace DotNet.Testcontainers.Containers
+namespace DotNet.Testcontainers.Containers
 {
+  using System.Globalization;
+  using System.Text;
+  using System.Threading;
+  using System.Threading.Tasks;
   using DotNet.Testcontainers.Configurations;
   using JetBrains.Annotations;
   using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


src/Testcontainers/Containers/Modules/Databases/MongoDbTestcontainer.cs
```diff
@@ -23,5 +27,26 @@ public override string ConnectionString
...
         : $"mongodb://{this.Username}:{this.Password}@{this.Hostname}:{this.Port}";
+
+    public async Task<ExecResult> ExecMongoScriptAsync(string script, CancellationToken ct = default)
+    {
+      var tempScriptFile = this.GetTempScriptFile();
+
+      await this.CopyFileAsync(tempScriptFile, Encoding.Default.GetBytes(script), 493, 0, 0, ct)
+                      .ConfigureAwait(false);
+
+      return await this.ExecAsync(new[] { "mongo", "db", tempScriptFile }, ct);
```
HofmeisterAn, line 38: Can't we just use the new MongoDB Shell (`mongosh`) and remove `ExecMongoScriptAsync`?

src/Testcontainers/Containers/Modules/Databases/MongoDbTestcontainer.cs
```diff
@@ -23,5 +27,26 @@ public override string ConnectionString
...
+    }
+
+    public override async Task<ExecResult> ExecScriptAsync(string scriptContent, CancellationToken ct = default)
+    {
+      var tempScriptFile = this.GetTempScriptFile();
+
+      await this.CopyFileAsync(tempScriptFile, Encoding.Default.GetBytes(scriptContent), 493, 0, 0, ct)
+                      .ConfigureAwait(false);
+
+      return await this.ExecAsync(new[] { "mongosh", "--port", this.ContainerPort.ToString(CultureInfo.InvariantCulture), "--eval", $"load('{tempScriptFile}')" }, ct)
```
HofmeisterAn, line 48: Can we pass the `ConnectionString` to `mongosh` (at least username and password)?

src/Testcontainers/Containers/Modules/Databases/MongoDbTestcontainer.cs
ptimothee: Absolutely. This was just in case there was a desire to support the legacy shell while it is still available. Will remove, thanks.

src/Testcontainers/Containers/Modules/Databases/MongoDbTestcontainer.cs
ptimothee: Agreed but will need to look into this one. After supplying the credentials to the mongosh I'm able to connect however my user did not have permissions to run scripts it seems. Will need to look into setting up that user before hand when usr/pwd are configured with the container builder.

PR #600, opened 2022-10-03
feat(#594): Add IDockerContainer.GetLogs, rename TestcontainersState to TestcontainersStates

HofmeisterAn:


src/Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -62,6 +63,25 @@ public async Task<long> GetExitCode(string id, CancellationToken ct = default)
...
+        Timestamps = true,
+      };
+
+      using (var stdOutAndErrStream = await this.Docker.Containers.GetContainerLogsAsync(id, false, logsParameters, ct)
+        .ConfigureAwait(false))
+      {
+        return await stdOutAndErrStream.ReadOutputToEndAsync(ct)
+          .ConfigureAwait(false);
+      }
+    }
```
vlaskal, line 83: I think it can be simplified
```suggestion
      using var stdOutAndErrStream = await this.Docker.Containers.GetContainerLogsAsync(id, false, logsParameters, ct).ConfigureAwait(false);
      return await stdOutAndErrStream.ReadOutputToEndAsync(ct).ConfigureAwait(false);
    }
```

vlaskal (approved):

src/Testcontainers/Clients/DockerContainerOperations.cs
HofmeisterAn: You are right, but I do not like `using`s without brackets. I would like to "decide" when it should be disposed (it is easier to read and helps to avoid mistakes IMO).

Issue #604, opened 2022-10-06
Docker image build fail with certain `.dockerignore` configuration


HofmeisterAn:
**Describe the bug**
A `.dockerignore` entry that ends with `/` may deny all files. The Tarball will be empty and the Docker image build will fail.

**To Reproduce**
1. Create `.dockerignore` that contains e.g. `bin/`.
2. Build a Docker image with `ImageFromDockerfileBuilder`.

**Expected behavior**
The docker image build is successful.

**Screenshots**
\-

**Desktop (please complete the following information):**
- `2.0.1`

**Additional context**
\-

PR #605, opened 2022-10-06
fix(#604): Do not deny all files in the Docker image tarball when a .dockerignore entry ends with /

HofmeisterAn:


HofmeisterAn:
Closes #604.

Issue #601, opened 2022-10-04
Option to always pull image when creating a container


BenasB:
**Is your feature request related to a problem? Please describe.**
I have a case where my tests should always retrieve the latest image. AFAIK, this library now pulls an image only if it's missing (sort of like `docker run --pull "missing" ...`, which is the default) but I would like to **always** pull the image (so it gets the most up to date version of it) before starting the container (like `docker run --pull "always" ...` does).

**Describe the solution you'd like**
An option that enables this behavior would be perfect, for example a `.WithImagePull(PullType)` builder method where `PullType` could be `Missing` (default behavior) or `Always`. Although I am not sure how this could be implemented.

**Describe alternatives you've considered**
The image pull can be done explicitly before building/starting the container by using [Docker.DotNet](https://github.com/dotnet/Docker.DotNet) directly and this would be okay, but testcontainers-dotnet already has support for credential helpers (AFAIK Docker.DotNet does not support that) which I'm trying to take advantage of (since I'm authenticating to a private registry)

How much would this be feasible? Thanks!

HofmeisterAn:
An equivalent to the Java [withImagePullPolicy](https://www.testcontainers.org/features/advanced_options/#image-pull-policy) would be great.

BenasB:
That's exactly what I've imagined, although I don't have experience with the Java version. I will try to take a look at how it's implemented there and see how I could implement a similar feature for testcontainers-dotnet

kiview:
@BenasB Let us know if you have any questions regarding the Java implementation 🙂

PR #606, opened 2022-10-06
Implement WithImagePullPolicy 

BenasB:
Trying to implement `WithImagePullPolicy` mentioned in #601. I tried to keep similar functionality to the [Java counterpart](https://www.testcontainers.org/features/advanced_options/#image-pull-policy) where you can write a custom policy depending on the information about the cached image (e. g. policy that pulls the image if the currently cached image's  `Created` property is 3 days old). Of course there's still the presets `PullPolicy.Never`, `PullPolicy.Missing` (default) and `PullPolicy.Always` like with `docker run` argument `--pull`.

I am not sure how I should proceed with tests – ideally I would want to for example test if `PullPolicy.Never` is working, so I would need to somehow delete the image if it's cached/pulled, but I am not sure how to do that, since I can't access the Docker.DotNet client underneath. Any suggestions appreciated.

How is this looking @HofmeisterAn ?

### DIFF
--- a/src/Testcontainers/Clients/TestcontainersClient.cs
+++ b/src/Testcontainers/Clients/TestcontainersClient.cs
@@ -259,8 +259,10 @@ namespace DotNet.Testcontainers.Clients
           .ConfigureAwait(false);
       }
 
-      if (!await this.images.ExistsWithNameAsync(configuration.Image.FullName, ct)
-        .ConfigureAwait(false))
+      var cachedImage = await this.images.ByNameAsync(configuration.Image.FullName, ct)
+        .ConfigureAwait(false);
+
+      if (configuration.ImagePullPolicy(cachedImage))
       {
         var authConfig = default(DockerRegistryAuthenticationConfiguration).Equals(configuration.DockerRegistryAuthConfig)
           ? this.registryAuthenticationProvider.GetAuthConfig(configuration.Image.GetHostname()) : configuration.DockerRegistryAuthConfig;
--- a/tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
+++ b/tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
@@ -7,12 +7,13 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
   using System.Net.Http;
   using System.Text;
   using System.Threading.Tasks;
-  using Docker.DotNet;
   using DotNet.Testcontainers.Builders;
   using DotNet.Testcontainers.Clients;
   using DotNet.Testcontainers.Configurations;
   using DotNet.Testcontainers.Containers;
+  using DotNet.Testcontainers.Images;
   using DotNet.Testcontainers.Tests.Fixtures;
+  using global::Docker.DotNet;
   using Xunit;
 
   public static class TestcontainersContainerTest
@@ -497,17 +498,17 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
         // Given
         // An image that actually exists but was not used/pulled previously
         // If this image is cached/pulled before, this test will fail
-        var previouslyUnusedImage = "alpine:edge";
+        var uncachedImage = "alpine:edge";
 
         // When
         var testcontainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
-          .WithImage(previouslyUnusedImage)
+          .WithImage(uncachedImage)
           .WithImagePullPolicy(PullPolicy.Never);
 
         // Then
         await using (ITestcontainersContainer testcontainer = testcontainersBuilder.Build())
         {
-          Assert.ThrowsAsync<DockerImageNotFoundException>(() => testcontainer.StartAsync());
+          await Assert.ThrowsAsync<DockerImageNotFoundException>(() => testcontainer.StartAsync());
         }
       }
     }
--- a/tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
+++ b/tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
@@ -497,7 +497,7 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
         // Given
         // An image that actually exists but was not used/pulled previously
         // If this image is cached/pulled before, this test will fail
-        var uncachedImage = "alpine:edge";
+        const string uncachedImage = "alpine:edge";
 
         // When
         var testcontainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
--- a/tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
+++ b/tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
@@ -13,7 +13,6 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
   using DotNet.Testcontainers.Containers;
   using DotNet.Testcontainers.Images;
   using DotNet.Testcontainers.Tests.Fixtures;
-  using global::Docker.DotNet;
   using Xunit;
 
   public static class TestcontainersContainerTest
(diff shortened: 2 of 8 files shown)


src/Testcontainers/Clients/TestcontainersClient.cs
```diff
@@ -259,8 +259,10 @@ public async Task<string> RunAsync(ITestcontainersConfiguration configuration, C
           .ConfigureAwait(false);
       }
 
-      if (!await this.images.ExistsWithNameAsync(configuration.Image.FullName, ct)
-        .ConfigureAwait(false))
+      var cachedImage = await this.images.ByNameAsync(configuration.Image.FullName, ct)
+        .ConfigureAwait(false);
+
+      if (configuration.ImagePullPolicy(cachedImage))
```
HofmeisterAn, line 265: Does it pull the image again, even if it exists (or do we need to remove it first)?

HofmeisterAn (commented):
👏 thanks for your contribution. PR looks good. I have just one question below.

> I am not sure how I should proceed with tests – ideally I would want to for example test if `PullPolicy.Never` is working

You can try to pull an image we do not use, like `alpine:edge`. `StartAsync` will thrown an exception that it could not find the image.

src/Testcontainers/Clients/TestcontainersClient.cs
BenasB: I don't understand your question fully, but it depends on what `ImagePullPolicy` returns (if it's `true` then it will pull the image, otherwise it skips the pulling). You can think of it like conditionally running `docker pull` before always running `docker run --pull never`.

There are three policies ready to use (in [`PullPolicy`](https://github.com/BenasB/testcontainers-dotnet/blob/feat/601_image_pull_policy/src/Testcontainers/Images/PullPolicy.cs)):

1. `PullPolicy.Never` will always skip the `docker pull` part and the `docker run --pull never` could fail if the cached image is not found (if you didn't run `docker pull` manually). This might be useful in scenarios where you want to manually pull the image before running the tests and not let testcontainers library influence the image in any way.
2. `PullPolicy.Missing` will only do `docker pull` if a cached image is not found. This is what happens by default in CLI `docker run` (no `--pull` argument) when you run for example `docker run alpine:latest` and then a month later (when there's a new `latest` version) run `docker run alpine:latest` again - it will not pull the new `latest` unless you `docker pull` it explicitly/manually. This is also the default pull policy in our library if you don't specify `WithImagePullPolicy` when building the container (meaning no breaking changes done).
3. `PullPolicy.Always` will always do `docker pull`. This is basically running `docker pull` before you run the `docker run --pull never` every time - it will enforce that the newest version of the tag is pulled. If it's up to date, it will return quickly (just like `docker pull` returns quickly when it checks and finds out the cached image tag is up to date with the registry)

BenasB:
Thanks for the suggestion, yes using an image we don't use would make sense for `PullPolicy.Never`, but not sure how to handle `PullPolicy.Always` - it's useful in scenarios when an image with tag `latest` is cached on the docker client, but there's a new version of the `latest` tag in the registry and you want to always check if your cached version is up to date with the version in the registry.

src/Testcontainers/Clients/TestcontainersClient.cs
HofmeisterAn: I understood the policies, but did you check if

```csharp
await this.images.CreateAsync(configuration.Image, authConfig, ct)
  .ConfigureAwait(false);
```

downloads (updates) an image again if it already exists?

HofmeisterAn:
> Thanks for the suggestion, yes using an image we don't use would make sense for `PullPolicy.Never`, but not sure how to handle `PullPolicy.Always` - it's useful in scenarios when an image with tag `latest` is cached on the docker client, but there's a new version of the `latest` tag in the registry and you want to always check if your cached version is up to date with the version in the registry.

Lets skip that for now. If `CreateAsync` overrides existing images we are good. I do not think `PullPolicy.Always` is easy to test.

src/Testcontainers/Clients/TestcontainersClient.cs
BenasB: Ahh sorry, I misunderstood you at first. Yes, [it goes directly](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers/Clients/DockerImageOperations.cs#L78) to Docker.DotNet's [`CreateImageAsync`](https://github.com/dotnet/Docker.DotNet/blob/master/src/Docker.DotNet/Endpoints/IImageOperations.cs#L86), which is equivalent to `docker pull`. As I briefly mentioned in #601 as a workaround I could call Docker.Dotnet's `CreateImageAsync` explicitly and achieve what I wanted ("always" policy). 

I will manually test this again just to be sure once I get the opportunity to.

HofmeisterAn: > as a workaround I could call Docker.Dotnet's CreateImageAsync explicitly and achieve what I wanted ("always" policy).

Ah sorry, missed that somehow. Do you want to add the tests for `PullPolicy.Never`?

BenasB: Yes I will add the test and mark this PR as ready for review

tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
```diff
@@ -7,6 +7,7 @@ namespace DotNet.Testcontainers.Tests.Unit.Containers.Unix
   using System.Net.Http;
   using System.Text;
   using System.Threading.Tasks;
+  using Docker.DotNet;
```
HofmeisterAn: `using DotNet.Testcontainers.Images;`

tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
```diff
@@ -489,6 +490,26 @@ await using (ITestcontainersContainer testcontainer = testcontainersBuilder.Buil
...
         }
       }
+
+      [Fact]
+      public async Task PullPolicyNever()
+      {
+        // Given
+        // An image that actually exists but was not used/pulled previously
+        // If this image is cached/pulled before, this test will fail
+        var previouslyUnusedImage = "alpine:edge";
```
HofmeisterAn: `const string uncachedImage = "alpine:edge";`

tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
```diff
@@ -489,6 +490,26 @@ await using (ITestcontainersContainer testcontainer = testcontainersBuilder.Buil
...
+
+        // When
+        var testcontainersBuilder = new TestcontainersBuilder<TestcontainersContainer>()
+          .WithImage(previouslyUnusedImage)
+          .WithImagePullPolicy(PullPolicy.Never);
+
+        // Then
+        await using (ITestcontainersContainer testcontainer = testcontainersBuilder.Build())
+        {
+          Assert.ThrowsAsync<DockerImageNotFoundException>(() => testcontainer.StartAsync());
```
HofmeisterAn: `await Assert.ThrowsAsync<Exception>(() => testcontainer.StartAsync());`

HofmeisterAn (changes requested):

tests/Testcontainers.Tests/Unit/Containers/Unix/TestcontainersContainerTest.cs
BenasB: `ThrowsAsync` won't catch `Exception` here since it catches only the **exact** exception type, see: https://stackoverflow.com/a/31686148. So I switched to `ThrowsAnyAsync`.

BenasB:
Sorry, I only had access to my corporate PC where I can't pull public GitHub repos, so I ended up winging it in the online vscode editor 😅 Finally got the chance to fix it, should be okay now. Excuse my commit history, it's a bit all over the place.

HofmeisterAn (approved):
> Excuse my commit history, it's a bit all over the place.

No worries, we squash it anyway. Thanks again for your contribution.

HofmeisterAn:
c48d858213db2e50c833ba03337ef7a1df6d8b79, closes #601.

PR #613, opened 2022-10-14
docs: Add persistence layer (MSSQL) to WeatherForecast example

HofmeisterAn:


HofmeisterAn:
@testcontainers/oss-team does anyone would like to review .NET PRs?

examples/WeatherForecast/src/WeatherForecast.Entities/HasId.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+  [JsonConstructor]
+  public HasId(Guid id)
+  {
+    Id = id;
+  }
+
+  [Key]
+  [JsonPropertyName("id")]
+  public Guid Id { get; }
```
eddumelendez, line 19: readonly?

examples/WeatherForecast/src/WeatherForecast/Program.cs
```diff
@@ -9,9 +12,22 @@
...
+  builder.Services.AddSingleton<IWeatherDataWriteOnlyRepository, WeatherDataWriteOnlyRepository>();
+  builder.Services.AddSingleton<ISearchCityOrZipCode, SearchCityOrZipCode>();
+}
+else
+{
+  builder.Services.AddDbContext<WeatherDataContext>(options => options.UseSqlServer(connectionString));
+  builder.Services.AddScoped<IWeatherDataReadOnlyRepository, WeatherDataReadOnlyContext>();
+  builder.Services.AddScoped<IWeatherDataWriteOnlyRepository, WeatherDataWriteOnlyContext>();
+  builder.Services.AddScoped<ISearchCityOrZipCode, SearchCityOrZipCode>();
+}
```
eddumelendez, line 30: should we have all singleton? and just

```
if (!string.IsNullOrWhiteSpace(connectionString)) {
    builder.Services.AddDbContext<WeatherDataContext>(options => options.UseSqlServer(connectionString));
}
...
```

eddumelendez: in general, I wonder if for integration test purposes would have only the else branch. Also, should remove `WeatherDataReadOnlyRepository` and `WeatherDataWriteOnlyRepository` for the example simplicity.

eddumelendez (commented):
sorry for the delay. Just adding few comments. Nothing critical but would like to understand few things related to dotnet.

thanks to @berczeck for your support reviewing this PR!

examples/WeatherForecast/src/WeatherForecast.Entities/HasId.cs
HofmeisterAn: Yes, everything is immutable. Do you see any disadvantages?

examples/WeatherForecast/src/WeatherForecast/Program.cs
HofmeisterAn: I kept the if/else, that devs can run the application without having a database (instead of just running the tests). But we can also use Testcontainers to spin up the database in the application if it is missing, might be a nice gimmick.

> should we have all singleton?

`DbContext` must be scoped.

PR #629, opened 2022-10-22
fix #626

### LINKED ISSUE #626: Copy of tc-java#3893
**Describe the bug**
The mysql container does not allow the username `root` for the `MYSQL_USER`, which makes the current TestContainerConfiguration not useful as the container will fail to start. 
https://github.com/testcontainers/testcontainers-dotnet/blob/08e7fa5a6e1d08c02bf249db79ac57e097276ba6/src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs#L40-L44

**To Reproduce**
```csharp
_container = new TestcontainersBuilder<MySqlTestcontainer>()
    .WithDatabase(new MySqlTestcontainerConfiguration { Database = "datadonation", Username = "root", Password = "example" })
    .WithImage("mysql:8.0.29")
    .WithCleanUp(true)
    .Build();
```

**Expected behavior**
Should incoporate the possibility to use `root` user (e.g. for migrations)

**Additional context**
Basically a copy of https://github.com/testcontainers/testcontainers-java/issues/3893 for dotnet.

DanielHabenicht:
fix handling of "root" username for mysql container


src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs
```diff
@@ -38,7 +38,8 @@ public override async Task<ExecResult> ExecScriptAsync(string scriptContent, Can
       await this.CopyFileAsync(tempScriptFile, Encoding.Default.GetBytes(scriptContent), 493, 0, 0, ct)
         .ConfigureAwait(false);
 
-      return await this.ExecAsync(new[] { "mysql", $"--host={this.Hostname}", $"--port={this.ContainerPort}", $"--user={this.Username}", $"--password={this.Password}", this.Database, $"--execute=source {tempScriptFile}" }, ct)
+      // TODO: Shouldn't we always use localhost, if we are inside the container? With 'Hostname' it is failing in Codespaces
```
DanielHabenicht: That is an open question. In Codespaces it always showed an IP address which could not be resolved within the container.

DanielHabenicht:
error seems unrelated to my change? 
Unrelated to that: Is there a way we could change the cake config, so one could build the nuget package locally too? 

https://github.com/testcontainers/testcontainers-dotnet/blob/08e7fa5a6e1d08c02bf249db79ac57e097276ba6/build.cake#L133

HofmeisterAn:
> error seems unrelated to my change?

Yes, the error does not relate to your change. I addressed it to the dev that contributed Azurite. I will take a look at your changes in the next days.

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
```diff
@@ -39,15 +42,36 @@ public override string Database
...
+      }
     }
 
     /// <inheritdoc />
     public override string Password
     {
-      get => this.Environments["MYSQL_PASSWORD"];
-      set => this.Environments["MYSQL_PASSWORD"] = value;
+      get {
+        if(this.Environments.ContainsKey("MYSQL_USER")) {
```
HofmeisterAn: Do we need to distinguish here? Can't we set `MYSQL_PASSWORD` and `MYSQL_ROOT_PASSWORD` together? Otherwise, it would require the dev to set the properties in a certain order.

HofmeisterAn (commented):
> Unrelated to that: Is there a way we could change the cake config, so one could build the nuget package locally too?

Usually I set `ShouldPublish` to true:

https://github.com/testcontainers/testcontainers-dotnet/blob/08e7fa5a6e1d08c02bf249db79ac57e097276ba6/.cake-scripts/parameters.cake#L55

src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs
HofmeisterAn: Good catch. Yes, the command runs inside the container and fail in DinD environments. I suggest we add `MYSQL_ROOT_PASSWORD` first, then take a look at that. `localhost` looks right though.

vlaskal:
> > error seems unrelated to my change?
> 
> Yes, the error does not relate to your change. I addressed it to the dev that contributed Azurite. I will take a look at your changes in the next days.

@DanielHabenicht and @HofmeisterAn I created bug #631 to track the Azurite failure.

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
DanielHabenicht: seems to work with your approach, although I am not sure about the implications.

DanielHabenicht:
> > Unrelated to that: Is there a way we could change the cake config, so one could build the nuget package locally too?
> 
> Usually I set `ShouldPublish` to true:

Directly in the script? I thought there would be a Command Line Option for that, this way I always have to check what I commit :)

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
HofmeisterAn: I do not think there are any. Java does it similar.

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
```diff
@@ -11,6 +11,9 @@ public class MySqlTestcontainerConfiguration : TestcontainerDatabaseConfiguratio
 
     private const int MySqlPort = 3306;
 
+    private const string RootUsername = "root";
+    private string _username = "root";
```
HofmeisterAn: Can we please remove these unnecessary fields? Usually we store the values directly in the dictionary.

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
```diff
@@ -39,15 +42,29 @@ public override string Database
...
     public override string Username
     {
-      get => this.Environments["MYSQL_USER"];
-      set => this.Environments["MYSQL_USER"] = value;
+      get {
+        return this._username;
+      }
+      set {
+        this._username = value;
+        if(value != RootUsername) {
```
HofmeisterAn: Do we really need to distinguish here? I think we need to make sure `MYSQL_ALLOW_EMPTY_PASSWORD` is set, then it does not matter. Without the if-else it would be easier to read.

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/MySqlTestcontainerNormalUserTest.cs
```diff
@@ -0,0 +1,70 @@
+namespace DotNet.Testcontainers.Tests.Unit
```
HofmeisterAn: Can we put the tests into a shared static class? Something like:

```csharp
public static class MySqlTestcontainerTest
{
  public sealed class MySqlRootUsername : IClassFixture<...>
  {
  }

  public sealed class MySqlCustomUsername : IClassFixture<...>
  {
  }
}
````

File: src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs (hunk 1 above)
Comment (DanielHabenicht) [reply src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs]: So should i create an issue for that and we can fix it in antother PR?

**Review by HofmeisterAn — CHANGES_REQUESTED** [review]
Can you please take a look at my recent comments and fix the code formatting in `MySqlTestcontainerConfiguration`? Then I am happy to merge it.

File: src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs (hunk 1 above)
Comment (HofmeisterAn) [reply src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs]: Yes please, I would like to update the other commands as well.

File: src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs (hunk 4 above)
Comment (DanielHabenicht) [reply src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs]: as we went ahead and set the password for both it should not hurt to also set "both" usernames.

File: src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs (hunk 1 above)
Comment (DanielHabenicht) [reply src/Testcontainers/Containers/Modules/Databases/MySqlTestcontainer.cs]: reverted this commit and created an issue at #632

**Comment by HofmeisterAn** [comment]
> Directly in the script? I thought there would be a Command Line Option for that, this way I always have to check what I commit :)

That is what I do. I ignore those changes with my global `.gitignore`, but OC we can add a mechanism to force skip the `WithCriteria` checks.

File: src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs (hunk 4 above)
Comment (DanielHabenicht) [reply src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs]: Nope, its not working. MySQL Image is complaining if the username is set to root. That's why I made this PR in the first place :D
```
2022-10-23 19:22:17+00:00 [ERROR] [Entrypoint]: MYSQL_USER="root", MYSQL_USER and MYSQL_PASSWORD are for configuring a regular user and cannot be used for the root user
    Remove MYSQL_USER="root" and use one of the following to control the root user password:
    - MYSQL_ROOT_PASSWORD
    - MYSQL_ALLOW_EMPTY_PASSWORD
    - MYSQL_RANDOM_ROOT_PASSWORD
```

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
DanielHabenicht: see [3aa809e](https://github.com/testcontainers/testcontainers-dotnet/pull/629/commits/3aa809ead984eaac32b08c927bb0bbb37d4741c8)

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
DanielHabenicht: I refactored a little: [3aa809e](https://github.com/testcontainers/testcontainers-dotnet/pull/629/commits/3aa809ead984eaac32b08c927bb0bbb37d4741c8)

tests/Testcontainers.Tests/Unit/Containers/Unix/Modules/Databases/MySqlTestcontainerNormalUserTest.cs
DanielHabenicht: yes 
8bd1b53

src/Testcontainers/Configurations/Modules/Databases/MySqlTestcontainerConfiguration.cs
HofmeisterAn: Thanks for checking,

HofmeisterAn (approved):

Issue #640, opened 2022-10-28
All Elasticsearch module tests fail


vlaskal:
**Describe the bug**
All Elasticsearch module tests fail on clean operating system.

**To Reproduce**
1. Install Docker desktop (Rancher desktop, Moby engine)
2. Run Elasticsearch module tests
3. All tests fail with following error message:  `Docker.DotNet.DockerApiException : Docker API responded with status code=Conflict, response={"message":"Container 776f26413636a94482827057706c065e5d697951e882c6a84a0827979d6aea6a is not running"}
`

**Expected behavior**
All tests should pass.

**Screenshots**
![image](https://user-images.githubusercontent.com/42292717/198579845-1717c7ef-8348-44a5-911a-61e206b1af50.png)

**Desktop (please complete the following information):**
- version: 2.1.0
- os: Windows 11 (WSL), Ubuntu Desktop 20.04 LTS
- docker: Docker dekstop, Rancher Desktop, Moby engine (Microsoft build)

**Additional context**
Elasticsearch container does not want to start and fail. Container logs provide the following evidences:
* As the last log message: `ERROR: [1] bootstrap checks failed. You must address the points described in the following [1] lines before starting Elasticsearch.`
* It refers to result of bootstrap check service: `bootstrap check failure [1] of [1]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]`
* Based on the following discussions there is several ways how to solve it:
  1. https://stackoverflow.com/questions/51445846/elasticsearch-max-virtual-memory-areas-vm-max-map-count-65530-is-too-low-inc
  2. https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-virtual-memory.html
  3. https://discuss.elastic.co/t/eck-elasticsearch-pod-not-starting-on-aks-max-virtual-memory-areas-vm-max-map-count-65530-is-too-low-increase-to-at-least-262144/305788
  
  The last one contains information regarding disabling the bootstrap check for non production and for low memory clusters. I think this is the best approach to fix tests. I will try to provide a PR to fix this based on last option.

kiview:
In Java, we implemented this change regarding default heapsize for ES, so it would probably help here as well? 
https://github.com/testcontainers/testcontainers-java/pull/5684

vlaskal:
Thank you @kiview  for quick reply.
I created an PR #641 where I disabled bootstrap check and based on documentation disallow use mmap. 
[...]

PR #641, opened 2022-10-28
fix(#640): Elasticsearch tests run without system level setup

vlaskal:
PR which makes all Elasticsearch tests run-able on any docker engine. It uses option number `iii` from bug #640. 
It also does not affect default setup of Elasticsearch configuration to prevent any breaking change introduction.


HofmeisterAn:
I will close this in favour of #652.

PR #645, opened 2022-10-31
[UPDATE]

navidZee:
remove setter from Username property

### DIFF
--- a/src/Testcontainers/Configurations/Modules/Databases/MsSqlTestcontainerConfiguration.cs
+++ b/src/Testcontainers/Configurations/Modules/Databases/MsSqlTestcontainerConfiguration.cs
@@ -40,7 +40,6 @@ namespace DotNet.Testcontainers.Configurations
     public override string Username
     {
       get => "sa";
-      set => throw new NotImplementedException();
     }
 
     /// <inheritdoc />


HofmeisterAn (commented):
Thanks for your PR. Can you add some context to your change? Why is this important? The MSSQL container does not allow to set a username. Removing the setter might be misleading.

navidZee:
I think when we create ability to set a value to properties, should not throw **NotImplementedException**.
I got confused when using testcontainers to create mssql database. Because following code has error,
image
and when i remove Username property from first initialize that was fine. 
image

HofmeisterAn:
> I think when we create ability to set a value to properties, should not throw **NotImplementedException**.

This behavior is intended. You cannot set the username in the MSSQL Docker container image that easy.

The first screenshot would be misleading if it won't throw an exception. 1) The user is not created as expected, 2) the connection string contains a wrong username, and you won't be able to connect to the database. I think there are two ways to improve it. Either add custom code that creates the user (similar to the database) or add more context (message) to the exception.

HofmeisterAn:
I will close this PR. As mentioned above, this behavior is intended. If you would like to add one of the suggestions above, please follow the [main contributing guidelines](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/docs/contributing.md).

PR #649, opened 2022-11-01
fix(#648): Always include .dockerignore and Dockerfile files in tarball if they exist

### LINKED ISSUE #648: Dockerfile ignored in .dockerignore, not sent in tarball to docker daemon
**Describe the bug**
When a tarball is created, it uses the .dockerignore file to filter the tarball contents and reduce the amount of files that are sent to the docker daemon. However, when a .dockerignore file ignores itself or a Dockerfile these files should still appear in the tarball.

**To Reproduce**
1. Create a new solution and project(s) (mine is a solution with multiple projects with their own dockerfile)
2. Add a .dockerignore file in the root that ignores itself  and any Dockerfile so they won't show up in the image
3. Run the ImageFromDockerfileBuilder
4. Open the tarball from the temp directory
5. See that the .dockerignore file in the root is missing and the Dockerfile files in the project directories are missing
6. See the ImageFromDockerfileBuilder throw an exception that no image was built. This is a bit of a misleading error due to limitations on Docker.DotNet errors. It can't even build an image because the Dockerfile is missing from the tarball.

**Expected behavior**
 When reading the [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#dockerignore-file) it states that:

>You can even use the .dockerignore file to exclude the Dockerfile and .dockerignore files. These files are still sent to the daemon because it needs them to do its job. But the ADD and COPY instructions do not copy them to the image.

I think this means that when creating a tarball to send to the docker daemon the .dockerignore and Dockerfile files should always be included if they exist. If either file appears in the .dockerignore file the ADD and COPY instructions will make sure not to copy them in the resulting image.

**Screenshots**

**Desktop (please complete the following information):**
- Windows 11 Pro 22000.1098
- Docker Desktop for Windows 20.10.17
- Testcontainers 2.2.0

**Additional context**

wterpstra:
Fixes #648 

I took the liberty to refactor the filesystem access out of the DockerIgnoreFile class into a FileSystem class. This way the DockerIgnoreFile class is testable without having to bother with creating temp files. This way I was able to test the dockerignore specific logic without having to change the IgnoreFile class or its tests. Let me know what you think!

### DIFF
--- a/src/Testcontainers/Images/FileSystem.cs
+++ b/src/Testcontainers/Images/FileSystem.cs
@@ -1,0 +1,25 @@
+﻿namespace DotNet.Testcontainers.Images
+{
+  using System.Collections.Generic;
+  using System.IO;
+  using System.Linq;
+
+  public class FileSystem : IFileSystem
+  {
+    public IEnumerable<string> ReadLines(string path)
+    {
+      return File.ReadLines(path);
+    }
+
+    public bool FileExists(string directory, string file)
+    {
+      var directoryInfo = new DirectoryInfo(directory);
+      return directoryInfo.GetFiles(file, SearchOption.TopDirectoryOnly).Any();
+    }
+
+    public string GetDirectoryFullName(string directory)
+    {
+      return new DirectoryInfo(directory).FullName;
+    }
+  }
+}
--- a/src/Testcontainers/Images/FileSystem.cs
+++ b/src/Testcontainers/Images/FileSystem.cs
@@ -4,7 +4,7 @@
   using System.IO;
   using System.Linq;
 
-  public class FileSystem : IFileSystem
+  internal class FileSystem : IFileSystem
   {
     public IEnumerable<string> ReadLines(string path)
     {
--- a/tests/Testcontainers.Tests/Fixtures/Images/DockerIgnoreFileFixture.cs
+++ b/tests/Testcontainers.Tests/Fixtures/Images/DockerIgnoreFileFixture.cs
@@ -37,6 +37,18 @@
         DirectoryFullName = "/mypath",
       });
 
+      var ignoreDirectoryWildcardDockerIgnoreAndWildcardDockerFile = new DockerIgnoreFile("/mypath/", ".dockerignore", logger, new FileSystemFixture()
+      {
+        Lines =
+        {
+          "**/bin/",
+          "**/.dockerignore",
+          "**/Dockerfile",
+        },
+        Exists = true,
+        DirectoryFullName = "/mypath",
+      });
+
       var dockerIgnoreFile = ".dockerignore";
       var dockerfile = "Dockerfile";
       var nestedDockerfile = "/lipsum/lorem/Dockerfile";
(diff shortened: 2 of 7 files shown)


wterpstra:
@HofmeisterAn Not sure what causes this, but have a look at the failing test.

If I have the following .dockerignore
```
**/bin/
**/.dockerignore
**/Dockerfile
```

It will accept a Dockerfile the in the root but it will deny any Dockerfile in a directory. Any idea?

src/Testcontainers/Images/FileSystem.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace DotNet.Testcontainers.Images
+{
+  using System.Collections.Generic;
+  using System.IO;
+  using System.Linq;
+
+  public class FileSystem : IFileSystem
```
HofmeisterAn: Do we need this class? What's wrong with the previous approach? Can we reduce the PR to a set of minimum changes?

HofmeisterAn (changes requested):
> @HofmeisterAn Not sure what causes this, but have a look at the failing test.

Thanks for your PR. I will take a closer look at it tomorrow.

src/Testcontainers/Images/FileSystem.cs
wterpstra: The DockerIgnoreFile class didn't have any tests. The DockereIgnoreFile class uses File.ReadAllLines() and DirectoryInfo to access the filesystem. In order to test any changes you need to either write some temporary files and read those back or abstract the filesystem away and substitute it. I chose to do the latter. Do you prefer a different approach?

HofmeisterAn: > The DockerIgnoreFile class didn't have any tests.

It is implicitly covered by `ImageFromDockerfileTest`.

> I chose to do the latter. Do you prefer a different approach?

Not sure. I like the idea, but otherwise I would like to keep the changes minimal, and avoid unnecessary layers - I mean we are just reading a file. Do you think we can add a `.dockerignore` for each case to the assets and read the files from there (something like `accept_root_docker_files.dockerignore`, `deny_nested_docker_files.dockerignore` maybe you have better names)?

tests/Testcontainers.Tests/Fixtures/Images/DockerIgnoreFileFixture.cs
```diff
@@ -0,0 +1,78 @@
...
+      this.Add(ignoreDirectory, binDirectory, false);
+
+      this.Add(ignoreDirectoryDockerIgnoreAndDockerFile, dockerIgnoreFile, true);
+      this.Add(ignoreDirectoryDockerIgnoreAndDockerFile, dockerfile, true);
+      this.Add(ignoreDirectoryDockerIgnoreAndDockerFile, nestedDockerfile, true);
+      this.Add(ignoreDirectoryDockerIgnoreAndDockerFile, binDirectory, false);
+
+      this.Add(ignoreDirectoryWildcardDockerIgnoreAndWildcardDockerFile, dockerIgnoreFile, true);
+      this.Add(ignoreDirectoryWildcardDockerIgnoreAndWildcardDockerFile, dockerfile, true);
+      this.Add(ignoreDirectoryWildcardDockerIgnoreAndWildcardDockerFile, nestedDockerfile, true);
```
HofmeisterAn, line 74: I think you expect a wrong behavior here. The `IgnoreFile` expects all paths within the context. It does not make sense to ignore (accept / deny) files that are not within the context (the implementation sees the paths as part of `/mypath`). Without double-checking the code, I expect ignoring this file is correct.

HofmeisterAn:
I will close this in favor of #708. I've added you as a co-author. Thanks for your contribution.

PR #660, opened 2022-11-07
refactor: Normalize container paths to forward slashes

HofmeisterAn:
## What does this PR do?

It normalizes some paths to forward slashes.

## Why is it important?

Interacting with most containers or f.g. tarballs requires paths to use a forward slash as path separator. Running Testcontainers on Windows combines paths with backward slashes. This is not always compatible (like in the mentioned cases above). Replace backward with forward slashes if it is required.

This PR contains some changes that do not relate to the description above regarding consistent spelling and refactoring a test.

## Related issues

\-

### DIFF
--- a/src/Testcontainers/Clients/TestcontainersClient.cs
+++ b/src/Testcontainers/Clients/TestcontainersClient.cs
@@ -183,6 +183,9 @@ namespace DotNet.Testcontainers.Clients
     /// <inheritdoc />
     public async Task CopyFileAsync(string id, string filePath, byte[] fileContent, int accessMode, int userId, int groupId, CancellationToken ct = default)
     {
+      IOperatingSystem os = new Unix(dockerEndpointAuthConfig: null);
+      var containerPath = os.NormalizePath(filePath);
+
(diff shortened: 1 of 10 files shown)


src/Testcontainers/Clients/TestcontainersClient.cs
```diff
@@ -212,7 +215,7 @@ await tarOutputStream.WriteAsync(fileContent, 0, fileContent.Length, ct)
 
         memStream.Seek(0, SeekOrigin.Begin);
 
-        await this.containers.ExtractArchiveToContainerAsync(id, "/", memStream, ct)
+        await this.containers.ExtractArchiveToContainerAsync(id, Path.AltDirectorySeparatorChar.ToString(), memStream, ct)
```
kiview, line 218: Quite a lot of [docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-6.0), just to explain
> The value of this field is a slash ('/') on both Windows and Unix-based operating systems.

😅

Issue #654, opened 2022-11-05
Add `WithOption` method into `ITestcontainersNetworkBuilder`


vlaskal:
**Is your feature request related to a problem? Please describe.**
There is not posibility to set additinal options when custome network is created. In my cas it is need to set `com.docker.network.driver.mtu` value when developer runs tests and is connected via VPN.

**Describe the solution you'd like**
Suggested solution is simple add following method into ` ITestcontainersNetworkBuilder`
```csharp
    /// <summary>
    /// Sets additional options of the Docker network.
    /// </summary>
    /// <param name="option">The additional option.</param>
    /// <param name="value">The option value.</param>
    /// <returns>A configured instance of <see cref="ITestcontainersNetworkBuilder" />.</returns>
    [PublicAPI]
    ITestcontainersNetworkBuilder WithOption(string option, string value);
```

**Describe alternatives you've considered**
Nothing else tested.

This is important to be able to use Rancher dekstop instead of Docker desktop where this is not needed, But developers can set even other options via this method therefore I think it should be added.

kiview:
I'd like to add, that Testcontainers for Java indirectly supports such a use case as well, through falling back to using `docker-java` directly:

```java
new BrowserWebDriverContainer<>()
            .withNetwork(Network.builder().createNetworkCmdModifiers(Arrays.asList(cmd -> {
                      cmd.withOptions(new HashMap<String, String>() {{
                        put("com.docker.network.driver.mtu", "1400");
                      }});
                    }))
                    .build())
```

HofmeisterAn:
> I'd like to add, that Testcontainers for Java indirectly supports such a use case as well, through falling back to using `docker-java` directly:

I was thinking about something similar, but this would require more changes. With the new module system I would like to expose the Docker.DotNet types generic to each builder. For now I have merged it, we need to refactor it anyway.

PR #661, opened 2022-11-08
chore(#654): Update CHANGELOG.md, refactore TestcontainerNetworkBuilderTest

HofmeisterAn:


tests/Testcontainers.Tests/Unit/Networks/TestcontainerNetworkBuilderTest.cs
```diff
@@ -1,77 +1,127 @@
...
+      using var dockerClientConfiguration = TestcontainersSettings.OS.DockerEndpointAuthConfig.GetDockerClientConfiguration();
+      using var dockerClient = dockerClientConfiguration.CreateClient();
 
       // When
-      var network = new TestcontainersNetworkBuilder()
-        .WithName(networkName)
-        .WithLabel("label", networkLabel)
-        .WithOption(networkMtuKey, networkMtuValue)
-        .Build();
+      var networkResponse = await dockerClient.Networks.InspectNetworkAsync(this.network.Id)
```
vlaskal, line 60: This line of command is in section `when` but we do not test `InspectNetworkAsync` method but I think we test `IDockerNetwork.Create()` method for specific network setup. I think that the test is not clear eanough what does it want to test.

HofmeisterAn: That's true somehow. I saw it like 1) we have a client, 2) inspecting a network resource and 3) expecting it to have a certain label or option.

Whould you recommend to remove the GWT?

Issue #493, opened 2022-06-27
Publish dedicated modules (pre-configured containers)


HofmeisterAn:
**Is your feature request related to a problem? Please describe.**
Testcontainers modules (pre-configured containers) don't use their hole potential. Dedicated modules can offer more functionalities like including vendor dependencies etc.

**Describe the solution you'd like**
Move each module into its own project (inside this repository) and deploy it as a dependency.

- [x] Replace module extension methods with a module API
- [x] Add lifecycle callbacks (events)
- [x] Add a dedicated example module (`Testcontainers.Example`, `Testcontainers.Example.Tests`, .NET template)
- [x] Publish example module to https://int.nugettest.org/
- [x] Add a common (shared) test structure (xUnit test attributes, logging configuration, test assets, etc.)
    - _xUnit test attributes_: Run tests only on specific environments
- [x] Split tests into multiple projects (especially tests that test static classes or properties)
- [x] Add a Cake task to deploy modules as NuGet dependencies
- [x] Migrate `AzuriteTestcontainer`
- [x] ~Migrate `CosmosDbTestcontainer`~
- [x] Migrate `CouchbaseTestcontainer`
- [x] Migrate `CouchDbTestcontainer`
- [x] Migrate `ElasticsearchTestcontainer`
- [x] Migrate `MariaDbTestcontainer`
- [x] Migrate `MongoDbTestcontainer`
- [x] Migrate `MsSqlTestcontainer`
- [x] Migrate `MySqlTestcontainer`
- [x] Migrate `Neo4jTestcontainer`
- [x] Migrate `OracleTestcontainer`
- [x] Migrate `PostgreSqlTestcontainer`
- [x] Migrate `RedisTestcontainer`
- [x] Migrate `KafkaTestcontainer`
- [x] Migrate `LocalStackTestcontainer`
- [x] Migrate `RabbitMqTestcontainer`

**Describe alternatives you've considered**
\-

**Additional context**

- Relates #421
- Relates #504
- Relates #507
- Relates #518
- Relates #747

## Follow-up

- [x] Should we add a base class e.g. for database modules such as `DatabaseContainer` to share commen functionalities?
- [x] Should the **Microsoft SQL Server** module create a custom database while the container starts? This is a general discussion we had already. How much or which convenient functionalities should a module contain? How is responsible for it?
[...]

PR #673, opened 2022-11-12
Seperate database/messagebroker modules to separate projects

KijongHan:
Thought I'd try giving this one a crack #493, atm its just the baseline module migrations into their separate projects.

I had considered separating the test project into separate Databases/MessageBrokers/more generic Docker test projects, I wonder if thats inline with how the general structure of the test projects is envisioned going forward.

Todo:
- [x] Separate modules into individual projects
- [ ] Split tests into multiple test projects and add common test structure
- [ ] Add/Update Cake task to deploy modules as Nuget dependencies


HofmeisterAn (commented):
Thanks for your PR. Like mentioned in the [main contributing guidelines](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/docs/contributing.md) it is better to discuss changes with the authors first, especially large changes.

#493 is not just about moving the current implementation into dedicated packages. The builder pattern has major flaws in regard to creating module specific container instances. It is necessary to fix this design flaw first. Afterwards we can start to migrate module after module incl. their tests. The branch `feature-disabled/493-add-example-module` contains the first proposal. If I remember correct merging the configuration is left (incl. adding the removed classes etc. again).

The first increment should update the builder pattern without breaking existing modules. Legacy modules are kept as long as the new modules are deployed due to breaking changes / compatibility.

KijongHan:
> Thanks for your PR. Like mentioned in the [main contributing guidelines](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/docs/contributing.md) it is better to discuss changes with the authors first, especially large changes.
> 
> #493 is not just about moving the current implementation into dedicated packages. The builder pattern has major flaws in regard to creating module specific container instances. It is necessary to fix this design flaw first. Afterwards we can start to migrate module after module incl. their tests. The branch `feature-disabled/493-add-example-module` contains the first proposal. If I remember correct merging the configuration is left (incl. adding the removed classes etc. again).
> 
> The first increment should update the builder pattern without breaking existing modules. Legacy modules are kept as long as the new modules are deployed due to breaking changes / compatibility.

Might've jumped the gun a bit here :(. Yeah as you say looking through that example module branch it seems to be quite a big deviation away from the existing builder pattern and test container inheritance/implementation - Im happy to close this PR

HofmeisterAn:
I'll add more information to #493 about the necessary changes next week. I appreciate your efforts, but as you said the deviation is quite big. Nevertheless, feedback about the current proposal would be great. If you would like to contribute, it makes probably sense to continue to work on the proposal (with pleasure together).

PR #674, opened 2022-11-12
Minor fixes to cake scripts

KijongHan:
I was running into a few intellisense errors when developing locally for the following cake scripts: `version.cake`, `parameters.cake`

- Add package reference to local `Cake.Git` in dependent script to pick up intellisense for `ICakeContext:GitBranchCurrent`
- Update to use `DotNetVerbosity`

### DIFF
--- a/.cake-scripts/version.cake
+++ b/.cake-scripts/version.cake
@@ -1,3 +1,5 @@
+#addin nuget:file:../.cake-tools/Addins/?package=Cake.Git
+
 internal sealed class BuildInformation
 {
   private BuildInformation()
--- a/.cake-scripts/version.cake
+++ b/.cake-scripts/version.cake
@@ -1,4 +1,4 @@
-#addin nuget:file:../.cake-tools/Addins/?package=Cake.Git
+#addin nuget:?package=Cake.Git&version=2.0.0
 
 internal sealed class BuildInformation
 {
--- a/.cake-scripts/parameters.cake
+++ b/.cake-scripts/parameters.cake
@@ -23,7 +23,7 @@ internal sealed class BuildParameters
   public bool IsReleaseBuild { get; private set; }
   public bool IsPullRequest { get; private set; }
   public bool ShouldPublish { get; private set; }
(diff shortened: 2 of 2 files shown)


.cake-scripts/version.cake
```diff
@@ -1,3 +1,5 @@
+#addin nuget:file:../.cake-tools/Addins/?package=Cake.Git
```
HofmeisterAn: Thanks, LGTM. I am just curious why do we need this change? Is there any documentation? Best practice?

KijongHan: I found that without that `addin` directive, the line
``` csharp
context.GitBranchCurrent(".");
```
shows as an intellisense error for me using vs code with all the prerequisite extensions installed (I guess its because within the context of that `version.cake` file it doesn't have information on that referenced library as the `Cake.Git` addin is propagated down from `build.cake`) - i.e. this is just to improve the developer experience and silence "false alarm" intellisense errors

Cake script build practices were hard to come by unfortunately :( but the main reason I used the "local reference" format for the addin reference was to prevent potential errors resulting from having multiple different versions of the same package reference when a developer forgets to keep them consistent etc.

HofmeisterAn: Ok, thanks. I am not using IntelliSense for Cake, that is why I was not aware of the "error" message. I think that should be fixed by the plugin. We can keep it, but the CLI complains about the missing version:

    The 'addin' directive is attempting to install the 'Cake.Git' package without specifying a package version number

I think we should change it to `#addin nuget:file:../.cake-tools/Addins/?package=Cake.Git&version=2.0.0`?

KijongHan: Ah I see thanks for pointing that out. After some digging around, "file source" for this addin is probably not a good idea (as its main usage pattern is for "unpublished" local packages) and I came across a few unwanted results when attempting to upgrade the Cake.Git package - I updated this line to use the default addin syntax instead (with version specification as you suggested) and it is working as intended

HofmeisterAn (approved):

PR #675, opened 2022-11-14
Add ability to specify the binding port for Kafka

pjwilliams2:
Give the user the ability to specify a non-default port for the Kafka broker. This is ideal for when testing on a system that already has Kafka running on a docker container.


kiview:
Hey @pjwilliams2, can you please elaborate what you mean with "This is ideal for when testing on a system that already has Kafka running on a docker container."? It is not clear to me, especially given that we configure the ports dynamically at container startup.

pjwilliams2:
@kiview After re-evaluating the issues our developers were running into, this change is no longer needed and can be closed. 

There is an enhancement that might be helpful in general (outside of this change). When running a command within the container via `KafkaTestcontainer.ExecAsync` users have to know to use `localhost:9093` as the bootstrap server instead of using the alluring `BootstrapServers` property. Maybe it would be prudent to establish an `InternalBootstrapServers` property (or something equivalent) to abstract that away and signify inside vs. outside use. Just a thought.

Issue #678, opened 2022-11-16
Option to specify host for communication with ResourceReaper container


BenasB:
**Is your feature request related to a problem? Please describe.**
Right now (by default), when you start any test container, a resource reaper container starts automatically beforehand. The starting of this container takes the same [`IDockerEndpointAuthenticationConfiguration` as the container you're trying to start](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers/Containers/ResourceReaper.cs#L55) which specifies the endpoint/hostname, which _is fine_ in my scenario. Then, when actually [communicating/sending data over TCP](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers/Containers/ResourceReaper.cs#L264) to the resource reaper container, it takes [the same host name](https://github.com/testcontainers/testcontainers-dotnet/blob/edc5d0f992c8e688dd9a3d2d9d3460e40218fc86/src/Testcontainers/Containers/ResourceReaper.cs#L232) as in the container starting step, which _is not fine_ for my scenario as that endpoint is used only to create containers/communicate with the Docker API (I communicate with the actual containers in my tests through a different host name).

**Describe the solution you'd like**
It would be great if a user could specify the host name on which to communicate with the resource reaper container. I am just not sure how best to propose it.
- There's [`ResourceReaperPublicHostPort`](https://github.com/testcontainers/testcontainers-dotnet/blob/edc5d0f992c8e688dd9a3d2d9d3460e40218fc86/src/Testcontainers/Configurations/TestcontainersSettings.cs#L66) already, so there's possibility for a `ResourceReaperHost`, although I'm not the biggest fan of this static setup.
- With a `.With` builder config (e. g. `WithResourceReaperCommunicationHost`) (but that goes against the fact that the resource reaper is shared between the containers)

**Describe alternatives you've considered**
[...]

PR #695, opened 2022-11-28
feat(#678): Add support of custom configuration TESTCONTAINERS_HOST_OVERRIDE and TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE

HofmeisterAn:
## What does this PR do?

Adds the support of the custom configurations `TESTCONTAINERS_HOST_OVERRIDE` and `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE`. The host resolution will favor the host override configuration, where the Resource Reaper will favor to mount the socket override configuration as Docker daemon socket.

## Why is it important?

Some test (CI) environments require overriding the configurations mentioned above. This allows more developers to use Testcontainers for .NET.

## Related issues

- Relates #678

### DIFF
--- a/src/Testcontainers/Containers/ResourceReaper.cs
+++ b/src/Testcontainers/Containers/ResourceReaper.cs
@@ -9,6 +9,7 @@ namespace DotNet.Testcontainers.Containers
   using DotNet.Testcontainers.Builders;
   using DotNet.Testcontainers.Clients;
   using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Images;
   using JetBrains.Annotations;
 
   /// <summary>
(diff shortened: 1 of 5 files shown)


src/Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -61,7 +63,7 @@ private ResourceReaper(Guid sessionId, IDockerEndpointAuthenticationConfiguratio
         .WithCleanUp(false)
         .WithExposedPort(RyukPort)
         .WithPortBinding(TestcontainersSettings.ResourceReaperPublicHostPort.Invoke(dockerEndpointAuthConfig), RyukPort)
-        .WithBindMount("/var/run/docker.sock", "/var/run/docker.sock", AccessMode.ReadOnly)
+        .WithBindMount(dockerSocketFilePath, "/var/run/docker.sock", AccessMode.ReadOnly)
```
HofmeisterAn, line 66: @eddumelendez @kiview @mdelapenya Does the socket always bind to `/var/run/docker.sock` (destination)? Is that similar to your language implementation?

src/Testcontainers/Containers/TestcontainersContainer.cs
```diff
@@ -87,6 +87,11 @@ public string Hostname
     {
       get
       {
+        if (!string.IsNullOrEmpty(TestcontainersSettings.DockerHostOverride))
```
HofmeisterAn, line 90: @eddumelendez @kiview @mdelapenya How do you favour the host override configuration? Is that similar to your language implementation?

src/Testcontainers/Containers/ResourceReaper.cs
kiview: It is the same as in Java, see: https://github.com/testcontainers/testcontainers-java/blob/cae12726c04ca80a7049b49245b8d005ac38e95d/core/src/main/java/org/testcontainers/utility/RyukContainer.java#L23

src/Testcontainers/Containers/TestcontainersContainer.cs
kiview: In Java, the `DOCKER_HOST_OVERRIDE` is not accessed through the configuration subsystem, but directly read from the ENV, see:
https://github.com/testcontainers/testcontainers-java/blob/f5ec43dc358a01f8c82713ec7e66b2552e386515/core/src/main/java/org/testcontainers/dockerclient/DockerClientProviderStrategy.java#L411

I am not sure what is the historic context for this decision, here is the original PR for this:
https://github.com/testcontainers/testcontainers-java/pull/2991

Since we know that this is required as a persistent config change in certain environments as well (e.g., Colima), using the regular configuration subsystem seems correct to me.

src/Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: Thanks

PR #712, opened 2022-12-13
chore: Replace CHANGELOG.md with GitHub release notes

HofmeisterAn:
## What does this PR do?

Replaces the `CHANGELOG.md` with the GitHub release notes. Release Drafter keeps the GitHub release notes up-to-date. The commit message and description of each labeled pull request are appended to the GitHub release notes.

I would be interested in your thoughts @kiview @eddumelendez: GitHub's draft release notes are not visible public. Changes between the latest released version and head are documented, but other developers cannot see them. My idea is to publish the upcoming (unreleased) changes as pre-release. The information about pre-releases are public.

https://github.com/testcontainers/testcontainers-dotnet/releases will always show the upcoming (unreleased) changes first, where GitHub's main page links to: https://github.com/testcontainers/testcontainers-dotnet/releases/latest.

I see only one "disadvantage". Creating a new release cycle (e.g. the next version `2.5.0`) will immediately create a tag and send a notification about it.

## Why is it important?

Organizing the `CHANGELOG.md` manually is cumbersome and inaccurate

## Related issues

\-


eddumelendez:
I like the idea of making it public if possible. Another way to see the list of changes is by using the milestone. In tc-java, `next` is used until the release is done, and then it is updated to `1.18.0`, for example. 

> I see only one "disadvantage". Creating a new release cycle (e.g. the next version 2.5.0) will immediately create a tag and send a notification about it.

So, the release notes remain with the same version, right? IMO, creating tags every time a PR is merged is not something I would like to do.

For simplicity, I would prefer not to do pre-release.

HofmeisterAn:
> Another way to see the list of changes is by using the milestone. In tc-java, `next` is used until the release is done, and then it is updated to `1.18.0`, for example.

I did that in the past. Devs asked explicitly to add something like a CHANGELOG.

> IMO, creating tags every time a PR is merged is not something I would like to do.

It’s not every time. Just once when you start a new development cycle (version). Similar like you update once the next milestone in a release. It will update the pre-release notes incl. the ref of the tag within the same version.

CHANGELOG.md
```diff
@@ -1,91 +1,3 @@
...
-
-- 497 Remove `ResourceReaperDiagnostics`
-
-### Fixed
-
-- 431 Fix `System.InvalidOperationException : cannot hijack chunked or content length stream` (update `Docker.DotNet` dependency)
-- 449 Fix wrong mapped public host port
-- 484 Fix unit tests that fail on Windows hosts (@vlaskal)
-- 507 Fix `ITestcontainersConfiguration` duplication, such as `DockerApiException : [...] Duplicate mount point [...]` (@alesandrino)
+The [release notes](https://github.com/testcontainers/testcontainers-dotnet/releases) are created and updates from pull requests. The pre-release notes contain all upcoming changes.
```
kiview:
```suggestion
The [release notes](https://github.com/testcontainers/testcontainers-dotnet/releases) are created and updated automatically through merged pull requests. The pre-release notes of the next planned release contain all upcoming changes.
```

kiview:
I think it is a better and more automated process as compared to the `next` milestone in Java. Manually creating the pre-release after the release is very similar to what we do, when we recreate the `next` milestone after release.

We also have an old `CHANGELOG.md` file in Java, but it points only to our releases page. 

@eddumelendez could you give some more context on why you are concerned about the pre-release, or was it just a concern about tag-per-PR (which would not be the case, as explained by @HofmeisterAn)? 

I think this implementation is also informed by this GH discussion of someone asking, how to achieve a public draft release:
https://github.com/community/community/discussions/4192#discussion-3414257

kiview (dismissed):
Approved under the assumption we resolve the ongoing discussion.

eddumelendez (dismissed):

PR #724, opened 2022-12-23
chore: Replace SharpZipLib TarArchive helper class with TarOutputStream

HofmeisterAn:
## What does this PR do?

Replaces the TarArchive helper class with the TarOutputStream.

## Why is it important?

This PR allows us to read the input file before creating the tar entry. In some cases (e.g. the input file is used by another process), we cannot read the file contents. The SharpZipLib's `TarArchive.WriteEntry(TarEntry, bool)` helper method creates and adds the tar entry before reading the file. In the case mentioned above, this will throw a `TarException` that hides the underlying exception (issue).

For a better development experience, we replace the helper class / method. In case of an error, developers will get the underlying error.

## Related issues

\-

### DIFF
--- a/src/Testcontainers/Images/DockerfileArchive.cs
+++ b/src/Testcontainers/Images/DockerfileArchive.cs
@@ -4,6 +4,7 @@
   using System.Collections.Generic;
   using System.IO;
   using System.Linq;
+  using System.Text;
   using System.Text.RegularExpressions;
   using DotNet.Testcontainers.Configurations;
   using ICSharpCode.SharpZipLib.Tar;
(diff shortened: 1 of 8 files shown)


src/Testcontainers/Images/DockerfileArchive.cs
```diff
@@ -76,11 +77,11 @@ public string Tar()
 
       var dockerIgnoreFile = new DockerIgnoreFile(dockerfileDirectoryPath, ".dockerignore", dockerfileFilePath, this.logger);
 
-      using (var stream = new FileStream(dockerfileArchiveFilePath, FileMode.Create))
+      using (var tarOutputFileStream = new FileStream(dockerfileArchiveFilePath, FileMode.Create, FileAccess.Write))
```
HofmeisterAn, line 80: Probably we can change the method to `public Task<string> Tar()`.

HofmeisterAn:
Probably we can change the method to `public Task<string> Tar()`.

PR #727, opened 2023-01-04
chore: Replace host port 0 with an empty string

HofmeisterAn:
## What does this PR do?

Changes the default host port from `0` to an empty string. Considering https://github.com/docker/for-mac/issues/5588#issuecomment-934600089 the implementation of Docker Desktop behaves different. We see this in the chosen ports too. The port ranges are different. This is just the preparation. If there are no issues we can remove `ResourceReaperPublicHostPort` and the host port `0` configurations from the modules.

## Why is it important?

To align with the Java implementation. All TC developers should experience the same behavior.

## Related issues

- Relates https://github.com/testcontainers/testcontainers-java/pull/6383#pullrequestreview-1235448254

### DIFF
--- a/src/Testcontainers/Builders/TestcontainersBuilder.cs
+++ b/src/Testcontainers/Builders/TestcontainersBuilder.cs
@@ -186,6 +186,9 @@ namespace DotNet.Testcontainers.Builders
     /// <inheritdoc cref="ITestcontainersBuilder{TDockerContainer}" />
     public ITestcontainersBuilder<TDockerContainer> WithPortBinding(string hostPort, string containerPort)
     {
+      // Prepare Java alignment. Use an empty string instead of 0: https://github.com/docker/for-mac/issues/5588#issuecomment-934600089.
+      hostPort = "0".Equals(hostPort, StringComparison.OrdinalIgnoreCase) ? string.Empty : hostPort;
+
       var portBindings = new Dictionary<string, string> { { containerPort, hostPort } };
       return this.MergeNewConfiguration(new TestcontainersConfiguration(portBindings: portBindings))
         .WithExposedPort(containerPort);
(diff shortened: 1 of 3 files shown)


src/Testcontainers/Builders/TestcontainersBuilder.cs
```diff
@@ -186,6 +186,9 @@ public ITestcontainersBuilder<TDockerContainer> WithPortBinding(string port, boo
     /// <inheritdoc cref="ITestcontainersBuilder{TDockerContainer}" />
     public ITestcontainersBuilder<TDockerContainer> WithPortBinding(string hostPort, string containerPort)
     {
+      // Prepare Java alignment. Use an empty string instead of 0: https://github.com/docker/for-mac/issues/5588#issuecomment-934600089.
+      hostPort = "0".Equals(hostPort, StringComparison.OrdinalIgnoreCase) ? string.Empty : hostPort;
```
kiview, line 190: Is this the right part in the code to make this change? This would mean, if a user consciously sets it to `0` (not sure if there is a valid reason for this TBH), we would replace it as well.

I would have expected we change all occurrences in our code, where we construct a random mapping to use `""` instead of `0`. 

But since you mentioned this is a preparation, you maybe wanted to follow up on this after all, then it is of course a good intermediate step as a small change.

HofmeisterAn: > I would have expected we change all occurrences in our code, where we construct a random mapping to use `""` instead of `0`.

Yes, but this would require changes in the modules incl. their extension methods. This change allows us to test the behavior with the least amount of changes.

> But since you mentioned this is a preparation, you maybe wanted to follow up on this after all, then it is of course a good intermediate step as a small change.

Soon as we start to replace the modules, we can remove this line.

PR #729, opened 2023-01-10
refactor(#493): Replace module extension methods with module API

HofmeisterAn:
## What does this PR do?

This PR prepares a better Testcontainers for .NET module support. It makes the old approach (through extension methods) obsolete and removes the dependency to itself. It increases the isolation between other modules. Modules become independent. In the future, they can rely on other dependencies or modules and contain much more and complex custom implementations (e.g. nested configurations, mix of different containers, networks, volumes, etc.). Besides community maintained modules, even other OSS projects or vendors can provide and maintain their module implementation (single- or mono-repo).

1. The PR replaces the builders `TestcontainersBuilder`, `TestcontainersNetworkBuilder` and `TestcontainersVolumeBuilder` with `ContainerBuilder`, `NetworkBuilder` and `VolumeBuilder` incl. all the related interfaces and classes.
2. The design makes heavily use of inheritance. All resources (container, network, volume) share common design principles and behave similar to each other. Each level adds more functionalities to the builder and its equivalent resource. Among others, this covers default values and input validations on each level.
4. The builder state is immutable. Each configuration call generates a new immutable instance that holds the configuration.
5. All upcoming obsolete classes and interfaces moved to `_OBSOLETE_`. Soon as their equivalent  module implementation is ready, we can start to remove them from the repository.
6. `BackwardCompatibility` holds interfaces to allow developers to switch gracefully to the new classes and interfaces. In some rare cases, we cannot provide backwards compatibility. Usually, it is enough to rename the class or interface to its new name (see _Breaking Changes_).

## Why is it important?

To support a proper module implementation without depending on this repository.

## Related issues

- Relates #493

## Breaking Changes

- We tried to keep most changes backwards compatible. All upcoming breaking changes are marked as obsolete. Please consider migrating to the new classes and interfaces in advance. We will remove the obsolete classes and interfaces in the next release. In some edge cases, it might be necessary to switch to the new classes and interfaces immediately:

    | Old name                                                                  | New name           |
    |---------------------------------------------------------------------------|--------------------|
    | `ITestcontainersContainer`, `IDockerContainer`, `IRunningDockerContainer` | `IContainer`       |
    | `IDockerImage`                                                            | `IImage`           |
    | `IDockerNetwork`                                                          | `INetwork`         |
    | `IDockerVolume`                                                           | `IVolume`          |
    | `TestcontainersBuilder`                                                   | `ContainerBuilder` |
    | `TestcontainersNetworkBuilder`                                            | `NetworkBuilder`   |
    | `TestcontainersVolumeBuilder`                                             | `VolumeBuilder`    |
    | `TestcontainersContainer`                                                 | `DockerContainer`  |

- Please use **`ContainerBuilder`** instead of `TestcontainersBuilder<TestcontainersContainer>` or `ContainerBuilder<DockerContainer>` for generic (non module) configurations. Upcoming modules will use their own builder to support more complex configurations and use cases.
- `ImageFromDockerfileBuilder.Build()` does not return a `Task<string>` anymore. It returns a configured implementation of `IFutureDockerImage`. To finally build the image call `CreateAsync(CancellationToken)`.

## Out of scope

~The `ImageBuilder` is not ready yet. You do not need to review `ContainerBuilder<TContainerEntity>`. This builder is already obsolete and only necessary to provide backwards compatibility~.

## Follow-ups

- ~Support low level modifications for all builders (image, network, volume)~
- Extend the lifecycle callbacks
- Update documentation
- ~Add module scaffolding via `dotnet` CLI~
- Remove `IDockerNetwork.Id`
- Start the Resource Reaper already on image, network, volume `CreateAsync` (not just container start)
- Add the temporary removed warnings again.


src/Testcontainers/BackwardCompatibility/BackwardsCompatibility.cs
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  [PublicAPI]
+  [Obsolete("Use the VolumeBuilder class instead.")]
+  public sealed class TestcontainersVolumeBuilder : VolumeBuilder
+  {
+  }
+
+  [PublicAPI]
+  [Obsolete("Use the VolumeBuilder class instead.")]
```
HofmeisterAn:
```suggestion
  [Obsolete("Use the ContainerBuilder class instead.")]
```

src/Testcontainers/BackwardCompatibility/BackwardsCompatibility.cs
```diff
@@ -0,0 +1,151 @@
...
+  public sealed class TestcontainersContainer : DockerContainer
+  {
+    public TestcontainersContainer(IContainerConfiguration configuration, ILogger logger)
```
HofmeisterAn:
```suggestion
    internal TestcontainersContainer(IContainerConfiguration configuration, ILogger logger)
```

src/Testcontainers/Builders/NetworkBuilder.cs
```diff
@@ -0,0 +1,91 @@
...
+    public NetworkBuilder WithOption(string name, string value)
+    {
+      var options = new Dictionary<string, string> { { name, value } };
+      return this.Merge(this.DockerResourceConfiguration, new NetworkConfiguration(options: options));
+    }
+
+    /// <inheritdoc />
+    public override INetwork Build()
+    {
+      this.Validate();
```
HofmeisterAn, line 70: Override `Validate()` and guard `Name`.

src/Testcontainers/Builders/VolumeBuilder.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <inheritdoc />
+    public VolumeBuilder WithName(string name)
+    {
+      return this.Merge(this.DockerResourceConfiguration, new VolumeConfiguration(name: name));
+    }
+
+    /// <inheritdoc />
+    public override IVolume Build()
+    {
+      this.Validate();
```
HofmeisterAn, line 55: Override `Validate()` and guard `Name`.

src/Testcontainers/BackwardCompatibility/BackwardsCompatibility.cs
```diff
@@ -0,0 +1,197 @@
+﻿#pragma warning disable SA1403
+
+namespace DotNet.Testcontainers
```
kiview, line 3: Asking as a Java guy here, this namespace for this API means we will avoid breaking changes for many users of the old API, right? They don't have to change any of their imports to make use of the backward compatibility API?

src/Testcontainers/Builders/AbstractBuilder`2.cs
```diff
@@ -0,0 +1,137 @@
...
+      this.DockerResourceConfiguration = dockerResourceConfiguration;
+    }
+
+    /// <summary>
+    /// Gets the Docker resource configuration.
+    /// </summary>
+    protected virtual TConfigurationEntity DockerResourceConfiguration { get; }
+
+    /// <inheritdoc cref="IAbstractBuilder{TBuilderEntity, TContainerEntity}" />
+    public TBuilderEntity WithDockerEndpoint(string endpoint)
```
kiview, line 37: Do we need another abstraction here, rather than specifying the Docker endpoint as a string? Such an abstraction would be e.g. `DockerClient`. 

`IDockerEndpointAuthenticationConfiguration` seems to represent such an abstraction. 

We can discuss it, since it can influence how the API is perceived and used, but also how things can be implemented internally. Does using a client abstraction as an object that can be reused make things easier, or is it easier if the implementation internally takes care of reusing a client if the same endpoint configuration is given? 

And for which users is this API relevant in the first place?

src/Testcontainers/Builders/AbstractBuilder`2.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// <inheritdoc cref="IAbstractBuilder{TBuilderEntity, TContainerEntity}" />
+    public TBuilderEntity WithDockerEndpoint(IDockerEndpointAuthenticationConfiguration dockerEndpointAuthConfig)
+    {
+      return this.Clone(new ResourceConfiguration(dockerEndpointAuthenticationConfiguration: dockerEndpointAuthConfig));
+    }
+
+    /// <inheritdoc cref="IAbstractBuilder{TBuilderEntity, TContainerEntity}" />
+    public TBuilderEntity WithCleanUp(bool cleanUp)
+    {
+      return this.WithResourceReaperSessionId(TestcontainersSettings.ResourceReaperEnabled && cleanUp ? ResourceReaper.DefaultSessionId : Guid.Empty);
```
kiview, line 57: Similar to my comment about the endpoint, here I am thinking whether a dedicated abstraction as an object is better. Thinking about this specifically, since in Java we provide the default container based Ryuk, but also added a JVM based Ryuk as a fallback for e.g. Podman users.

This topic is further complicated, since there is a direct relationship between the Docker client/endpoint and the available resource reapers.

src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
```diff
@@ -1,116 +1,126 @@
...
     {
+      this.DockerResourceConfiguration = dockerResourceConfiguration;
     }
 
     /// <inheritdoc />
-    public IImageFromDockerfileBuilder WithName(string name)
+    protected override IImageFromDockerfileConfiguration DockerResourceConfiguration { get; }
+
+    /// <inheritdoc />
+    public ImageFromDockerfileBuilder WithName(string name)
```
kiview, line 53: Is this `name` + `tag`? Can I build images without specifying it?

src/Testcontainers/Clients/ContainerConfigurationConverter.cs
```diff
@@ -8,15 +8,15 @@ namespace DotNet.Testcontainers.Clients
   using DotNet.Testcontainers.Networks;
   using JetBrains.Annotations;
 
-  internal sealed class TestcontainersConfigurationConverter
+  internal sealed class ContainerConfigurationConverter
```
kiview, line 11: What is a `ConfigurationConverter`?

src/Testcontainers/Configurations/Commons/ResourceConfiguration.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <param name="newValue">The new Docker resource configuration.</param>
+    protected ResourceConfiguration(IResourceConfiguration oldValue, IResourceConfiguration newValue)
+      : this(
+        dockerEndpointAuthenticationConfiguration: BuildConfiguration.Combine(oldValue.DockerEndpointAuthConfig, newValue.DockerEndpointAuthConfig),
+        labels: BuildConfiguration.Combine(oldValue.Labels, newValue.Labels))
+    {
+    }
+
+    /// <inheritdoc />
+    public Guid SessionId { get; }
```
kiview, line 49: What is a `SessionId` here? The Resource reaper one, right?

src/Testcontainers/BackwardCompatibility/BackwardsCompatibility.cs
HofmeisterAn: Yes, exactly. The class just helps us to remove the obsolete classes and interfaces with the next release (2.5.0) after this release (2.4.0). Users won't notice a difference except a warning that some classes and interfaces are obsolete.

src/Testcontainers/Builders/AbstractBuilder`2.cs
HofmeisterAn: > `IDockerEndpointAuthenticationConfiguration` seems to represent such an abstraction.

The overloaded method `WithDockerEndpoint(IDockerEndpointAuthenticationConfiguration)` allows developers to pass a "custom" Docker client implementation to Testcontainers for .NET.

> or is it easier if the implementation internally takes care of reusing a client if the same endpoint configuration is given?

.NET caches the client object internal (it is not created immediately). The [auto discovery value](https://github.com/testcontainers/testcontainers-dotnet/blob/c126fb91f7e7872a72b9086f0dcca76b03b54507/src/Testcontainers/Builders/AbstractBuilder%602.cs#L109) is set as default endpoint for each configuration.

> And for which users is this API relevant in the first place?

I know developers are using it to configure the HTTP client.

src/Testcontainers/Builders/AbstractBuilder`2.cs
HofmeisterAn: A fallback in .NET through runtime hooks or something like that is super unreliable. We had that in the past; lots of leftovers. I am very happy about the Ruyk support. Developers can configure each resource individual if necessary. Something like an automatic deletion implementation for images is not necessary. Each resource behaves the same.

OC, an abstraction might still be useful for other use cases or technologies in the future. Something we might wanna discuss (not sure, if I understood everything correct).

src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
HofmeisterAn: The `DockerImage` implementation appends `latest` if the developer passes a string without tag.

src/Testcontainers/Clients/ContainerConfigurationConverter.cs
HofmeisterAn: It maps (converts) TC for .NET types to Docker.DotNet types. There is not always a 1:1 mapping.

src/Testcontainers/Configurations/Commons/ResourceConfiguration.cs
HofmeisterAn: Yes, it is basically the Resource Reaper id which is equivalent to `x-tc-sid`.

src/Testcontainers/Builders/AbstractBuilder`2.cs
kiview: We will resolve this comment by extending the method documentation to bring an emphasis on this being a method not normally called by users (since they should rely on autoconfig).

src/Testcontainers/Builders/AbstractBuilder`2.cs
kiview: We will make `WithResourceReaperSessionId` an internal method, since we are not aware of user use cases for this.

src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
kiview: We will create an issue for this where we will describe how to align the properties of `IImage` with the Docker domain language (regarding `registry`, `repository` and `tag`). 

Docker is a bit messy here themselves, but we just try our best to hit it eventually.

kiview:
Regarding default image names, that's the format how we construct it in Java:
```
"localhost/testcontainers/" + Base58.randomString(16).toLowerCase()
```

Having this aligned might make it look better for users inspecting images, that use multiple Testcontainers implementations.

src/Testcontainers/Builders/AbstractBuilder`2.cs
HofmeisterAn: I've added

> Testcontainers automatically discovers the Docker environment and applies the configuration.
> It is recommended to use the auto discovery. Only certain edge cases require this enhanced API.

to the XML (code) documentation.

src/Testcontainers/Builders/AbstractBuilder`2.cs
HofmeisterAn: `WithResourceReaperSessionId` is not public anymore (protected now).

HofmeisterAn:
> Regarding default image names, that's the format how we construct it in Java:
> 
> ```
> "localhost/testcontainers/" + Base58.randomString(16).toLowerCase()
> ```
> 
> Having this aligned might make it look better for users inspecting images, that use multiple Testcontainers implementations.

The network and volume builder set a default name (Guid). The image builder changed the default image name to something like `localhost/testcontainers/11c43ee8-b9d3-4e51-b73f-bd9dda66e29c:latest`.

src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
HofmeisterAn: https://github.com/testcontainers/testcontainers-dotnet/issues/740

PR #730, opened 2023-01-10
Add EventStore Testcontainer

meysamhadeli:
Add PR for feature EventStore to Testcontainer: #730

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -33,5 +33,7 @@
     <PackageReference Update="RabbitMQ.Client" Version="6.2.4" />
     <PackageReference Update="StackExchange.Redis" Version="2.6.48" />
     <PackageReference Update="System.Data.SqlClient" Version="4.8.3" />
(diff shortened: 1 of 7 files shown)


HofmeisterAn (commented):
The module configuration looks good, thanks. Unfortunately, it follows the old approach. I will finalize the new module API next week. Do you mind if we wait until #729 is done? Then we do not have to migrate afterwards again.

meysamhadeli:
> The module configuration looks good, thanks. Unfortunately, it follows the old approach. I will finalize the new module API next week. Do you mind if we wait until #729 is done? Then we do not have to migrate afterwards again.

It's ok. You can add this module to old version too and also after #729 is done I can add it again to new version. Because the structure was unknown for the new version and there was no test, I implemented it on the previous structure.

meysamhadeli:
I try to add this feature with new style of project.

PR #735, opened 2023-01-16
feat: Add GitHub Codespaces configuration (Dev Container)

HofmeisterAn:
## What does this PR do?

This PR contains three major changes. It updates the NuGet dependencies (1) and removes the global output directory (2). Unfortunately, the latest versions of the .NET SDK break more and more things in regards of the global output directory. Considering the GitHub issues it does not look like MS will add the compatibility again (it has worked since the .NET Framework).

Furthermore, it adds a Dev Container configuration to use Testcontainers with GitHub Codespaces (3)

## Why is it important?

Simplifies the .NET build and adds GitHub Codespaces support.

## Related issues

- https://github.com/dotnet/sdk/issues/19534
- https://github.com/dotnet/sdk/issues/17645
- https://stackoverflow.com/q/67416581/690017


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,20 @@
...
+  ],
+  "extensions": [
+    "formulahendry.dotnet-test-explorer",
+    "ms-azuretools.vscode-docker",
+    "ms-dotnettools.csharp"
+  ],
+  "settings": {
+    "omnisharp.path": "latest" // https://github.com/OmniSharp/omnisharp-vscode/issues/5410#issuecomment-1284531542.
+  },
+  "postCreateCommand": "git config --system --add safe.directory ${containerWorkspaceFolder} && nohup /bin/sh -c 'dockerd &'",
```
kiview: This is a very unusual approach to give Docker support, you normally enable it through devcontainers feature definition:
https://github.com/testcontainers/workshop/blob/main/.devcontainer/devcontainer.json#L12

kiview (dismissed):
We can probably start like this, but I would suggest exploring Devcontainers Features to make the config a bit simpler (e.g. regarding Docker):
https://code.visualstudio.com/blogs/2022/09/15/dev-container-features

Can be done as a follow-up, to just start from a working state, so in favor of merging like this 👍

eddumelendez:
The same can be applied to git, gh https://github.com/testcontainers/testcontainers-java/pull/6412/files#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686R37-R39

eddumelendez:
Looks like dotnet can be also added as a feature `"ghcr.io/devcontainers/features/dotnet:1": {}`

HofmeisterAn (commented):
Thanks for the feedback. I replaced the Dockerfile with the Dev Container features.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,35 @@
+{
+  "image": "ubuntu:22.04",
```
eddumelendez: I think we should use the ones provided by microsoft?  I generated the devcontainer via vscode and it uses the following image `"mcr.microsoft.com/devcontainers/base:jammy"`

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,35 @@
...
+    "DOTNET_CLI_TELEMETRY_OPTOUT": "true",
+    "DOTNET_NOLOGO": "true",
+    "DOTNET_SKIP_FIRST_TIME_EXPERIENCE": "true",
+    "DOTNET_USE_POLLING_FILE_WATCHER": "true",
+    "NUGET_XMLDOC_MODE": "skip"
+  },
+  "forwardPorts": [
+    2375,
+    2376
+  ],
```
eddumelendez: do we need this?

HofmeisterAn: No, not really. I added it to get rid of the pop up message.

.devcontainer/devcontainer.json
HofmeisterAn: I changed the image.

.devcontainer/devcontainer.json
HofmeisterAn: I removed `forwardPorts`.

Issue #736, opened 2023-01-16
[Bug]: Cannot connect a Schema Registry running from another container due to missing kafka broker alias


leonasa:
### Testcontainers version

2.3.0

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host arch

x64

### .NET version

6.0.x

### Docker version

```shell
Client: Docker Engine - Community
 Cloud integration: v1.0.29
 Version:           20.10.21
 API version:       1.41
 Go version:        go1.18.7
 Git commit:        baeda1f
 Built:             Tue Oct 25 18:02:28 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Desktop
 Engine:
  Version:          20.10.21
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.7
  Git commit:       3056208
  Built:            Tue Oct 25 18:00:19 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.10
  GitCommit:        770bd0108c32f3fb5c73ae1264f7e503fe7b2661
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.13.0)
  dev: Docker Dev Environments (Docker Inc., v0.0.5)
  extension: Manages Docker extensions (Docker Inc., v0.2.16)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.22.0)

Server:
 Containers: 4
  Running: 4
  Paused: 0
  Stopped: 0
 Images: 33
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
[...]

PR #737, opened 2023-01-16
[Bug]: #736 Cannot connect a Schema Registry running from another containe…

leonasa:
## What does this PR do?
Fix the problem when connecting a schema registry to the Kafka broker running in test containers. Bug #736 

## Why is it important?
Without the fix, a Kafka broker launched from a Testcontainer cannot be accessible inside the docker network.
That being said, we cannot run a schema registry in a test container and connect to the kafka broker.

## Related issues
- Closes #737

-

### DIFF
--- a/src/Testcontainers/Builders/TestcontainersBuilderKafkaExtension.cs
+++ b/src/Testcontainers/Builders/TestcontainersBuilderKafkaExtension.cs
@@ -19,6 +19,7 @@ namespace DotNet.Testcontainers.Builders
       return builder
         .WithImage(configuration.Image)
         .WithCommand(configuration.Command)
+        .WithNetworkAliases(configuration.KafkaBrokerAlias)
         .WithPortBinding(configuration.Port, configuration.DefaultPort)
         .WithWaitStrategy(configuration.WaitStrategy)
         .WithStartupCallback(configuration.StartupCallback)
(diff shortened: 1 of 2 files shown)


HofmeisterAn (commented):
Thanks for your contribution. If I understood your suggestion correct, the idea is to assign a network alias that allows other containers to access the Kafka instance. Developers still need to assign the containers to the same network, right?

#729 will enhance the module API a lot. It allows to combine different Docker resources together, like multiple contains, networks, volumes, etc. The extension method is not really capable of that.

You figured out a work around in regards of #736, right? Do you mind waiting until the new module API is available? Otherwise, we need to look into the failing test (this is [Java's](https://github.com/testcontainers/testcontainers-java/blob/111d519e303d3659e2f94c025ca5baa14363870a/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L11) configuration).

leonasa:
Hello. Thank you for the quick response.
Yes, that is what I have done. With a particularity for Kafka where we need
to use the network alias in the config KAFKA_ADVERTISED_LISTENERS before
the container starts.
But I do have a workaround creating a new KafkaTestcontainerConfiguration.
As you can see here:
KafkaTestContainersTests/KafkaTestContainerTest.cs at main ·
leonasa/KafkaTestContainersTests (github.com)
<https://github.com/leonasa/KafkaTestContainersTests/blob/main/KafkaTestContainersTests/KafkaTestContainerTest.cs>

HofmeisterAn:
Since we have migrated to the new module API, this pull request is no longer up-to-date. As a result of the issue #845, I have already updated the startup callback configuration by modifying the `KAFKA_ADVERTISED_LISTENERS` environment variable. I am uncertain whether the fix resolves your issue as well. I initially attempted to use the network alias, which did not work properly in my setup. I am unsure of the correct configuration to support all of the various use cases. Perhaps we can continue the discussion in the corresponding issue #736 and then create a follow-up pull request. Meanwhile, I will close this issue.

PR #756, opened 2023-02-02
docs: Replace obsolete builders

HofmeisterAn:
## What does this PR do?

Removes the obsolete classes and interfaces in the documentation. For the module builder, we need to stick to the obsolete builder until the new modules are finished. Due to the old generic and module builder uses the same syntax, we cannot distinguish between them.

## Why is it important?

To keep the documentation up-to-date and not rely on obsolete classes or interfaces.

## Related issues

\-

### DIFF
--- a/docs/modules/index.md
+++ b/docs/modules/index.md
@@ -4,7 +4,7 @@ Modules are a great example of the capabilities of Testcontainers for .NET. Choo
 
 !!!warning
 
-    We are redesigning modules and removing the extension method in the future. Modules will become independent projects that allow more complex and advanced features. Due to an old design flaw, we cannot distinguish between a generic and module builder. If you rely on a module you will get an obsolete warning until the next version of Testcontainers gets released. You will find more information [here](https://github.com/testcontainers/testcontainers-dotnet/issues/750#issuecomment-1412257694).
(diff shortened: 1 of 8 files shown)


docs/modules/index.md
```diff
@@ -2,6 +2,10 @@
 
 Modules are a great example of the capabilities of Testcontainers for .NET. Choose one of the pre-configurations below or find further examples in [TestcontainersContainerTest][testcontainers-container-tests] as well as in the [database][testcontainers-database-tests] or [message broker][testcontainers-message-broker-tests] tests to set up your test environment.
 
+!!!warning
+
+    We are redesigning modules and removing the extension method in the future. Modules will become independent projects that allow more complex and advanced features. Due to an old design flaw, we cannot distinguish between a generic and module builder. If you rely on a module you will get an obsolete warning until the next version of Testcontainers gets released. You will find more information [here](https://github.com/testcontainers/testcontainers-dotnet/issues/750#issuecomment-1412257694).
```
HofmeisterAn:
```suggestion
    We are redesigning modules and removing the extension method in the future. Modules will become independent projects that allow more complex and advanced features. Due to a design flaw in the current module system, we cannot distinguish between a generic and module builder. If you rely on a module you will get an obsolete warning until the next version of Testcontainers gets released. You will find more information [here](https://github.com/testcontainers/testcontainers-dotnet/issues/750#issuecomment-1412257694).
```

PR #757, opened 2023-02-03
feat: Replace Microsoft SQL Server module

HofmeisterAn:
## What does this PR do?

Replaces the old Microsoft SQL Server module approach (using an extension method) with a proper independent module.

## Why is it important?

We do not follow the old module approach anymore. We will replace the old approach with a new API and release each module as independent dependency. This module replaces Microsoft SQL Server.

## Related issues

- Relates #493


src/Testcontainers.MsSql/MsSqlBuilder.cs
```diff
@@ -0,0 +1,129 @@
+namespace Testcontainers.MsSql;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class MsSqlBuilder : ContainerBuilder<MsSqlBuilder, MsSqlContainer, MsSqlConfiguration>
```
HofmeisterAn, line 5: @eddumelendez @kiview @mdelapenya this is basically a complete module.

PR #773, opened 2023-02-10
feat: Add LocalStack module

dominikus1993:
What does this PR do?
Add support for amazon LocalStack testcontainer module.

Why is it important?
I think that adding this testcontainer will be useful for developers who want to test the operation of multiple aws modules

Related issues
Any of issues are no releated

How to test this PR
Running the new integration test will suffice.


src/Testcontainers.LocalStack/AwsService.cs
```diff
@@ -0,0 +1,132 @@
+﻿namespace Testcontainers.Minio;
```
HofmeisterAn:
```suggestion
﻿namespace Testcontainers.LocalStack;
```

It looks like the namespace is wrong in all files. Please move the usings to global usings too.

src/Testcontainers.LocalStack/AwsService.cs
```diff
@@ -0,0 +1,132 @@
...
+
+/// <summary>
+/// AwsService
```
HofmeisterAn:
```suggestion
/// An AWS service.
```

src/Testcontainers.LocalStack/AwsService.cs
```diff
@@ -0,0 +1,132 @@
...
+[PublicAPI]
+public sealed class AwsService
+{
+    /// <summary>
+    /// The name of the AWS service.
+    /// </summary>
+    public string Name { get; }
+
+    /// <summary>
+    /// Initializes a new instance of the `AwsService` class.
```
HofmeisterAn:
```suggestion
    /// Initializes a new instance of the <see cref="AwsService" /> class.
```

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /// <inheritdoc />
+    protected override LocalStackConfiguration DockerResourceConfiguration { get; }
+    
+    /// <summary>
+    /// Sets the LocalStack Default Ports Start.
+    /// </summary>
+    /// <param name="port">The LocalStack Default Ports Start.</param>
+    /// <returns>A configured instance of <see cref="LocalStackBuilder" />.</returns>
```
HofmeisterAn:
```suggestion
    /// <summary>
    /// Sets the LocalStack external service start port.
    /// </summary>
    /// <param name="port">The LocalStack external service start port.</param>
    /// <returns>A configured instance of <see cref="LocalStackBuilder" />.</returns>
```

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        return Merge(DockerResourceConfiguration, new LocalStackConfiguration(externalServicePortStart: port))
+            .WithEnvironment("EXTERNAL_SERVICE_PORTS_START", port);
+    }
+    
+    /// <summary>
+    /// Sets the LocalStack Default Ports End.
+    /// </summary>
+    /// <param name="port">The LocalStack Default Ports End.</param>
+    /// <returns>A configured instance of <see cref="LocalStackBuilder" />.</returns>
```
HofmeisterAn:
```suggestion
    /// <summary>
    /// Sets the LocalStack external service end port.
    /// </summary>
    /// <param name="port">The LocalStack external service end port.</param>
    /// <returns>A configured instance of <see cref="LocalStackBuilder" />.</returns>
```

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        return Merge(DockerResourceConfiguration, new LocalStackConfiguration(externalServicePortEnd: port))
+            .WithEnvironment("EXTERNAL_SERVICE_PORTS_END", port);
+    }
+    
+    /// <summary>
+    /// Sets the LocalStack Services.
+    /// </summary>
+    /// <param name="services">The LocalStack services.</param>
+    /// <returns>A configured instance of <see cref="LocalStackBuilder" />.</returns>
```
HofmeisterAn:
```suggestion
    /// <summary>
    /// Sets the LocalStack services.
    /// </summary>
    /// <param name="services">A list of LocalStack services.</param>
    /// <returns>A configured instance of <see cref="LocalStackBuilder" />.</returns>
```

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    public LocalStackBuilder WithServices(params AwsService[] services)
+    {
+        if (services?.Length is null)
+        {
+            return this;
+        }
+
+        string servicesNames = string.Join(",", services.Select(service => service.Name));
+        return Merge(DockerResourceConfiguration, new LocalStackConfiguration(services: services))
+            .WithEnvironment("SERVICES", servicesNames);
```
HofmeisterAn:
```suggestion
        var localStackBuilder = Merge(DockerResourceConfiguration, new LocalStackConfiguration(services: services));
        return localStackBuilder.WithEnvironment("SERVICES", string.Join(",", localStackBuilder.DockerResourceConfiguration.Services.Select(service => service.Name)));
```

This one is a bit tricky. Your `SERVICES` environment variable and `DockerResourceConfiguration. Services` property diverges. The environment variable will contain the services from the `services` argument, but the property will contain all configured services. Imagine someone configures:

```csharp
_ = new LocalStackBuilder()
    .WithServices(AwsService.S3)
    .WithServices(AwsService.Route53)
    .Build();
```

Either the environment variable contains all services too, or we set the property to the latest configuration.

I removed the `null` check. IMO it should fail fast if someone passes `null` to the member. WDYT about the suggestion?

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+
```
HofmeisterAn:
```suggestion

```

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    protected override LocalStackBuilder Init()
+    {
+        return base.Init()
+            .WithImage(LocalStackImage)
+            .WithExternalServicePortStart("4510")
+            .WithExternalServicePortEnd("4559")
+            .WithServices()
+            .WithPortBinding(LocalStackPort, true)
+            .WithWaitStrategy(Wait.ForUnixContainer().AddCustomWaitStrategy(new UntilReady()));
```
HofmeisterAn:
```suggestion
        return base.Init()
            .WithImage(LocalStackImage)
            .WithPortBinding(LocalStackPort, true)
            .WithExternalServicePortStart("4510")
            .WithExternalServicePortEnd("4559")
            .WithServices()
            .WithWaitStrategy(Wait.ForUnixContainer().AddCustomWaitStrategy(new UntilReady()));
```

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    
+    private sealed class UntilReady : IWaitUntil
+    {
+        public async Task<bool> UntilAsync(IContainer container)
+        {
+            var (stdout, _) = await container.GetLogs()
+                .ConfigureAwait(false);
+            return stdout != null && stdout.Contains("Ready.\n");
+        }
+    }
```
HofmeisterAn:
```suggestion
    /// <inheritdoc cref="IWaitUntil" />
    private sealed class UntilReady : IWaitUntil
    {
        /// <inheritdoc />
        public async Task<bool> UntilAsync(IContainer container)
        {
            var (stdout, _) = await container.GetLogs()
                .ConfigureAwait(false);
            return stdout != null && stdout.Contains("Ready.\n");
        }
    }
```

src/Testcontainers.LocalStack/LocalStackContainer.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Gets the AWS access key id.
+    /// </summary>
+    /// <returns>The AWS access key id.</returns>
+    public string GetAccessKeyId() => "dummy";
+
+    /// <summary>
+    /// Gets the AWS access secret.
+    /// </summary>
+    /// <returns>The AWS access secret.</returns>
+    public string GetAccessSecret() => "dummy";
```
HofmeisterAn: We do not really need `GetAccessKeyId()` and `GetAccessSecret()` right? What do you think if we use something like this in `LocalStackContainerTest`:

https://github.com/testcontainers/testcontainers-dotnet/blob/9f86d9f273e460b909c76df35177259fa674fd42/tests/Testcontainers.DynamoDb.Tests/DynamoDbContainerTest.cs#L7-L11

tests/Testcontainers.LocalStack.Tests/LocalStackContainerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+using Amazon.SimpleNotificationService.Model;
+using Amazon.SQS;
+using Amazon.SQS.Model;
+
+namespace Testcontainers.Minio;
+
+public sealed class LocalStackContainerTest : IAsyncLifetime
+{
+    private readonly LocalStackContainer _localStackContainer = new LocalStackBuilder()
+        .WithServices(AwsService.S3, AwsService.DynamoDb, AwsService.CloudWatchLogs, AwsService.Sqs, AwsService.Sns)
```
HofmeisterAn: xUnit.net will spin up a container for each `[Fact]`. Is it faster (better) to spin up services individually? E.g. the S3 test does not require DynamoDb, or doesn't that matter because it is quick anyway?

HofmeisterAn (commented):
Thanks again for your contribution! I have just one comment regarding `WithServices(params AwsService[])` and a question about the tests (I am not done with the review of the tests yet). Anything else is just cosmetic.

I will talk to the other maintainers next week to align the LocalStack module with the other Testcontainers language implementations. I know there is a WIP in Go too.

tests/Testcontainers.LocalStack.Tests/LocalStackContainerTests.cs
dominikus1993: In newer local stack it doesn't matter because all services are enabled by default
My mistake was that i thought it works the same as in 2019

src/Testcontainers.LocalStack/AwsService.cs
dominikus1993: fact, fixed it

src/Testcontainers.LocalStack/AwsService.cs
dominikus1993: I just removed whole aws service class because in newer local stack version all services are enabled by default

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: fixed

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: fixed

src/Testcontainers.LocalStack/AwsService.cs
dominikus1993: fixed it by deleting whole class

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: fixed

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: fixed

src/Testcontainers.LocalStack/LocalStackContainer.cs
dominikus1993: Your solution is much nicer, I fixed the code

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: I just removed whole class and used better wait until method

dgenezini:
The new health endpoint for localstack container is `/_localstack/health`.

Using the `/health` endpoint of localstack gives the following warning in the console:

```
WARN --- [   asgi_gw_0] localstack.deprecations    : /health is deprecated (since 1.3.0) and will be removed in upcoming releases of LocalStack! Use /_localstack/health instead.
```

dominikus1993:
thanks, I didn't noticed I
fixed

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    protected override LocalStackBuilder Init()
+    {
+        return base.Init()
+            .WithImage(LocalStackImage)
+            .WithPortBinding(LocalStackPort, true)
+            .WithExternalServicePortStart("4510")
+            .WithExternalServicePortEnd("4559")
+            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request => request.ForPath("/_localstack/health").ForPort(LocalStackPort)));
```
HofmeisterAn, line 45: I am not very familiar with LocalStack, how likely developers will use an older version (overrides the image) that does not use the health check endpoint? Does it even make sense to use and older version?

mdelapenya: In Java, the wait strategy waits for a log entry. See https://github.com/testcontainers/testcontainers-java/blob/343c5a533f4c6148d6e6eea3a3374116365bc91d/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java#L98, we adopted it for the Go implementation (https://github.com/testcontainers/testcontainers-go/pull/800/files#diff-bd71ec18d411b5b0b51723761d6d0a95e60ad7d5578828df544e14d6518a222dR53) and suggested it for Rust https://github.com/testcontainers/testcontainers-rs/pull/462#discussion_r1102602196

mdelapenya: To be fair, I do prefer having the wait for an HTTP endpoint, as it defines a more strict contract than waiting for a log entry. Probably something to discuss with the other languages cc/ @eddumelendez @kiview

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,102 @@
...
+    
+    /// <summary>
+    /// Sets the LocalStack external service start port.
+    /// </summary>
+    /// <param name="port">The LocalStack external service start port.</param>
+    /// <returns>A configured instance of <see cref="LocalStackBuilder" />.</returns>
+    public LocalStackBuilder WithExternalServicePortStart(string port)
+    {
+        return Merge(DockerResourceConfiguration, new LocalStackConfiguration(externalServicePortStart: port))
+            .WithEnvironment("EXTERNAL_SERVICE_PORTS_START", port);
```
HofmeisterAn: I do not understand when this environment variable is necessary. The configuration does not seem to bind the ports. What exactly connects to the container? Is it container-to-container communication?

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dgenezini: > I am not very familiar with LocalStack, how likely developers will use an older version (overrides the image) that does not use the health check endpoint? Does it even make sense to use and older version?

It may make sense to use an older version if some feature gets removed or when trying to find a bug.

Wait for a log entry may be an option, but I would benchmark both the health check and log entry strategies and go with the fastest. I find the speed more important than the specific scenario of someone overriding to an older version.

Other option would be to create both strategies and a specific method to override the default wait strategy. SOmething like this:

```
.WithWaitStrategy(LocalStackWaitStrategy.WaitForHealthCheck)
```

dominikus1993: @mdelapenya thanks for your tip. i just reverted my current changes to version with waiting for a log entry

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: I unnecessarily followed the previous implementation, it seems that it is not needed because these ports are set by default.

src/Testcontainers.LocalStack/LocalStackBuilder.cs
HofmeisterAn: I think we favor the health check 😬. We are just not sure if it makes sense to use an older version of LocalStack. I would imagine not, who wants to test against an older version of a cloud product? I am 👍 for the health check.

dominikus1993: ok no problem i just revert my previous changes :)

src/Testcontainers.LocalStack/LocalStackBuilder.cs
HofmeisterAn: Ok, was probably already unnecessary at that time. I think the other language implementations don't use it either.

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,66 @@
+using System.Threading.Tasks;
```
HofmeisterAn: Move to global using. Clean up usings.

src/Testcontainers.LocalStack/LocalStackBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+    protected override LocalStackConfiguration DockerResourceConfiguration { get; }
+
+    /// <inheritdoc />
+    public override LocalStackContainer Build()
+    {
+        Validate();
+        return new LocalStackContainer(DockerResourceConfiguration, TestcontainersSettings.Logger);
+    }
+
+    protected override LocalStackBuilder Init()
```
HofmeisterAn:
```suggestion

    /// <inheritdoc />
    protected override LocalStackBuilder Init()
```

src/Testcontainers.LocalStack/LocalStackContainer.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <param name="configuration">The container configuration.</param>
+    /// <param name="logger">The logger.</param>
+    public LocalStackContainer(LocalStackConfiguration configuration, ILogger logger)
+        : base(configuration, logger)
+    {
+        _configuration = configuration;
+    }
+
+    /// <summary>
+    /// Gets the Minio endpoint.
```
HofmeisterAn:
```suggestion
    /// Gets the LocalStack endpoint.
```

tests/Testcontainers.LocalStack.Tests/LocalStackContainerTests.cs
```diff
@@ -0,0 +1,143 @@
+﻿using System.Collections.Generic;
+using Amazon.CloudWatchLogs;
+using Amazon.CloudWatchLogs.Model;
+using Amazon.DynamoDBv2;
+using Amazon.DynamoDBv2.Model;
+using Amazon.Runtime;
+using Amazon.SimpleNotificationService;
+using Amazon.SimpleNotificationService.Model;
+using Amazon.SQS;
+using Amazon.SQS.Model;
```
HofmeisterAn: Move to global using. Clean up usings.

tests/Testcontainers.LocalStack.Tests/Testcontainers.LocalStack.Tests.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <PackageReference Include="coverlet.msbuild" Version="3.2.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5" />
+    <PackageReference Include="xunit" Version="2.4.2" />
+    <PackageReference Include="AWSSDK.S3" Version="3.7.103.3" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(SolutionDir)src/Testcontainers.LocalStack/Testcontainers.LocalStack.csproj" />
+    <ProjectReference Include="$(SolutionDir)tests/Testcontainers.Commons/Testcontainers.Commons.csproj" />
+  </ItemGroup>
+</Project>
```
HofmeisterAn:
```suggestion
<Project Sdk="Microsoft.NET.Sdk">
    <PropertyGroup>
        <TargetFrameworks>net6.0</TargetFrameworks>
        <IsPackable>false</IsPackable>
        <IsPublishable>false</IsPublishable>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.4.1"/>
        <PackageReference Include="coverlet.msbuild" Version="3.2.0"/>
        <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5"/>
        <PackageReference Include="xunit" Version="2.4.2"/>
        <PackageReference Include="AWSSDK.CloudWatchLogs" Version="3.7.104.14"/>
        <PackageReference Include="AWSSDK.DynamoDBv2" Version="3.7.101.42"/>
        <PackageReference Include="AWSSDK.S3" Version="3.7.103.3"/>
        <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.101.7"/>
        <PackageReference Include="AWSSDK.SQS" Version="3.7.100.71"/>
    </ItemGroup>
    <ItemGroup>
        <ProjectReference Include="$(SolutionDir)src/Testcontainers.LocalStack/Testcontainers.LocalStack.csproj"/>
        <ProjectReference Include="$(SolutionDir)tests/Testcontainers.Commons/Testcontainers.Commons.csproj"/>
    </ItemGroup>
</Project>
```

tests/Testcontainers.LocalStack.Tests/LocalStackContainerTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+namespace Testcontainers.LocalStack;
+
+public sealed class LocalStackContainerTest : IAsyncLifetime
+{
+    private readonly LocalStackContainer _localStackContainer = new LocalStackBuilder().Build();
+
+    public LocalStackContainerTest()
```
HofmeisterAn:
```suggestion
    static LocalStackContainerTest()
```

HofmeisterAn (commented):
If we do not need a specific type to set services and we are ok with the health check wait strategy, I am fine with the PR. WDYT @eddumelendez @mdelapenya?

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: fixed

src/Testcontainers.LocalStack/LocalStackBuilder.cs
dominikus1993: fixed

tests/Testcontainers.LocalStack.Tests/LocalStackContainerTests.cs
dominikus1993: fixed

tests/Testcontainers.LocalStack.Tests/Testcontainers.LocalStack.Tests.csproj
dominikus1993: fixed

tests/Testcontainers.LocalStack.Tests/LocalStackContainerTests.cs
dominikus1993: fixed

HofmeisterAn (approved):

Issue #762, opened 2023-02-07
Add webdriver container


refaeli-io:
**Is your feature request related to a problem? Please describe.**
Hi with the java library of test containers, there is the option to add containers of Webdriver:
https://www.testcontainers.org/modules/webdriver_containers/
I want the same option with dotnet.

I would like to help develop this feature because I am an automation developer with selenium and dotnet that a lot about the DevOps world (see my projects )

how we can start developing this? 
Collaborate with me :) I will like to help

HofmeisterAn:
Thanks for reaching out! We appreciate your interest in Testcontainers for Java Webdriver container and agree that equivalent support for .NET would be a great addition. With the latest .NET release, the ability to implement independent modules opens up the possibility for us to implement the Webdriver container for .NET as well. While it might not be an easy task, we would be more than happy to accept a contribution in this regard. I would suggest starting by getting familiar with the Java implementation and the new [module](https://www.atomicjar.com/2023/02/how-to-create-a-testcontainers-for-dotnet-module/) API in .NET. I am available to provide any assistance needed.

refaeli-io:
@HofmeisterAn 
Thanks :sunglasses:

ShaharM7:
The code is merged 😀  :confetti_ball:
This issue can be close :tada:

refaeli-io:
The code is merged 😀 🎊
This issue can be closed 🎉

PR #780, opened 2023-02-13
feat: Add WebDriver module

refaeli-io:
WebDriver container - use selenium docker [docker-selenium](https://github.com/SeleniumHQ/docker-selenium).
The idea is very similar to the dynamic grid approach. When there is the demand for containers then it is raised.
The idea is: 
1) Mapping between the RemoteWebDriver and the WebDriverCotnainer by entering to RemoteWebDriver as the entry point of the WebDriverCotnainer (on the inner port of 4444)
2) Wait for the container to be ready for running tests.
3) Support multi browsers running
4) Support video recording and the option to set the volume for reporting 

## What does the p.r describe?
The p.r describe how to raise up a webdriver container -> selenium standalone on demand of which browser to run on.
For starter the design architecture that the image will rise up for tests is a selenium standalone image that containers all the selenium grid components: [selenium-grid-componenets](https://www.selenium.dev/documentation/grid/components/) it includes the: 
router, session-queue, session map (his image call: [sessions](https://hub.docker.com/r/selenium/sessions), event bus, and distributer which are the components of the [hub](https://www.selenium.dev/documentation/grid/components/).|
The hub joins with the node - where the tests will run. 
The possible browsers are: 
[firefox](https://hub.docker.com/r/selenium/standalone-firefox)
[edge](https://hub.docker.com/r/selenium/standalone-edge)
[chrome](https://hub.docker.com/r/selenium/standalone-chrome)
The running can contain videos, different running options, an inspection of vnc server, and so on.

## Why is it important?
When we are running e2e tests using a selenium grid, we run the tests on the prepared beforehand selenium grid, it is using a lot of computing resources because we have a full application that always running with a const number of machines that can be run on. when one developer runs his tests it is no problem, he has enough machine - nodes to run on.
But when 10 developers running e2e testing it become a more problematic and messages start be stack in the session-queue and then we get session timeout exception.
Also, we can use a dynamic grid that selenium gives us this option. But again node docker restricts us with the number of containers we are running.
If we will run on demand by white box testing approach. Each test class will run in the same container and then we will get more indications of where the tests fail.
In addition, we can get a video of the running tests by running another container in parallel to the standalone container running.
By that, we can get the attached correct video running to the test.

## Related issues

Require #762 

## How to test this PR
To start we need to do unit testing for example: When creating a webdriver with some options then the container that will rise up contains the correct options running.


src/Testcontainers.WebDriver/Testcontainers.WebDriver.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+        <LangVersion>latest</LangVersion>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="JetBrains.Annotations" Version="2022.3.1" />
+        <PackageReference Include="Selenium.Support" Version="4.8.0" />
+        <PackageReference Include="Selenium.WebDriver" Version="4.8.0" />
```
HofmeisterAn: Modules do not contain 3rd party dependencies, although the corresponding test does. Otherwise, developers that are using different versions of the dependency may struggle to use the module.

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,95 @@
+namespace WebDriver;
```
HofmeisterAn: The namespace should be `Testcontainers.WebDriver`

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,95 @@
...
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class WebDriverBuilder : ContainerBuilder<WebDriverBuilder, WebDriverContainer, WebDriverConfiguration>
+{
+    public const string ChromeStandaloneImage = "selenium/standalone-chrome";
+    public const string FirefoxStandaloneImage = "selenium/standalone-firefox";
+    public const string EdgeStandaloneImage = "selenium/standalone-edge";
+    public const string OperaStandaloneImage = "selenium/standalone-opera";
```
HofmeisterAn: @kiview @eddumelendez does it make sense to pin the version of the browsers?

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,95 @@
...
+        //    Append the module configuration to base.Init() e.g. base.Init().WithImage("alpine:3.17") to set the modules' default image.
+        DockerResourceConfiguration = Init().DockerResourceConfiguration;
+
+        // 2) To customize the ContainerBuilder validation override the "void Validate()" method.
+        //    Use Testcontainers' Guard.Argument<TType>(TType, string) or your own guard implementation to validate the module configuration.
+
+        // 3) Add custom builder methods to extend the ContainerBuilder capabilities such as "WebDriverBuilder WithWebDriverConfig(object)".
+        //    Merge the current module configuration with a new instance of the immutable WebDriverConfiguration type to update the module configuration.
+
+        // DockerResourceConfiguration = Init().DockerResourceConfiguration;
```
HofmeisterAn:
```suggestion
        DockerResourceConfiguration = Init().DockerResourceConfiguration;
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="WebDriverBuilder" /> class.
+    /// </summary>
+    /// <param name="resourceConfiguration">The Docker resource configuration.</param>
+    private WebDriverBuilder(WebDriverConfiguration resourceConfiguration) : base(resourceConfiguration)
+    {
+        DockerResourceConfiguration = resourceConfiguration;
+    }
+
+    // /// <inheritdoc />
```
HofmeisterAn:
```suggestion
    /// <inheritdoc />
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,95 @@
...
+    // /// Sets the WebDriver config.
+    // /// </summary>
+    // /// <param name="config">The WebDriver config.</param>
+    // /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    // public WebDriverBuilder WithWebDriverConfig(object config)
+    // {
+    //     // Extends the ContainerBuilder capabilities and holds a custom configuration in WebDriverConfiguration.
+    //     // In case of a module requires other properties to represent itself, extend ContainerConfiguration.
+    //     return Merge(DockerResourceConfiguration, new WebDriverConfiguration(config: config));
+    // }
```
HofmeisterAn:
```suggestion
```

src/Testcontainers.WebDriver/WebDriverConfiguration.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="WebDriverConfiguration" /> class.
+    /// </summary>
+    /// <param name="oldValue">The old Docker resource configuration.</param>
+    /// <param name="newValue">The new Docker resource configuration.</param>
+    public WebDriverConfiguration(WebDriverConfiguration oldValue, WebDriverConfiguration newValue)
+        : base(oldValue, newValue)
+    {
+        // // Create an updated immutable copy of the module configuration.
+        // Config = BuildConfiguration.Combine(oldValue.Config, newValue.Config);
```
HofmeisterAn:
```suggestion
```

tests/Testcontainers.WebDriver.Tests/WebDriverContainerTest.cs
```diff
@@ -0,0 +1,31 @@
+namespace Testcontainers.WebDriver;
+
+using System;
+using Xunit.Abstractions;
+
+public class WebDriverContainerTest : IAsyncLifetime
+{
+  private readonly ITestOutputHelper testOutputHelper;
```
HofmeisterAn: I woud favor the container logger (`_webDriverContainer.Logger`), but only writes to the debug output (not test output).

HofmeisterAn (commented):
In general, the project looks good. Still, the interesting part is missing (adding and combining all the required resources to run Webdriver). Keep in mind you need to hold more resources to start the Webdriver (network and multiple container).

BTW, I [replied](https://stackoverflow.com/a/75434066/690017) yesterday to a similar topic on StackOverflow, maybe it's interesting for you.

src/Testcontainers.WebDriver/WebDriverBuilder.cs
eddumelendez: In Testcontainers for Java, the approach is to declare which images are compatible with the Testcontainers class implementation and if there is a custom image then `DockerImageName.parse("my-custom-image").asCompatibleSubstituteFor("mysql")`. I think there is not similar support in dotnet, yet. IMO, dotnet should follow its own approach until the same support to Java is added to the project.

kiview: The way we do it in Java is a bit uncommon for this class, since we infer the image from the webdriver capabilities:
https://github.com/testcontainers/testcontainers-java/blob/77a3c3726500dc07159762cfddc052596e27f420/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java#L260-L274

We don't need to replicate this logic in the .NET implementation. 

Regarding @eddumelendez's point, in Java we only assert for the image name for compatibility, not the actual version.

src/Testcontainers.WebDriver/Testcontainers.WebDriver.csproj
refaeli-io: O.k
I need to think again if the responsible for mapping between the: RemoteWebDriver that charges on sending the request running test, and the WebDriverContainer is another us or another developer.
Because there is  [stereotype ](https://www.selenium.dev/documentation/grid/configuration/toml_options/#configuring-and-customising-drivers)
To fit between them.
And then it also covers the browsers issue

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: That is what I think to do - the [answer ](https://github.com/testcontainers/testcontainers-dotnet/pull/780#discussion_r1106900073) for @HofmeisterAn.
I think also to not replicate this logic in the .NET implementation.
So maybe we **don't** assert the image name and version from compatibility - let the developer - the client does it ?

tests/Testcontainers.WebDriver.Tests/WebDriverContainerTest.cs
refaeli-io: O.k

src/Testcontainers.WebDriver/Testcontainers.WebDriver.csproj
vlaskal: Implementation in Java can help you to make your decision. Here is a [new getter method](https://github.com/testcontainers/testcontainers-java/blob/main/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java#L293) to get just Selenium URL. And [old method](https://github.com/testcontainers/testcontainers-java/blob/main/modules/selenium/src/main/java/org/testcontainers/containers/BrowserWebDriverContainer.java#L334) to get whole WebDriver is marked as obsolete. 
It has certain small disadvantage that video recording container starts immediately when Selenium container is in running state. It means that it records just Selenium logo till the  point when WebDriver session is created which can be in my case even several minutes. But I think with good API design it can be easily extended by developers.

vlaskal:
@ShaharM7 would be part of the `WebDriverBuilder` also automatic video recording container? 
It could work like
```csharp
var videoFilePath = "test/result/folder/video.mp4";
var container = new WebDriverBuilder().WithVideoRecording(videoFilePath).Build();
await container.StartAsync();
```  
and both containers can be spun up. 
In case I can help. I have already somewhere implementation of this.

HofmeisterAn:
> would be part of the `WebDriverBuilder` also automatic video recording container?

Yes, the idea is that the WebDriver module sets up everything internally. Calling `StartAsync()` should create all necessary resources and spin up the containers. We probably need something to store and resolve dependent resources, similar to Java's [dependsOn](https://www.testcontainers.org/features/startup_and_waits/#depending-on-another-container), but for all kinds of resources.

refaeli-io:
@HofmeisterAn 
Great idea, we can make a generic depend on and then a custom for selenium
[The order will be](https://github.com/SeleniumHQ/docker-selenium#video-recording): 
1) create a network 
2) run selenium standalone container - browser: firefox, chrome, and edge
3) run a selenium video container that uses the same network
4) stop the video container and remove it 
5) stop the standalone selenium container and remove it 
6) delete the network

refaeli-io:
@vlaskal I will like your help :) contact me with .slack

src/Testcontainers.WebDriver/Testcontainers.WebDriver.csproj
refaeli-io: @vlaskal 
Sorry about the delay I was seak
For our purposes, I don't want to implement it like java, because they use the DriverOptions (chroneOptions and so on) 
Yes, I want to do the mapping between the Url of `new RemoteWebDriver(string URL,  DriverOptions options)` and `new WebDriverBuilder().Build()`
The next stage will be to connect it to the video recording cotnainer

Testcontainers.sln.DotSettings
```diff
@@ -39,6 +39,7 @@
   <s:Boolean x:Key="/Default/UserDictionary/Words/=testcontainer/@EntryIndexedValue">True</s:Boolean>
   <s:Boolean x:Key="/Default/UserDictionary/Words/=testcontainers/@EntryIndexedValue">True</s:Boolean>
   <s:Boolean x:Key="/Default/UserDictionary/Words/=tlsverify/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Toml/@EntryIndexedValue">True</s:Boolean>
```
HofmeisterAn:
```suggestion
	<s:Boolean x:Key="/Default/UserDictionary/Words/=toml/@EntryIndexedValue">True</s:Boolean>
```

If you add something (a user dictionary item) to `Testcontainers.sln.DotSettings` you need to add it to `Testcontainers.dic` too.

src/Testcontainers.WebDriver/BrowserType.cs
```diff
@@ -0,0 +1,22 @@
...
+        public static BrowserType Firefox => new BrowserType("firefox");
+        public static BrowserType MicrosoftEdge => new BrowserType("MicrosoftEdge");
+        public static BrowserType Opera => new BrowserType("opera");
+        
+        public override string ToString()
+        {
+            return Value;
+        }
+    }
+}
```
HofmeisterAn:
```suggestion
namespace Testcontainers.WebDriver;

public class BrowserType
{
    private BrowserType(string value)
    {
        Value = value;
    }

    public static BrowserType Chrome = new BrowserType("Chrome");

    public static BrowserType Firefox = new BrowserType("Firefox");

    public static BrowserType Edge = new BrowserType("Edge");

    public static BrowserType Opera = new BrowserType("Opera");

    public string Value { get; }

    public override string ToString()
    {
        return Value;
    }
}
```

src/Testcontainers.WebDriver/Usings.cs
```diff
@@ -0,0 +1,11 @@
...
+global using System.Collections.Generic;
+global using System.Linq;
+global using Docker.DotNet.Models;
+global using DotNet.Testcontainers;
+global using DotNet.Testcontainers.Builders;
+global using DotNet.Testcontainers.Configurations;
+global using DotNet.Testcontainers.Containers;
+global using JetBrains.Annotations;
+global using Microsoft.Extensions.Logging;
+global using static Testcontainers.WebDriver.WebDriverBuilder;
```
HofmeisterAn:
```suggestion
global using System;
global using System.Collections.Generic;
global using System.Linq;
global using Docker.DotNet.Models;
global using DotNet.Testcontainers;
global using DotNet.Testcontainers.Builders;
global using DotNet.Testcontainers.Configurations;
global using DotNet.Testcontainers.Containers;
global using JetBrains.Annotations;
global using Microsoft.Extensions.Logging;
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,178 @@
...
+                            .GetProperty("ready")
+                            .GetBoolean();
+                        
+                        return isReady;
+                    }
+                    catch
+                    {
+                        return false;
+                    }
+                }));
```
HofmeisterAn: Move the wait strategy to `Init()`. Your wait strategy does not depend on any configuration that is a user input. I suggest something like:

```csharp
    protected override WebDriverBuilder Init()
    {
        return base.Init()
            .WithImage(FirefoxStandaloneImage)
            .WithPortBinding(WebDriverPort, true)
            .WithPortBinding(VncServerPort, true)
            .WithPortBinding(NoVncServerPort, true)
            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request 
                => request.ForPath("/wd/hub/status").ForPort(WebDriverPort).ForResponseMessageMatching(SeleniumIsReady)));
    }
```

Then implement your HTTP response predicate in `SeleniumIsReady`.

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// <summary>
+    /// Sets the WebDriver config.
+    /// </summary>
+    /// <param name="configTomlFilePath">The config toml file path.</param>
+    /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    public WebDriverBuilder SetConfigurationFromTomlFile(string configTomlFilePath)
+    {
+        return Merge(DockerResourceConfiguration, new WebDriverConfiguration())
+            .WithVolumeMount(configTomlFilePath, "/opt/bin/config.toml");
```
HofmeisterAn:
```suggestion
            .WithResourceMapping(configTomlFilePath, "/opt/bin/config.toml");
```

tests/Testcontainers.WebDriver.Tests/Usings.cs
```diff
@@ -0,0 +1,4 @@
+global using System.Threading.Tasks;
+global using DotNet.Testcontainers.Commons;
+global using Xunit;
+global using WebDriver;
```
HofmeisterAn:
```suggestion
global using System.Threading.Tasks;
global using DotNet.Testcontainers.Commons;
global using Xunit;
```

tests/Testcontainers.WebDriver.Tests/Testcontainers.WebDriver.Tests.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <PackageReference Include="coverlet.msbuild" Version="3.2.0"/>
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5"/>
+    <PackageReference Include="xunit" Version="2.4.2"/>
+    <PackageReference Include="StackExchange.Redis" Version="2.6.90"/>
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(SolutionDir)src/Testcontainers.WebDriver/Testcontainers.WebDriver.csproj"/>
+    <ProjectReference Include="$(SolutionDir)tests/Testcontainers.Commons/Testcontainers.Commons.csproj"/>
+  </ItemGroup>
+</Project>
```
HofmeisterAn:
```suggestion
<Project Sdk="Microsoft.NET.Sdk">
    <PropertyGroup>
        <TargetFrameworks>net6.0</TargetFrameworks>
        <IsPackable>false</IsPackable>
        <IsPublishable>false</IsPublishable>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.4.1"/>
        <PackageReference Include="coverlet.msbuild" Version="3.2.0"/>
        <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5"/>
        <PackageReference Include="xunit" Version="2.4.2"/>
    </ItemGroup>
    <ItemGroup>
        <ProjectReference Include="$(SolutionDir)src/Testcontainers.WebDriver/Testcontainers.WebDriver.csproj"/>
        <ProjectReference Include="$(SolutionDir)tests/Testcontainers.Commons/Testcontainers.Commons.csproj"/>
    </ItemGroup>
</Project>
```

tests/Testcontainers.WebDriver.Tests/WebDriverContainerTest.cs
```diff
@@ -0,0 +1,31 @@
...
+  {
+    return _webDriverContainer.DisposeAsync().AsTask();
+  }
+
+  [Fact]
+  [Trait(nameof(DockerCli.DockerPlatform), nameof(DockerCli.DockerPlatform.Linux))]
+  public void CreateContainer()
+  {
+  }
+}
```
HofmeisterAn:
```suggestion
namespace Testcontainers.WebDriver;

public class WebDriverContainerTest : IAsyncLifetime
{
    private readonly WebDriverContainer _webDriverContainer = new WebDriverBuilder().Build();

    public Task InitializeAsync()
    {
        return this._webDriverContainer.StartAsync();
    }

    public Task DisposeAsync()
    {
        return this._webDriverContainer.DisposeAsync().AsTask();
    }

    [Fact]
    [Trait(nameof(DockerCli.DockerPlatform), nameof(DockerCli.DockerPlatform.Linux))]
    public void CreateContainer()
    {
    }
}
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// <summary>
+    /// Sets the WebDriver config.
+    /// </summary>
+    /// <param name="options">The options as string list for starting a hub or a node.</param>
+    /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    public WebDriverBuilder WithConfigurationOptions(string options)
+    {
+        return Merge(DockerResourceConfiguration, new WebDriverConfiguration())
+            .WithEnvironment("SE_OPTS", options);
```
HofmeisterAn: I noticed you use often `Merge(DockerResourceConfiguration, new WebDriverConfiguration())`. If you do not set any property in the configuration object (`WebDriverConfiguration`) and just forward the member call to its base class you do not need to call `Merge(WebDriverConfiguration, WebDriverConfiguration)`. In the example above it is enough to return immediately.

```csharp
return WithEnvironment("SE_OPTS", options);
```

HofmeisterAn (commented):
I've added some suggestions. Please add an EditorConfig file to the test project directory too.

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: I will change it 
But why with WithResourceMapping? 
I based on [that](https://github.com/SeleniumHQ/docker-selenium#configuration-example)

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: o.k 
Thank you
I will change it

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: Needs for waiting for the message after the container rise
So it shouldn't be in the `Build()` method?

Testcontainers.dic
```diff
@@ -31,4 +31,5 @@ sqlplus
 testcontainer
 testcontainers
 tlsverify
-vstest
\ No newline at end of file
+vstest
+toml
```
HofmeisterAn:
```suggestion
toml
vstest
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    /// <summary>
+    /// Sets the grid additional commandline parameters for starting.
+    /// </summary>
+    /// <remarks>
+    /// https://github.com/SeleniumHQ/docker-selenium#se_opts-selenium-configuration-options
+    /// </remarks>
+    /// <param name="options">The options as string list for starting a hub or a node.</param>
+    /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    public WebDriverBuilder WithConfigurationOptions(string options)
```
HofmeisterAn: Does it make sense to use a dictionary as arg type (then join to string)?

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    /// <summary>
+    /// Sets the environment variable to java process.
+    /// </summary>
+    /// <remarks>
+    /// https://github.com/SeleniumHQ/docker-selenium#se_java_opts-java-environment-options
+    /// </remarks>
+    /// <param name="javaOptions">The java options environment variables as string list.</param>
+    /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    public WebDriverBuilder WithJavaEnvironmentOptions(string javaOptions)
```
HofmeisterAn: Does it make sense to use a dictionary as arg type (then join to string)?

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// </summary>
+    /// <remarks>
+    /// https://github.com/SeleniumHQ/docker-selenium#grid-url-and-session-timeout
+    /// </remarks>
+    /// <param name="sessionTimeout">The Grid  session timeout config.</param>
+    /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    public WebDriverBuilder SetSessionTimeout(int sessionTimeout = 300)
+    {
+        return WithEnvironment("SE_NODE_SESSION_TIMEOUT", sessionTimeout.ToString());
+    }
```
HofmeisterAn:
```suggestion
    public WebDriverBuilder SetSessionTimeout(TimeSpan sessionTimeout = default)
    {
        return WithEnvironment("SE_NODE_SESSION_TIMEOUT", sessionTimeout.TotalSeconds.ToString(CultureInfo.InvariantCulture));
    }
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    /// <summary>
+    /// Sets grid time zone by env variable.
+    /// </summary>
+    /// <param name="timeZone">The desirable time zone.</param>
+    /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    public WebDriverBuilder SetTimeZone(string timeZone)
```
HofmeisterAn: Can we use the `TimeZoneInfo` type instead of `string`?

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    /// <inheritdoc />
+    protected override WebDriverBuilder Init()
+    {
+        return base.Init()
+            .WithBrowser(WebDriverType.Chrome)
+            .WithPortBinding(WebDriverPort, true)
+            .WithPortBinding(VncServerPort, true)
+            .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(VncServerPort))
+            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request
```
HofmeisterAn: Your 2nd wait strategy overrides the 1st. I guess we can remove `.WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(VncServerPort))`.

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,228 @@
...
+                    .ForPort(WebDriverPort)
+                    .ForResponseMessageMatching(IsGridReadyAsync)
+            ));
+    }
+
+    /// <inheritdoc />
+    protected override void Validate()
+    {
+    }
+
```
HofmeisterAn:
```suggestion
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,228 @@
...
+        try
+        {
+            var isReady = JsonDocument.Parse(jsonString)
+                .RootElement
+                .GetProperty("value")
+                .GetProperty("ready")
+                .GetBoolean();
+
+            return isReady;
```
HofmeisterAn:
```suggestion
            return JsonDocument.Parse(jsonString)
                .RootElement
                .GetProperty("value")
                .GetProperty("ready")
                .GetBoolean();
```

src/Testcontainers.WebDriver/WebDriverType.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Testcontainers.WebDriver;
+
+public readonly struct WebDriverType
+{
+    public static WebDriverType Chrome => new WebDriverType("chrome", new DockerImage("selenium/standalone-chrome"));
+    public static WebDriverType Firefox => new WebDriverType("firefox", new DockerImage("selenium/standalone-firefox"));
+    public static WebDriverType MicrosoftEdge => new WebDriverType("MicrosoftEdge", new DockerImage("selenium/standalone-edge"));
+    public static WebDriverType Opera => new WebDriverType("opera", new DockerImage("selenium/standalone-opera"));
+    public static WebDriverType Video => new WebDriverType("ffmpeg", new DockerImage("selenium/video"));
```
HofmeisterAn: Can we pin the versions please?

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: Yes, it is possible and maybe then divide it for each loop?

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: Great idea

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: and then use TimeZoneInfo.DisplayName ?

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: I think also
I will remove it

src/Testcontainers.WebDriver/WebDriverBuilder.cs
HofmeisterAn: You can simply join the key value pairs then:

```csharp
string.Join(Delimiter, options.Select(option => string.Join("=", option.Key, option.Value)));
```

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: I will remove

src/Testcontainers.WebDriver/WebDriverBuilder.cs
refaeli-io: Nice

src/Testcontainers.WebDriver/WebDriverType.cs
refaeli-io: Yes we  can pin
But the struct is public 
so the client can create a unique WebDriverType

src/Testcontainers.WebDriver/WebDriverBuilder.cs
HofmeisterAn: I am not 100% which propery. Something we need to look up, but it makes development easier. Then others do not need to think about it.

refaeli-io: This is why I think this is the display name property : 
https://learn.microsoft.com/en-us/dotnet/api/system.time
zoneinfo.local?view=net-7.0#system-timezoneinfo-local

HofmeisterAn: It looks like .NET does not return IANA. Lets keep the string then.

HofmeisterAn: Still the question if we need to expose an API for that if it simply sets and env variable.

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,219 @@
+namespace Testcontainers.WebDriver;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class WebDriverBuilder : ContainerBuilder<WebDriverBuilder, WebDriverContainer, WebDriverConfiguration>
```
HofmeisterAn, line 8: @kiview @eddumelendez I am interested in your opinion on the builder configuration and the test. Since TC for .NET does not have the `dependsOn` feature, I think the implementation looks very good. I am particularly uncertain about the APIs that rely on environment variables. It is the old discussion about which APIs we expose. WDTY?

tests/Testcontainers.WebDriver.Tests/WebDriverContainerTest.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            // When
+            driver.Navigate().GoToUrl(_helloWorldEndpoint.ToString());
+            var headingElementText = driver.FindElementByTagName("h1").Text;
+
+            await _fixture._webDriverContainer.StopAsync()
+                .ConfigureAwait(false);
+
+            await _fixture._webDriverContainer.ExportVideo(Path.Combine(TestSession.TempDirectoryPath, Path.GetRandomFileName()))
+                .ConfigureAwait(false);
```
HofmeisterAn: @kiview @eddumelendez to read the video from the FFmpeg container, it is necessary to stop it first. How does the VNC screen recording container does it?

HofmeisterAn: We should add an assert here, I think.

refaeli-io: I can tell that you have to stop, otherwise, the recording file becomes corrupt.
I talked with one of selenium developers on that

refaeli-io: `Assert.True(File.Exists(_videoRecordingPath));`

src/Testcontainers.WebDriver/WebDriverBuilder.cs
eddumelendez: Setting env vars in the container configuration couple the implementation with the image itself. Unless, of course, there is certain logic to do with those as in other implementations. I think those can be removed `WithConfigurationOptions`, `WithJavaEnvironmentOptions`, `SettingScreenResolution`, `SetSessionTimeout`, `SetTimeZone` because can be easily added with `WithEnvironment(k, v)`. It is the same case as in Azurite implementation IIRC.

Besides that, LGTM

eddumelendez: In tc-java, currently, there is no usage of `dependsOn` between `BrowserWebDriverContainer` and `VncContainer`. `VncContainer` is started manually once the `BrowserWebDriverContainer` started.

tests/Testcontainers.WebDriver.Tests/WebDriverContainerTest.cs
eddumelendez: It could be that the `selenium/video` image works differently than `testcontainers/vnc-recorded`. Given that tc-java is using `testcontainers/vnc-recorded`, it doesn't need to stop it in order to copy the video.

src/Testcontainers.WebDriver/WebDriverBuilder.cs
HofmeisterAn: > Setting env vars in the container configuration couple the implementation with the image itself. Unless, of course, there is certain logic to do with those as in other implementations.

Indeed. I have the same opinion.

> In tc-java, currently, there is no usage of `dependsOn` between `BrowserWebDriverContainer` and `VncContainer`. `VncContainer` is started manually once the `BrowserWebDriverContainer` started.

I see. In any case, once we have the `dependsOn` feature, implementation will be even simpler.

tests/Testcontainers.WebDriver.Tests/WebDriverContainerTest.cs
refaeli-io: @eddumelendez 
How do you get updated about the testcontainers/vnc-recorded` image? 
If is there a change in the configuration or env? 
I prepare to be depend only on selenium

src/Testcontainers.WebDriver/WebDriverBuilder.cs
```diff
@@ -0,0 +1,225 @@
...
+    /// </summary>
+    /// <remarks>
+    /// Time Zone Database name (IANA).
+    /// </remarks>
+    /// <param name="timeZone">The time zone.</param>
+    /// <returns>A configured instance of <see cref="WebDriverBuilder" />.</returns>
+    public WebDriverBuilder SetTimeZone(string timeZone)
+    {
+        return WithEnvironment("TZ", timeZone);
+    }
```
HofmeisterAn:
```suggestion
```

As we previously discussed [here](https://github.com/testcontainers/testcontainers-dotnet/pull/780#discussion_r1130782659), it does not make much sense to expose individual APIs for environment variables. It is always a question of which ones are worth exposing, who keeps them up-to-date, etc. Since developers can simply add environment variables by calling `WithEnvironment(k, v)`, I propose removing these APIs.

refaeli-io: I agree, for example, config from the Toml file is not mentioned in selenium docs and this is an approach that selenium adopts and adds docs most of the chance they will not remove it 

But maybe declarative about that in docs or some examples?

HofmeisterAn (approved):
Thank you for submitting the WebDriver module, we appreciate your contribution and the time you took to create this PR. Everything looks good to me 🙏.

refaeli-io:
Thanks so much for your help :) and support too

PR #782, opened 2023-02-13
feat: Introduce new Azurite module

vlaskal:
## What does this PR do?
This PR brings replacement of old Azurite module by new one with new module structure. 

## Why is it important?
Be able to remove obsolete Azurite module from the core testcontainers-dotnet.

## Related issues

- Relates #493 

## How to test this PR
All the Azurite test should pass. 

## Follow-ups
Missing features:
- [x] HTTPS support
- [x] OAuth support
- [x] Remove obsolete Azurite folder

All features were already requested and will be added into this PR but discussion can already start.


src/Testcontainers.Azurite/AzuriteConfiguration.cs
```diff
@@ -0,0 +1,213 @@
...
+
+/// <inheritdoc cref="ContainerConfiguration" />
+[PublicAPI]
+public sealed class AzuriteConfiguration : ContainerConfiguration
+{
+  /// <summary>
+  ///   Initializes a new instance of the <see cref="AzuriteConfiguration" /> class.
+  /// </summary>
+  /// <param name="azuriteServices">A value indicating which Azurite service is turned on.</param>
+  /// <param name="blobPort">The Azurite container blob port.</param>
```
HofmeisterAn: Why is it necessary to configure the port of the service running inside the container? Why can't we just use the default ports like other modules do?

src/Testcontainers.Azurite/AzuriteConfiguration.cs
```diff
@@ -0,0 +1,213 @@
...
+  /// <param name="blobPortBinding">The host blob port to bind.</param>
```
HofmeisterAn: For modules, I do not like to promote fixed host port binding. This is considered as bad practice. We're seeing developers using it and having trouble to understand the consequences.

src/Testcontainers.Azurite/AzuriteConfiguration.cs
```diff
@@ -0,0 +1,213 @@
...
+  /// <param name="queuePort">The Azurite container queue port.</param>
+  /// <param name="queuePortBinding">The host queue port to bind.</param>
+  /// <param name="tablePort">The Azurite container table port.</param>
+  /// <param name="tablePortBinding">The host table port to bind.</param>
+  /// <param name="workspaceLocation">The Azurite workspace directory path.</param>
```
HofmeisterAn: Here, I have the same opinion as above (see `blobPort`). Why would it be necessary to configure the workspace location inside the container?

If we keep the default location, developers can simply use the container builder members `WithMount(...)`, `WithBindMount(...)`, `WithVolumeMount(...)` or `WithTmpfsMount(...)` to configure it. Now, the are bound to `WithBindMount(...)` which is considered a bad practice again.

src/Testcontainers.Azurite/AzuriteBuilder.cs
```diff
@@ -0,0 +1,481 @@
...
+    return "azurite";
+  }
+
+  private string[] GetEnabledServices()
+  {
+    const string defaultRemoteEndpoint = "0.0.0.0";
+
+    IList<string> args = new List<string>();
+
+    if (AllServicesEnabled || BlobServiceOnlyEnabled)
```
HofmeisterAn: I think it would make sense to put a service in its own class. Then we do not need all the `if-else` blocks. We can simply iterate over the list. Either add enabled services to the list or enable them with a property (bool).

HofmeisterAn (commented):
Hi Vlastimil, thank you for submitting your pull request. However, I had a more object-oriented and simpler approach in mind.

Before I discuss the details of the implementation, I would like to have a discussion with you regarding the module configuration in general. Please take a look at the comments I left on the pull request, and let me know your thoughts.

/cc @eddumelendez @kiview @mdelapenya I am interested in your thoughts too.

Thank you again for your contribution!

eddumelendez (commented):
I agree with @HofmeisterAn comments. Azurite is quite simple to use with Testcontainers using the defaults. I don't see any benefit by exposing all the options but if those are needed then the Testcontainers for DotNet is already flexible enough to allow it.

src/Testcontainers.Azurite/AzuriteConfiguration.cs
vlaskal: In general there is no need. Even if you run several Azurite containers at once it is fine. 

I know about few cases where the ports were changed and it was because of defensive programming. Developers connect to wrong container instance and system behaves unpredictable. And second use case was when Azurite was hidden behind Mockserver to simulate error responses from one specific service or some cases.

My original intention was to implement full featured module based on [documentation](https://github.com/Azure/Azurite#supported-command-line-options) for Azurite to eliminate needs to do own implementations. Basically take u module and based and set it up same way as I can do directly via docker or in command line.

I was thinking also to add `string ContainerConnectionString { get; }` or maybe `GetContainerConnectionString(string nameOrAliasOrIp)` which is value used by other containers running in same docker, network etc. And if ports will be changed then I need to know new port for each service to be able to generate any connection string properly.

As you know I asked what is Testcontainers approach because I am aware that kibd of this module will need more effort for maintenance when newer version of Ayurite will be released. If you think it is too much then I will use just standard ports as other modules do.

src/Testcontainers.Azurite/AzuriteConfiguration.cs
vlaskal: I know and I understand and is valid for all CI runners. But there is also important second use case which is developing and troubleshooting tests. In this situation I as developer want to fix binded port between debug sessions to be able to connect via other tools. It is really annoing to change port in connection strings for each run. 

Azurite offers three services inside container therefore I need to know which port is binded to which service to be able to construct connection string correctly for developers. But if we use only standard ports then developer can use core method `.WithPortBindig()` to bind them.

src/Testcontainers.Azurite/AzuriteConfiguration.cs
vlaskal: Yes `WorkspaceLocation` and `WorkspaceLocationBinding` are same as `BlobPort` and `BlobPortBinding` cases. I will update them based on our decision. Because there is one location then it is easier situation than with ports.

src/Testcontainers.Azurite/AzuriteBuilder.cs
vlaskal: This is good idea because there is XOR relationship between services: ALL xor BLOB xor QUEUE xor TABLE. we can provide 4 independent containers maybe with one generic builder. Not sure right now. I am just thinking out loud.

vlaskal:
I agree with you both @HofmeisterAn and @eddumelendez . But I would like to point out that I see 3 main benefit of the Testcontainers Modules : 

1. start/stop container
2. configure container specific stuff easily via descriptive API
3. get outputs usually connection strings or urls to dynamically glued services together

But if something is changed from default values then eg. connection strings (3) cannot be generated, I must use more generic (less descriptive) API (2) and 1 is not much at this point and whole module is useless. So I was thinking what real value it brings and how that value increase. This is how I was thinking about it.

eddumelendez:
Agree with 1 and 3 but having all of those atributes to define every option for the container will couple the builder definition with the image. As mentioned before this can be done with the Testcontainers API.

src/Testcontainers.Azurite/AzuriteConfiguration.cs
HofmeisterAn: > Basically take u module and based and set it up same way as I can do directly via docker or in command line.

Something I also had to learn. Looking into Java's implementation (it has mature modules), modules are not a 1:1 implementation of their image. Its kinda configuration that is immediately ready to use. Users should not need to spend a lot of time configuring modules.

I still do not understand the benefit of configuring the container ports. Developers should use the container's connection string. For container-to-container communication you can just use the default ports.

> I was thinking also to add `string ContainerConnectionString { get; }` or maybe `GetContainerConnectionString(string nameOrAliasOrIp)`

That's a nice idea as general follow-up, but still, you can use the default ports here too. @eddumelendez @kiview does Java construct connection strings for container-to-container communication?

src/Testcontainers.Azurite/AzuriteConfiguration.cs
HofmeisterAn: >But there is also important second use case which is developing and troubleshooting tests. In this situation I as developer want to fix binded port between debug sessions to be able to connect via other tools.

In that case developers can still override the module configuration e.g.:

```csharp
_ = new MsSqlBuilder().WithPortBinding(1433, MsSqlBuilder.MsSqlPort).Build();
```

> Azurite offers three services inside container therefore I need to know which port is binded to which service to be able to construct connection string correctly for developers.

The question that arises is whether the services need to be configurable at all.

E.g. what is the benefit from disabling services? When does a developer need to disable a service? If we look into the LocalStack module we do not configure any services either. The container is not faster nor slower.

HofmeisterAn:
> But if something is changed from default values then eg. connection strings (3) cannot be generated, I must use more generic (less descriptive) API (2) and 1 is not much at this point and whole module is useless.

Can you elaborate that? Maybe with an example? I do not understand it.

I think this [PR](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4198/files) shows the benefits of modules very well. The OTel maintainers can replace all changes with a single line in the future, and they do not need to bother around with the configuration at all:

    private readonly MsSqlContainer msSqlContainer = new MsSqlBuilder().Build();

The [PostgreSql](https://github.com/testcontainers/testcontainers-dotnet/blob/f1a153359eac2593464b2f95de30a3f8223acba8/src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs#L82-L85) module sets best practices to increase the test performance. Developers will get better performance by just using the default module configuration.

src/Testcontainers.Azurite/AzuriteConfiguration.cs
eddumelendez: yes, we do as a quick acces to the endpoint

src/Testcontainers.Azurite/AzuriteConfiguration.cs
vlaskal: The service should be configured eg. to simulate that service is unavailable etc. or what ever resson developer can have like performance etc. Probably it differs between basic / standard usage and andvance usage.

vlaskal:
The basic usage is offered by the module as others does. You do not need to care about in case you do not need. But it case you need to configure container and not bindings then you can and you do not need to study documentation and write a module once again.

> Can you elaborate that? Maybe with an example? I do not understand it.

That sample is easy let use this code:
```csharp
var container = new PostgreSqlBuilder()
    .WithCommand("-p", "5433")
    .WithExposedPort(5433)
    .WithPortBinding(5433, true)
    .Build();
var connString = container.GetConnectionString();
```
and `connString` contains wrong conenction string and I do not see any other benefit to use the module. 

But I see your points and now to know how to finalize PR.

HofmeisterAn:
> That sample is easy let use this code:
> 
> ```cs
> var container = new PostgreSqlBuilder()
>     .WithCommand("-p", "5433")
>     .WithExposedPort(5433)
>     .WithPortBinding(5433, true)
>     .Build();
> var connString = container.GetConnectionString();
> ```

Let's consider the example above returns a connection string that maps to the right (overridden) container port. What has changed for the developer? IMO, nothing — it's still a connection string with a random assigned host port. For container-to-container communication you will use a network-alias anyway because it is required.

vlaskal:
> Let's consider the example above returns a connection string that maps to the right (overridden) container port. What has changed for the developer? IMO, nothing — it's still a connection string with a random assigned host port. For container-to-container communication you will use a network-alias anyway because it is required.

It changes a lot because developer has connection string which does not work with incorrect random mapped port. Becuase conenction string property/method tries to get mapped port for default 5432 because container does not know anything about changes provided by generic methods.

gbpsp:
Hi, I see this PR has been opened for a while now. What should we use for Azurite container? The old version is marked as obsolete but there is no documented alternative.
Thanks

HofmeisterAn:
> Hi, I see this PR has been opened for a while now. What should we use for Azurite container? The old version is marked as obsolete but there is no documented alternative. Thanks

I've created a small follow-up [PR](https://github.com/vlaskal/dotnet-testcontainers/pull/2) that updates the formatting. Something I would like to get merged first (consistent style), before I continue with the review.

HofmeisterAn:
We published the [Azurite](https://www.nuget.org/packages/Testcontainers.Azurite) module a few days ago. Although it does not include all the features from this pull request, we can add them as follow-up PRs, one after another if necessary. Meanwhile, I will close this issue.

Issue #777, opened 2023-02-12
[Bug]: Testcontainers is not detecting port mapping correctly for auto assigned ports


shoter:
### Testcontainers version

2.4.0 and current source code

### Using the latest Testcontainers version?

Yes

### Host OS

Windows 10 Build 19044

### Host arch

x64

### .NET version

6.0.400

### Docker version

```shell
Client:
 Cloud integration: v1.0.22
 Version:           20.10.11
 API version:       1.41
 Go version:        go1.16.10
 Git commit:        dea9396
 Built:             Thu Nov 18 00:42:51 2021
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.11
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.9
  Git commit:       847da18
  Built:            Thu Nov 18 00:35:39 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.12
  GitCommit:        7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc:
  Version:          1.0.2
  GitCommit:        v1.0.2-0-g52b36a2
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.1)
  scan: Docker Scan (Docker Inc., 0.9.0)

Server:
 Containers: 3
  Running: 3
  Paused: 0
  Stopped: 0
 Images: 35
 Server Version: 20.10.11
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7b11cfaabd73bb80907dd23182b9347b4245eb5d
[...]

PR #791, opened 2023-02-17
fix: Wait until container ports are mapped, removed generated credentials

HofmeisterAn:
## What does this PR do?

Sets default module credentials.

## Why is it important?

If we generate credentials, we will be unable to detect whether a resource is reusable or not.

## Related issues

- Relates #777


src/Testcontainers/Containers/DockerContainer.cs
```diff
@@ -405,19 +405,29 @@ await this.client.AttachAsync(this.container.ID, this.configuration.OutputConsum
...
+      var timeout = (int)TimeSpan.FromSeconds(15).TotalMilliseconds;
+
+      async Task<bool> CheckPortBindings()
+      {
+        this.container = await this.client.InspectContainer(this.container.ID, ct)
+          .ConfigureAwait(false);
+
+        var boundPorts = this.container.NetworkSettings.Ports.Values.Where(portBindings => portBindings != null).SelectMany(portBinding => portBinding).Count(portBinding => !string.IsNullOrEmpty(portBinding.HostPort));
+        return this.configuration.PortBindings == null || /* IPv4 or IPv6 */ this.configuration.PortBindings.Count == boundPorts || /* IPv4 and IPv6 */ 2 * this.configuration.PortBindings.Count == boundPorts;
+      }
```
HofmeisterAn: @cristianrgreco @eddumelendez @kiview @mdelapenya  I am interessted in your opinion on how you handle a particular situation in other language implementations. When using the Docker.DotNet `InspectContainer(string, CancellationToken)` member, it is possible to receive a response that does not immediately contain all port bindings. In most cases, this is not an issue. Soon as the wait strategy indicate readiness, we receive an instance that contains all information.

However, there are two cases where it may be useful to wait until the instance received all port bindings:

- The user did not set a wait strategy (although this is probably not a critical one).
- Custom code is being run during startup, as it is the case with [Couchbase](https://github.com/testcontainers/testcontainers-java/blob/74e8954f55401aa4fea64aed8d63e4cc3e258316/modules/couchbase/src/main/java/org/testcontainers/couchbase/CouchbaseContainer.java#L336-L350).

`CheckPortBindings()` compares the developer's configuration against the response and checks if it contains the same amount of port bindings.

kiview: As part of the container startup code in Java, we will wait until the inspect command returns all mapped ports:
https://github.com/testcontainers/testcontainers-java/blob/6fe954ab940ccb674d85aca543fe4c979e8932f0/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L450

Only afterwards the callback method of extending containers is called. So the callback methods can be sure to already get a consistent response regarding the ports.

PR #793, opened 2023-02-19
feat: Replace output consumer with `IContainer.GetLogsAsync(DateTime, DateTime, bool, CancellationToken)`

HofmeisterAn:
## What does this PR do?

This pull request sets the `WithOutputConsumer(IOutputConsumer)` member obsolete, as it is no longer necessary to attach a stream to consume the output of a container. Instead, getting `stdout` or `stderr` is much easier now, as developers can simply call `IContainer.GetLogsAsync(DateTime, DateTime, bool, CancellationToken)`.

## Why is it important?

This PR is an important step in continuing the clean up of the API, as it further reduces the amount of configuration necessary to get log messages.

## Related issues

\-

### DIFF
--- a/src/Testcontainers/Containers/DockerContainer.cs
+++ b/src/Testcontainers/Containers/DockerContainer.cs
@@ -244,9 +244,9 @@ namespace DotNet.Testcontainers.Containers
     }
 
     /// <inheritdoc />
-    public Task<(string Stdout, string Stderr)> GetLogs(DateTime since = default, DateTime until = default, CancellationToken ct = default)
+    public Task<(string Stdout, string Stderr)> GetLogs(DateTime since = default, DateTime until = default, bool timestampsEnabled = true, CancellationToken ct = default)
     {
-      return this.client.GetContainerLogs(this.Id, since, until, ct);
+      return this.client.GetContainerLogs(this.Id, since, until, timestampsEnabled, ct);
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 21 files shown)


src/Testcontainers/Containers/DockerContainer.cs
```diff
@@ -385,13 +397,10 @@ protected virtual async Task UnsafeStartAsync(CancellationToken ct = default)
       await this.UnsafeCreateAsync(ct)
         .ConfigureAwait(false);
 
-      await this.client.AttachAsync(this.container.ID, this.configuration.OutputConsumer, ct)
```
andriivitiv, line 388: This was such a useful feature. Don't understand why it got removed. In my case I use it to write to xunit's `ITestOutputHelper` and can see logs in IDE at runtime. Now the only option is to run a background task that will periodically call `GetLogsAsync`. For now, as a workaround to preserve the previous behavior, I override `UnsafeStartAsync` and call `AttachAsync` via reflection.

HofmeisterAn: The intention is explained in the pull request. We have noticed that developers frequently use the consumer wrong. Using the `GetLogsAsync(DateTime, DateTime, bool, CancellationToken)` method is much easier. However, please create an issue and include your use case in it, and we will consider adding the attach call again. Having both is probably not a big deal. Thanks for bringing this up.

jacobjmarks:
I second that this deprecation seems strange. I am looking to redirect output from the container on an "as-written" basis; I do not simply want to grab logs on an ad-hoc basis via a point-in-time method call. As far as I can tell, this log streaming use case is not supported without the Output Consumer API.

PR #802, opened 2023-02-24
feat: Replace Kafka module

HofmeisterAn:
## What does this PR do?

Replaces the old Kafka module approach (using an extension method) with a proper independent module.

## Why is it important?

We do not follow the old module approach anymore. We will replace the old approach with a new API and release each module as independent dependency. This module replaces Kafka.

## Related issues

- Relates #493


HofmeisterAn:
@nbarnwell can you please share the exception. I will double-check it, but I do not see an issue in the previous code right away.

nbarnwell:
I will try and replicate it, but have you already dismissed my change
anyway? When I click "view it on GitHub" in the email just now, it takes me
to PR 802, which doesn't mention me at all. I can only find my changes
through a branch I have on a fork I have. All I did was copy/paste the code
into a .cs file in a project and run the test.

Neil.

HofmeisterAn:
@nbarnwell Sorry Neil, that was my bad. I replied to the wrong PR first (I opened the wrong tab). Here is my answer: https://github.com/docker/docs/pull/16710#issuecomment-1444002965 to the right PR. Sorry for the confusion.

PR #832, opened 2023-03-13
feat: Add container builder member to depend on other Docker resources

### LINKED ISSUE #826: [Enhancement]: Support depending on other Testcontainers (Docker) resources
### Problem

The Testcontainers builder implementation does not currently support depending on other Testcontainers (Docker) resources, such as containers, networks, or volumes. It would be a valuable addition to have the ability to resolve an internal dependency graph and automatically create, start and resolve dependent resources.

### Solution

\-

### Benefit

Automatically resolving dependent resources will significantly help developers in writing simpler tests. With this feature, developers will no longer need to create the dependent resources manually.

### Alternatives

\-

Yes

HofmeisterAn:
## What does this PR do?

This pull request adds the member `DependsOn(IContainer)` to the container builder, allowing it to depend on other containers. Internally, we will create and start the dependent containers (resources) first. The same applies for networks and volumes. Please note that we do not have an overloaded member for network and volume. Instead, we use the equivalent member `WithNetwork(INetwork)` and `WithVolumeMount(Ivolume, string)`.

## Why is it important?

This makes it easier for developers to configure test environments and configurations that depend on other resources. We can also use it internally, for example, the WebDiver container configuration will be much simpler.

## Related issues

- Closes #826

## Follow-ups

- Remove (dispose) dependent resources. This is not super urgent, because the Resource Reaper will take care of it anyway.


tests/Testcontainers.Platform.Linux.Tests/DependsOnTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        .DependsOn(new ContainerBuilder()
+            .WithImage(CommonImages.Alpine)
+            .WithLabel(DependsOnKey, DependsOnValue)
+            .Build())
+        .WithNetwork(new NetworkBuilder()
+            .WithLabel(DependsOnKey, DependsOnValue)
+            .Build())
+        .WithVolumeMount(new VolumeBuilder()
+            .WithLabel(DependsOnKey, DependsOnValue)
+            .Build(), "/workdir")
```
HofmeisterAn: @cristianrgreco @eddumelendez @mdelapenya @kiview I am interested in what do you think about the following API. It resolves dependent resources internally and creates or starts them before starting the actual container. I am unsure whether we should also replace `WithNetwork(INetwork)` and `WithVolumeMount(IVolume, string)` with `DependsOn(INetwork)`, `DependsOn(IVolume, string)` to achieve a more consistent API. Developers no longer need to create resources upfront.

tests/Testcontainers.Platform.Linux.Tests/DependsOnTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        .Build();
+
+    public Task InitializeAsync()
+    {
+        return _container.StartAsync();
+    }
+
+    public Task DisposeAsync()
+    {
+        return _container.DisposeAsync().AsTask();
```
HofmeisterAn: @cristianrgreco @eddumelendez @mdelapenya @kiview Currently, dependent resources are not being removed when the parent container is disposed, as these resources might be shared across different configurations. Ryuk is responsible for cleaning them up.

My idea is to maintain an internal reference counter that increments with every create call. When the reference counter equals 1, we can remove the dependent resources as well (while disposing the parent). WDYT?

A general review would be appreciated. PR contains minor general improvements..

PR #833, opened 2023-03-13
feat: Replace CosmosDb module

AButler:
## What does this PR do?

Replaces the old CosmosDb module with the new module format.

## Why is it important?

Migrates one of the obsolete modules to the new format.

## Related issues

- Relates #493 

All the CosmosDb tests should pass


HofmeisterAn (commented):
Thank you for submitting the pull request! I was wondering if you know if any changes were made to the Cosmos DB image or the GH-hosted runners? Cosmos DB did not run properly or was flaky running on Ubuntu-22.04.

I will review your pull request tomorrow. Thank you once again for your contribution!

AButler:
Hi, unfortunately the emulator is still flaky on some runners due to certain CPUs. Hopefully MS will fix this soon. I still find this useful for Testcontainers since we use Testcontainers when running the tests locally and use a real cosmos container in the pipelines (at least until the issue is fixed).

Really hoping MS will fix it soon!

src/Testcontainers.CosmosDb/CosmosDbBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// <summary>
+    /// Sets the overridden IP address.
+    /// </summary>
+    /// <param name="ipAddress">The overridden IP address.</param>
+    /// <returns>A configured instance of <see cref="CosmosDbBuilder" />.</returns>
+    public CosmosDbBuilder WithIpAddressOverride(string ipAddress)
+    {
+        return Merge(DockerResourceConfiguration, new CosmosDbConfiguration(ipAddressOverride: ipAddress))
+            .WithEnvironment("AZURE_COSMOS_EMULATOR_IP_ADDRESS_OVERRIDE", ipAddress);
+    }
```
HofmeisterAn: Do we really need these properties? They are never used except in the `Validate()` member. Normally, we do not provide APIs for simple environment variables. Instead, developers can set them using 'WithEnvironment(k, v)'. It can be difficult to determine which environment variables are worth an API and which are not. We generally avoid creating a dedicated API if the value is only needed as an environment variable.

HofmeisterAn (commented):
The PR looks good. I just have a question about `PartitionCount` and `IpAddressOverride`.

src/Testcontainers.CosmosDb/CosmosDbBuilder.cs
AButler: Ah OK, probably not. I just migrated the important properties from the obsolete version. I can remove them.

AButler: I've removed these now 👍

Melethainiel:
Could it be possible to integrate the /Port configuration from last comment in #421 to avoid HttpClient exception ? Maybe something in the builder `.WithPort()` ?

AButler:
> Could it be possible to integrate the /Port configuration from last comment in #421 to avoid HttpClient exception ? Maybe something in the builder `.WithPort()` ?

If you use the `HttpClient` from the `CosmosDbBuilder` (see example in the test: https://github.com/testcontainers/testcontainers-dotnet/pull/833/files#diff-7c5e50ab4a8d9fd82823a579b5c9d3d61d52e6d13f706b54a8b3877acec2a45dR30) then it will automatically rewrite the host/port and accept the dev certificate.

Example:
```c#
var cosmosDbContainer = new CosmosDbBuilder().Build();

var cosmosClientOptions = new CosmosClientOptions {
  ConnectionMode = ConnectionMode.Gateway,
  HttpClientFactory = () => cosmosDbContainer.HttpClient
};

var cosmosClient = new CosmosClient(cosmosDbContainer.GetConnectionString(), cosmosClientOptions);
```

Melethainiel:
> > Could it be possible to integrate the /Port configuration from last comment in #421 to avoid HttpClient exception ? Maybe something in the builder `.WithPort()` ?
> 
> If you use the `HttpClient` from the `CosmosDbBuilder` (see example in the test: https://github.com/testcontainers/testcontainers-dotnet/pull/833/files#diff-7c5e50ab4a8d9fd82823a579b5c9d3d61d52e6d13f706b54a8b3877acec2a45dR30) then it will automatically rewrite the host/port and accept the dev certificate.
> 
> Example:
> 
> ```cs
> var cosmosDbContainer = new CosmosDbBuilder().Build();
> 
> var cosmosClientOptions = new CosmosClientOptions {
>   ConnectionMode = ConnectionMode.Gateway,
>   HttpClientFactory = () => cosmosDbContainer.HttpClient
> };
> 
> var cosmosClient = new CosmosClient(cosmosDbContainer.GetConnectionString(), cosmosClientOptions);
> ```

only if this is the 8081 port used. If not you have to rewrite the url with an [HttpMessageHandler](https://github.com/testcontainers/testcontainers-dotnet/issues/421#issuecomment-1318549493) or use the [/Port argument](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2784#issuecomment-940179747)

But the random port will likely not be 8081.

AButler:
> only if this is the 8081 port used. If not you have to rewrite the url with an [HttpMessageHandler](https://github.com/testcontainers/testcontainers-dotnet/issues/421#issuecomment-1318549493) or use the [/Port argument](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2784#issuecomment-940179747)

Yes, the `CosmosDbBuilder.HttpClient` already has the HttpMessageHandler to do the rewrite built in. Even if the port was randomly generated and passed through to the Cosmos emulator, you'd still need a message handler to rewrite the host and disable the certificate validation so not sure passing the port through adds anything.

Melethainiel:
> > only if this is the 8081 port used. If not you have to rewrite the url with an [HttpMessageHandler](https://github.com/testcontainers/testcontainers-dotnet/issues/421#issuecomment-1318549493) or use the [/Port argument](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2784#issuecomment-940179747)
> 
> Yes, the `CosmosDbBuilder.HttpClient` already has the HttpMessageHandler to do the rewrite built in. Even if the port was randomly generated and passed through to the Cosmos emulator, you'd still need a message handler to rewrite the host and disable the certificate validation so not sure passing the port through adds anything.

I don't know the magic behind :/ The only thing I know is that all the errors disappear when you set /Port=xxx and bind the docker port xxx to the host port xxx as in the second link I provided.

It's just a question, because with the last implementation I had the issue and ended by creating my own implementation with the /Port arg

HofmeisterAn:
> only if this is the 8081 port used. If not you have to rewrite the url with an [HttpMessageHandler](https://github.com/testcontainers/testcontainers-dotnet/issues/421#issuecomment-1318549493) or use the [/Port argument](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2784#issuecomment-940179747)
> 
> But the random port will likely not be 8081.

The container port will always be 8081. We resolve it to the random assigned host port her: https://github.com/testcontainers/testcontainers-dotnet/blob/d2ee73b77547cc1ca64a55f0f99ec7e0f972c706/src/Testcontainers.CosmosDb/CosmosDbContainer.cs#L27

I think the configuration is fine.

Melethainiel:
> The container port will always be 8081. We resolve it to the random assigned host port her:
> 
> https://github.com/testcontainers/testcontainers-dotnet/blob/d2ee73b77547cc1ca64a55f0f99ec7e0f972c706/src/Testcontainers.CosmosDb/CosmosDbContainer.cs#L27
> 
> I think the configuration is fine.

The problem is not the connection string. It's from having a random port assign to 8081.
Let me show what I mean with docker-compose example :

Here is the equivalent of what the builder will produce with a random port mapped to the 8081
```
version: '2'
services:
  cosmosdb:
    image: mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator
    container_name: cosmosdb
    ports:
      - 14559:8081 
```

And here is what it should be to avoid error

```
version: '2'
services:
  cosmosdb:
    image: mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator
    container_name: cosmosdb
    ports:
      - 14559:14559 
    environment:
      AZURE_COSMOS_EMULATOR_ARGS: /Port=14559
```
From [Microsoft doc](https://learn.microsoft.com/en-us/azure/cosmos-db/emulator-command-line-parameters#command-line) and the [issue](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2784#issue-1022495944) when both port are not the same.


And here with some code :

The cosmosClientBuilder which doesn't change
```csharp
  cosmosClientBuilder.WithHttpClientFactory(
                    () =>
                    {
                        HttpMessageHandler httpMessageHandler = new HttpClientHandler
                        {
                            ServerCertificateCustomValidationCallback =
                                HttpClientHandler.DangerousAcceptAnyServerCertificateValidator
                        };

                        return new HttpClient(httpMessageHandler);
                    })
                .WithConnectionModeGateway();
```

The exception I got when the port is mapped 8082:8081
```
System.IO.IOException:  Received an unexpected EOF or 0 bytes from the transport stream.
```

with port mapped 8082:8082 with the /Port argument all work smoothly

AButler:
In your above example, you are not using the HttpClient provided by the Testcontainers module. If you did then the code would be:

```c#
// Create test container
var cosmosDbContainer = new CosmosDbBuilder().Build();

cosmosClientBuilder
    .WithHttpClientFactory(() => cosmosDbContainer.HttpClient) // Use http client from Testcontainers module
    .WithConnectionModeGateway();
```

Melethainiel:
> In your above example, you are not using the HttpClient provided by the Testcontainers module. If you did then the code would be:
> 
> ```cs
> // Create test container
> var cosmosDbContainer = new CosmosDbBuilder().Build();
> 
> cosmosClientBuilder
>     .WithHttpClientFactory(() => cosmosDbContainer.HttpClient) // Use http client from Testcontainers module
>     .WithConnectionModeGateway();
> ```

It would imply to add the library as reference to the main project or pass down the httpclient which is not needed with the fix I suggest.

AButler:
> It would imply to add the library as reference to the main project and pass down the httpclient which is not needed with the fix I suggest.

No, as this is a test library then I would normally use dependency injection to inject the `CosmosClient`. It is not as though you would ever want the certificate validation disabled in the main project, nor for it to be forced to use gateway mode.

Melethainiel:
> No, as this is a test library then I would normally use dependency injection to inject the `CosmosClient`. It is not as though you would ever want the certificate validation disabled in the main project, nor for it to be forced to use gateway mode.

Nevermind, It was just a suggestion

HofmeisterAn (dismissed):

HofmeisterAn (dismissed):

HofmeisterAn (approved):
Sorry for the back and forth, I was trying to get the tests running on the Windows agent, but that is even more a mess. I think we need to skip the tests for now. I will merge it later in the day.

PR #837, opened 2023-03-17
feat: Add JanusGraph module

rngcntr:
## What does this PR do?

This PR adds a module for the Graph Database [JanusGraph](https://janusgraph.org).
For the moment, only the most basic configuration of JanusGraph is supported. 
Support for other backends can be added later. However, this simple configuration should be sufficient to for the majority of use cases.

## Why is it important?

JanusGraph is used by a variety of companies to build scalable graph database solutions. By providing a Testcontainers module, we lower the entry barrier for new developers to write meaningful tests.

## Follow-ups

- [ ] Wait for final 1.0.0 Release of JanusGraph
- [ ] Support different storage backends


rngcntr:
For some reason, the container is unable to start properly in the CI. I'll have to push a few commits to add some temporary code for debugging.

src/Testcontainers.JanusGraph/Testcontainers.JanusGraph.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+        <LangVersion>latest</LangVersion>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="JetBrains.Annotations" Version="2022.3.1" />
+        <PackageReference Include="JanusGraph.Net" Version="0.4.2" />
```
HofmeisterAn: We do not recommend and include client dependencies in modules. Otherwise, developers will run into dependency clashes (forced to use specific versions, etc.). We favor configurations without dependencies. The intention is nice, but can we remove it?

src/Testcontainers.JanusGraph/JanusGraphBuilder.cs
```diff
@@ -0,0 +1,70 @@
+namespace Testcontainers.JanusGraph;
+
+using System;
+using System.Net.WebSockets;
```
HofmeisterAn: Can we move these usings to the gloab usings file?

HofmeisterAn (changes requested):
Thank you so much for submitting this pull request! I've reviewed your changes and overall, everything looks great. However, I have a small change request regarding the 3rd party client dependencies.

src/Testcontainers.JanusGraph/Testcontainers.JanusGraph.csproj
rngcntr: I get your point. Would it make more sense to you to move such a dependency to the tests so that users can refer to the tests for usage examples?

HofmeisterAn: Absolute, that's what we usually do too.

rngcntr: Alright then, fixed it!

HofmeisterAn (dismissed):

src/Testcontainers.JanusGraph/JanusGraphBuilder.cs
```diff
@@ -0,0 +1,67 @@
+namespace Testcontainers.JanusGraph;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class JanusGraphBuilder : ContainerBuilder<JanusGraphBuilder, JanusGraphContainer, JanusGraphConfiguration>
+{
+    public const string JanusGraphImage = "janusgraph/janusgraph:0.6.3";
```
HofmeisterAn: `0.6.3` seems to require a different wait strategy, it never completes the readiness check.

rngcntr: I will have a look at it tomorrow.

rngcntr: It appears this is a previously unknown issue with JanusGraph 0.6.3: https://github.com/JanusGraph/janusgraph/issues/3681

Nothing serious, it's just the default log level that has changed. I'll get back to you once either
1) the issue with 0.6.3 is resolved, or
2) I have figured out an environment setting to manually set the log level, or
3) 1.0.0 is released

HofmeisterAn: The mentioned issue seems to be fixed. Do you know if there is an image we can use that works, so we can close this pull request?

rngcntr: Sorry for the late response, I was on vacation. Let's see if the CI succeeds when using 0.6.4

rngcntr: So it works with the latest pre-release, but the version that contains the fix is not yet publicly released.

HofmeisterAn: No, worries. Ok, thanks. Then let's wait until it's public available?

HofmeisterAn: It looks like there is a `1.0.0` release?

rngcntr: Yes! Thanks for the heads-up. I updated the version, let's see if the pipeline succeeds.
If everything looks good, I will also rebase the branch and then we're good to go

HofmeisterAn (approved):

PR #840, opened 2023-03-17
chore: Update Redpanda image

eddumelendez:
Currently, `RedpandaContainer` uses `docker.redpanda.com/vectorized/redpanda`
but must use `docker.redpanda.com/redpandadata/redpanda` instead.


HofmeisterAn (approved):
👍

Issue #841, opened 2023-03-19
[Enhancement]: Support identitytoken based authentication (OAuth2)


lbruun:
### Problem

Some containers registries use OAuth2 protocol for authentication. For example Azure Container Registry. 
After obtaining such credentials, Docker's `config.json` file look something like this:

```json
{
    "auths": {
        "myprivateregistry.azurecr.io": {
            "auth": "MDAwMDAwMDAtMDAwMC0wMDAwLTAwMDAtMDAwMDAwMDAwMDAwOg==",
            "identitytoken": "gwHEWT356ASfg....eUWR"
        }
    }
}
```
This is a use-case which testcontainers-dotnet cannot handle at the moment. 

With the above, testcontainers-dotnet will produce the following exception:
```text
Docker.DotNet.DockerApiException : Docker API responded with status code=InternalServerError, response=
{"message":"Head \"https://myprivateregistry.azurecr.io/v2/myapp/manifests/latest\": 
unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information."}
```

This is because testcontainers-dotnet completely ignores the `identitytoken`. It solely looks at the `auth` value and attempts to extract username/password from that.  In the above case the username is `00000000-0000-0000-0000-000000000000` and the password is empty string. It will use those values for authentication which will of course fail.
  
Note: The above config file works with the Docker CLI, but doesn't work with testcontainers-dotnet.

### Solution

Use the `identitytoken` value if present. Only try to extract username/password from the `auth` property if the `identitytoken` is _not_ present.

See https://docs.docker.com/engine/api/v1.42/#section/Authentication

The fix is trivial as Docker.DotNet library has full support for this use-case.

### Benefit

Support for Azure Container Registry.
(especially from a CI pipeline)

### Alternatives

No known workarounds.

It would actually work as-is if the CI pipeline had a credstore available, because testcontainers-dotnet actually [already support this use-case with identitytoken](https://github.com/testcontainers/testcontainers-dotnet/blob/581a3cd85cb2b9b33b2550b6a82bf4b3913677ce/src/Testcontainers/Configurations/AuthConfigs/DockerRegistryAuthenticationConfiguration.cs#L41). However, CI workflows typically do not have such facility available, hence the `docker login` command will put the identitytoken directly in the config file.

### Would you like to help contributing this enhancement?

Yes

lbruun:
Created PR #842  as a proposed fix.

[...]

PR #842, opened 2023-03-19
support identitytoken based authentication

lbruun:
This form of authentication is typically used by private registries which use OAuth2 authentication, for example Azure Container Registry (ACR). Specifically this PR fixes the situation where the identity token is given directly in the Docker `config.json` file.

## What does this PR do?

Proposes a fix for [issue 841](https://github.com/testcontainers/testcontainers-dotnet/issues/841)

Note: It is outside the scope of testcontainers library to _obtain_ the token.

Ref :  [Docker's documentation](https://docs.docker.com/engine/api/v1.42/#section/Authentication)

## Why is it important?

Support for Azure Container Registry from CI pipeline.

## Related issues

Relates #841

## How to test this PR

1. Have the Azure CLI installed.
1. Use a private Azure Container Registry. By private is meant one that doesn't allow un-authenticated pulls.
1. Login to Azure with an identity which is allowed to pull from that registry: `az login`
1. Perform implicit docker login:  `az acr login -name myprivateregistry`. This obtains a 3 hour token and uses the `docker` CLI to write the identitytoken into your local Docker config file. If you are doing this on a host which has a credstore available then "unfortunately" the Docker CLI is clever enough to use that. This is not good, as then you'll be testing something which testcontainers-dotnet already knows how to handle. The workaround for that is to do as follows:
```bash
TOKEN=$(az acr login --name myprivateregistry --expose-token --output tsv --query accessToken)
docker login myprivateregistry.azurecr.io --username 00000000-0000-0000-0000-000000000000 --password $TOKEN
```
4. Look in your `~/.docker/config.json` and verify that is has an explicit `identitytoken`.
5. Do the test.

## Implementation notes

This PR simply enhances on the  `Base64Provider` class. This is a little bit quick-and-dirty as technically identitytoken-based authentication has nothing to do with base64, so the name is slightly misleading for the identitytoken use-case. It may be "cleaner" to create a new provider class, e.g. `IdentityTokenProvider.cs`, although it would share 90% of its code with the existing `Base64Provider`.


src/Testcontainers/Builders/Base64Provider.cs
```diff
@@ -62,11 +62,26 @@ public IDockerRegistryAuthenticationConfiguration GetAuthConfig(string hostname)
...
+      {
+        if (JsonValueKind.String.Equals(identitytoken.ValueKind))
+        {
+          this.logger.DockerRegistryCredentialFound(hostname);
+          return new DockerRegistryAuthenticationConfiguration(authProperty.Name, null, null, identitytoken.GetString());
+        }
+      }
+
+      // ... otherwise we expect the 'auth' property to contain the username and password
+      // in base64 encoded form and separated by a colon char.
```
HofmeisterAn, line 83:
```suggestion
      if (JsonValueKind.Undefined.Equals(authProperty.Value.ValueKind))
      {
        return null;
      }

      if (authProperty.Value.TryGetProperty("identityToken", out var identityToken) && !string.IsNullOrEmpty(identityToken.GetString()))
      {
        _logger.DockerRegistryCredentialFound(hostname);
        return new DockerRegistryAuthenticationConfiguration(authProperty.Name, null, null, identityToken.GetString());
      }
```

Can you please cover the changes with [tests](https://github.com/testcontainers/testcontainers-dotnet/blob/710987a2562f7bb5a58d64bb74cba17cc8681db4/tests/Testcontainers.Tests/Unit/Configurations/DockerRegistryAuthenticationProviderTest.cs#L63).

HofmeisterAn:
Due to inactivity, I will be closing this pull request. However, please do not hesitate to reopen it again if you decide to resume working on it.

Issue #847, opened 2023-03-23
[Bug]: Update to 3.0.0 breaks runing test container inside other containers


lukasswm:
### Testcontainers version

3.0.0

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host arch

x86

### .NET version

7.0.202

### Docker version

```shell
Client:
 Cloud integration: v1.0.31
 Version:           20.10.23
 API version:       1.41
 Go version:        go1.18.10
 Git commit:        7155243
 Built:             Thu Jan 19 17:43:10 2023
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.17.1 (101757)
 Engine:
  Version:          20.10.23
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.10
  Git commit:       6051f14
  Built:            Thu Jan 19 17:32:04 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.18
  GitCommit:        2456e983eb9e37e47538f59ea18f2043c9a73640
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.10.3)
  compose: Docker Compose (Docker Inc., v2.15.1)
  dev: Docker Dev Environments (Docker Inc., v0.1.0)
  extension: Manages Docker extensions (Docker Inc., v0.2.18)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.25.0)
  scout: Command line tool for Docker Scout (Docker Inc., v0.6.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 3
 Server Version: 20.10.23
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #852, opened 2023-03-27
feat: Always use Docker's assigned random host port (Ryuk)

HofmeisterAn:
## What does this PR do?

This PR extends the .NET documentation and explains that the Docker Wormhole configurations should use Docker's raw socket due to a bug in Docker Desktop. Furthermore, it sets the `ResourceReaperPublicHostPort` delegate obsolete. The Resource Reaper no longer uses the API to configure the assigned host port.

## Why is it important?

The PR provides more information about the Docker Wormhole configuration and use the same port assignment everywhere. The Resource Reaper was the only exception that uses the `0` assignments instead of `string.Empty`.

## Related issues

- Relates #847
- Relates https://github.com/docker/for-mac/issues/5588#issuecomment-934600089

### DIFF
--- a/docs/examples/dind.md
+++ b/docs/examples/dind.md
@@ -1,6 +1,16 @@
-# Compose
+# Running inside another container
 
 ## Docker Desktop
+
+### Sibling containers
+
+If you choose to run your tests in a Docker Wormhole configuration, which involves using sibling containers, it is necessary to mount Docker's raw socket `/var/run/docker.sock.raw`. You find more information and an explanation of the bug in this [comment](https://github.com/docker/for-mac/issues/5588#issuecomment-934600089).
+
+```console
+docker run -v /var/run/docker.sock.raw:/var/run/docker.sock $IMAGE dotnet test
+```
+
+### Compose
 
 To use Docker's Compose tool to build and run a Testcontainers environment in a Docker Desktop Wormhole configuration,
 it is necessary to override Testcontainers' Docker host resolution and set the environment variable `TESTCONTAINERS_HOST_OVERRIDE` to `host.docker.internal`.
--- a/docs/examples/dind.md
+++ b/docs/examples/dind.md
@@ -4,7 +4,7 @@
 
 ### Sibling containers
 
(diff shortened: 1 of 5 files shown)


src/Testcontainers/Containers/ResourceReaper.cs
```diff
@@ -61,8 +61,7 @@ private ResourceReaper(Guid sessionId, IDockerEndpointAuthenticationConfiguratio
         .WithPrivileged(requiresPrivilegedMode)
         .WithAutoRemove(true)
         .WithCleanUp(false)
-        .WithExposedPort(RyukPort)
-        .WithPortBinding(TestcontainersSettings.ResourceReaperPublicHostPort.Invoke(dockerEndpointAuthConfig), RyukPort)
```
HofmeisterAn: @kiview Should we remove the delegate assignment immediately? Or should we introduce the change after the next release?

docs/examples/dind.md
```diff
@@ -1,7 +1,17 @@
-# Compose
+# Running inside another container
 
 ## Docker Desktop
 
+### Sibling containers
+
+If you choose to run your tests in a Docker Wormhole configuration, which involves using sibling containers, it is necessary to mount Docker's raw socket `/var/run/docker.sock.raw`. You find more information and an explanation of the bug in this [comment](https://github.com/docker/for-mac/issues/5588#issuecomment-934600089).
```
HofmeisterAn:
```suggestion
If you choose to run your tests in a Docker Wormhole configuration, which involves using sibling containers, it is necessary to mount Docker's raw socket `/var/run/docker.sock.raw`. You find more information and an explanation of the Docker bug in this [comment](https://github.com/docker/for-mac/issues/5588#issuecomment-934600089).
```

src/Testcontainers/Containers/ResourceReaper.cs
kiview: Are there any potential risks or potentially breaking behaviour involved in removing it? If you don't see any bigger risks, I think it's fine to remove it for this release already 👍

kiview (dismissed):

src/Testcontainers/Containers/ResourceReaper.cs
HofmeisterAn: I have added the old code again, which replaces `0` with `string.Empty`. This should behave as intended. In case someone is using the delegate and encounters problems after removing it, we will not break the entire project. They can re-add it. I believe that being cautious in this situation is a bit better. However, I do not expect any issues, as this delegate was introduced in a very early version.

PR #860, opened 2023-03-31
Dapr Testcontainers

olitomlinson:
## What does this PR do?

Introduce a Testcontainers for Dapr https://dapr.io/

## Why is it important?

Dapr users can benefit from a Dapr-as-a-Testcontainers for use in integration testing.


src/Testcontainers.Dapr/DaprBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+            .WithCommand("--dapr-http-port", daprHttpPort.ToString())
+            .WithPortBinding(daprHttpPort, true);
+    }
+
+    public DaprBuilder WithDaprGrpcPort(int daprGrpcPort)
+    {
+        return Merge(DockerResourceConfiguration, new DaprConfiguration())
+            .WithCommand("--dapr-grpc-port", daprGrpcPort.ToString())
+            .WithPortBinding(daprGrpcPort, true);
+    }
```
HofmeisterAn: Could you please explain why it's necessary to configure container ports? Why can't we simply use the default ports? In our experience, configuring container ports doesn't provide any benefit, as Testcontainers maps container ports to random public host ports anyway.

I recommend removing those APIs and relying on the default ports like all the other modules.

src/Testcontainers.Dapr/DaprBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+        Validate();
+
+        return new DaprContainer(DockerResourceConfiguration, TestcontainersSettings.Logger);
+    }
+
+    protected override DaprBuilder Init()
+    {       
+        return base.Init()
+            .WithImage(DaprImage)
+            .WithCommand("./daprd")
```
HofmeisterAn:
```suggestion
            .WithEntrypoint("./daprd")
```

src/Testcontainers.Dapr/DaprBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+            .WithPortBinding(DaprGrpcPort, true)
+            .WithWaitStrategy(Wait.ForUnixContainer().AddCustomWaitStrategy(new WaitUntilLogIsFound()))
+            .WithLogLevel(LogLevel);
+    }
+
+    protected override void Validate()
+    {
+         base.Validate();
+
+         _ = Guard.Argument(DockerResourceConfiguration.AppId, nameof(DockerResourceConfiguration.AppId))
```
HofmeisterAn: Does it make sense to provide a default app id?

src/Testcontainers.Dapr/DaprBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    protected override DaprBuilder Merge(DaprConfiguration oldValue, DaprConfiguration newValue)
+    {
+        return new DaprBuilder(new DaprConfiguration(oldValue, newValue));
+    }
+
+    private sealed class WaitUntilLogIsFound : IWaitUntil
+    {    
+        // this is horrific, but it works for now...
```
HofmeisterAn:
```suggestion
```

What is the reason for that? It is not uncommon to wait for a log message.

src/Testcontainers.Dapr/DaprConfiguration.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public DaprConfiguration(DaprConfiguration oldValue, DaprConfiguration newValue)
+        : base(oldValue, newValue)
+    {
+        AppId = BuildConfiguration.Combine(oldValue.AppId, newValue.AppId);
+    }
+
+    public string AppId { get; }
+    public string LogLevel { get; }
+    public int AppPort { get; set; }
```
HofmeisterAn: You are not using these properties in any other context except in the builder. It is not necessary to keep them in the configuration class unless you require access to the configured values, such as in the container implementation.

tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        return _daprContainer.DisposeAsync().AsTask();
+    }
+
+    [Fact]
+    [Trait(nameof(DockerCli.DockerPlatform), nameof(DockerCli.DockerPlatform.Linux))]
+    public async Task CheckDaprSideCarIsHealthy()
+    {            
+        Environment.SetEnvironmentVariable("DAPR_HTTP_PORT", _daprContainer.DaprHttpPort.ToString());
+        Environment.SetEnvironmentVariable("DAPR_GRPC_PORT", _daprContainer.DaprGrpcPort.ToString());
```
HofmeisterAn: Can we please provide a connection string from `DaprContainer`? The host is not always `localhost`. Depending on the Docker environment, it could be something else.

```csharp
_ = new UriBuilder(Uri.UriSchemeHttp, Hostname, GetMappedPublicPort(DaprBuilder.DaprHttpPort)).ToString();
_ = new UriBuilder(Uri.UriSchemeHttp, Hostname, GetMappedPublicPort(DaprBuilder.DaprGrpcPort)).ToString();
```

tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        var _daprClient = new DaprClientBuilder().Build();
+        var healthy = await _daprClient.CheckHealthAsync();
+
+        Assert.True(true);
```
HofmeisterAn: While this assert is true 😀, it does not make sense.

tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        //use this to hold the container in an active state for debugging
+        //await Task.Delay(TimeSpan.FromHours(1));
```
HofmeisterAn:
```suggestion
```

HofmeisterAn (changes requested):
Thank you for taking the time to submit the Dapr module. We appreciate your contribution to the project.

I kindly request a few changes. Could you please remove any changes that do not belong to the Dapr module? Revert changes to `certificate.crt`, `devcontainer.json`, `strongname.snk` and remove the VS Code configuration. Once these changes incl. the comments below are addressed, I would be happy to merge the module.

src/Testcontainers.Dapr/DaprBuilder.cs
olitomlinson: > Thank you for taking the time to submit the Dapr module. We appreciate your contribution to the project.
> 
> I kindly request a few changes. Could you please remove any changes that do not belong to the Dapr module? Revert changes to `certificate.crt`, `devcontainer.json`, `strongname.snk` and remove the VS Code configuration. Once these changes incl. the comments below are addressed, I would be happy to merge the module.

Thanks for the comments.

I raised this as a 'draft' PR as I know there is a bit of Junk in there, but I just wanted to get early eyes on it from a maintainer to see If I was going in the right direction. Will absolutely clean out the junk before I mark the PR as 'ready for review'

> Could you please explain why it's necessary to configure container ports? Why can't we simply use the default ports? In our experience, configuring container ports doesn't provide any benefit, as Testcontainers maps container ports to random public host ports anyway. I recommend removing those APIs and relying on the default ports like all the other modules.

Yeah, I think you're right.

P.s. I'm still learning about TestContainers as I go through this, so it's quite possible I've got some of the fundamentals wrong :)

src/Testcontainers.Dapr/DaprBuilder.cs
olitomlinson: Hmmm no, I don't think it does.

The dapr app-id is not really a runtime thing. Apps depend on it being well known. 

image

src/Testcontainers.Dapr/DaprBuilder.cs
olitomlinson: Depending on a string is a bit of a flaky API IMO, so I wouldn't want to keep this.

I have had success using a [probe](https://github.com/olitomlinson/dapr-pluggable-state-store/blob/a0ca9843e678f816e40503a749e918d50509bdae/IntegrationTests/PluggableContainer.cs#L63), when waiting for the dapr process to become initialised, but I'm having trouble making it work when trying to build a TestContainer for dapr in this instance, so I've parked it for now. 

Will definitely be revisiting this though.

src/Testcontainers.Dapr/DaprBuilder.cs
HofmeisterAn: > I'm still learning about TestContainers as I go through this, so it's quite possible I've got some of the fundamentals wrong :)

No worries. If you have any questions or issues with configuring something, please let me know. You can contact me either here or on Slack.

src/Testcontainers.Dapr/DaprConfiguration.cs
olitomlinson: Ok I will revisit.

tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs
olitomlinson: Hmmmm. 

I'll think on this, as I think the Dapr Client library is hardcoded to always use `localhost`

https://github.com/dapr/dotnet-sdk/blob/b97af45c5f2689ecfd85246b5954b6bc2f528e07/src/Shared/DaprDefaults.cs#L79

HofmeisterAn: The `DaprClientBuilder` offers `UseHttpEndpoint(string)` and `UseGrpcEndpoint(string)`, both receives an URI endpoint. I guess that's the right way to configure the Daper client. I would be surprised if only `localhost` is supported (also only IPv4).

src/Testcontainers.Dapr/DaprBuilder.cs
HofmeisterAn: Ok, I see. Indeed, that might be a bit better. If you need any help, let me know.

tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs
olitomlinson: Dapr is mostly used in a kubernetes environments, where each k8s service has its own dapr sidecar within the pod, so overriding the host (`127.0.0.1`) is quite rare/unnecessary tbh.

However, sure, I can include the connection string for posterity.

olitomlinson:
Hey @HofmeisterAn  I've got a bit stuck, and I would appreciate a pointer.

So, my `dapr` container can't reach the port of another container, and I'm not sure why.

The `dapr` container depends on being able to reach the `nginx` container port (`58768`), but it is getting blocked somewhere. My guess is its a Docker network issue, but unsure how to verify this.

Pasted Graphic

I can see `dapr` is attempting to connect on the correct port, with this log from the `dapr` `stdout`  :

> 2023-04-01 22:33:50 time="2023-04-01T21:33:50.252819841Z" level=info msg="application protocol: http. waiting on port **58768**.  This will block until the app is listening on that port." app_id=testicorns instance=be392285d04b scope=dapr.runtime type=log ver=1.10.4

I can curl against the `nginx` port fine from my host machine, this is why i think its a networking issue between the `dapr` container and the `nginx` container.

`Dapr` will be calling the `nginx` container via `127.0.0.1:58768` as [this code implies](https://github.com/dapr/dapr/blob/4abbf68000379fc458b25dc48094d7e2817cd868/pkg/runtime/runtime.go#L2910)

I've included both `nginx` container and `dapr` container on the same network, and used the `GetMappedPublicPort` which I assume is configured correctly https://github.com/olitomlinson/testcontainers-dotnet-dapr/blob/af41f6d7325638a603671bead36f6cff0eac16d1/tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs#L50

Any pointers on how I can find out why the `nginx` port can't be reached would be much appreciated, as this is blocking my progress.

Many thanks!

tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs
HofmeisterAn: I've addressed this in my latest changes:

https://github.com/testcontainers/testcontainers-dotnet/blob/bf4bcbb32758d90113ea76087c80088d0e8815f0/src/Testcontainers.Dapr/DaprContainer.cs#L17-L26

It is important to understand that connections from the **test host** requires to resolve `Hostname` and the random assigned host port (`GetMappedPublicPort(ushort))`. While a container-to-container communication does not. They can simply use the network alias and port to connect to (**note**: they connect straight to the container port).

HofmeisterAn:
> `Dapr` will be calling the `nginx` container via `127.0.0.1:58768` as [this code implies](https://github.com/dapr/dapr/blob/4abbf68000379fc458b25dc48094d7e2817cd868/pkg/runtime/runtime.go#L2910)

Is this really used by `daprd`? If `daprd` waits until an HTTP port opens on `127.0.0.1` it cannot work. Nginx is running in a different container. Usually, container-to-container communication relies on the network alias (or hostname of the container). Considering my changes:

https://github.com/testcontainers/testcontainers-dotnet/blob/bf4bcbb32758d90113ea76087c80088d0e8815f0/tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs#L9-L30

`_daprContainer` can access `nginxContainer` via `http://nginx:80`.

olitomlinson:
> > `Dapr` will be calling the `nginx` container via `127.0.0.1:58768` as [this code implies](https://github.com/dapr/dapr/blob/4abbf68000379fc458b25dc48094d7e2817cd868/pkg/runtime/runtime.go#L2910)
> 
> Is this really used by `daprd`? If `daprd` waits until an HTTP port opens on `127.0.0.1` it cannot work. Nginx is running in a different container. Usually, container-to-container communication relies on the network alias (or hostname of the container). Considering my changes:
> 
> https://github.com/testcontainers/testcontainers-dotnet/blob/bf4bcbb32758d90113ea76087c80088d0e8815f0/tests/Testcontainers.Dapr.Tests/DaprContainerTest.cs#L9-L30
> 
> `_daprContainer` can access `nginxContainer` via `http://nginx:80`.

> Usually, container-to-container communication relies on the network alias (or hostname of the container)

Usually this is true, however dapr is largely used in kubernetes, the app & dapr sidecar both reside in the same pod, so communicating via `localhost:<port>` works fine.

But agree, this use-case is currently blocked because daprd will only ever call out to an App via `127.0.0.1`.

https://github.com/dapr/dapr/issues/6177

src/Testcontainers.Dapr/DaprBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+        return Merge(DockerResourceConfiguration, new DaprConfiguration(appId: appId))
+            .WithCommand("--app-id", appId);
+    }
+
+    public DaprBuilder WithAppPort(int appPort)
+    {
+        // TODO: What happens if developers call WithAppId(string) multiple times?
+        return Merge(DockerResourceConfiguration, new DaprConfiguration())
+            .WithCommand("--app-port", appPort.ToString());
+    }
```
HofmeisterAn, line 37: Do you think it makes sense to put the app information in a dedicated type?

src/Testcontainers.Dapr/DaprBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+        // TODO : Introduce an Enum for logLevel values.
+        return Merge(DockerResourceConfiguration, new DaprConfiguration(logLevel: logLevel))
+            .WithCommand("--log-level", logLevel);
+    }
+
+    public DaprBuilder WithAppChannelAddress(string appChannelHost)
+    {
+        // TODO : Introduce an Enum for logLevel values.
+        return Merge(DockerResourceConfiguration, new DaprConfiguration(appChannelAddress: appChannelHost))
+            .WithCommand("--app-channel-address", appChannelHost);
```
HofmeisterAn, line 50: Is `--app-channel-address` a new CLI switch?

src/Testcontainers.Dapr/DaprBuilder.cs
```diff
@@ -0,0 +1,101 @@
+namespace Testcontainers.Dapr;
+
+[PublicAPI]
+public sealed class DaprBuilder : ContainerBuilder<DaprBuilder, DaprContainer, DaprConfiguration>
+{
+    public const string DaprImage = "daprio/daprd:nightly-2023-04-28";
```
eddumelendez, line 6: Let's not forget to update the version :D

HofmeisterAn:
Your PR hasn't seen any recent activities, I will close it. Please, do not hesitate to reopen it again if you continue to work on it.

PR #865, opened 2023-04-03
Feature/862 add spanner module

### LINKED ISSUE #862: [Enhancement]: Add spanner (emulator) module
### Problem

We are runnign dontnet services on GCP, where spanner is our database service of choice for some services. Running integration tests on this, controlled from unit test style code is part of our process. Test containers is ideal for this, however so far there is no spanner module available. 

### Solution

We created a testcontainers.spanner module for internal use and like to contribute it back to the community.
We like to create a feature branch on this, make the adoptions needed to make it an official module and integrate in in the testcontainers CI and ultimately raise a pull reqwuest for this.

The module now contains these functionalities:
- On the builder 
-- WithInstanceId
-- WIthProjectId 
-- WithDabaseId
- On the SpannerContainer class
-- Expose the mapped port for grpc
-- Expose the port mapped for rest
-- expose the connectionstring
-- On start set required environmentvariable in dotnet for connecting to a spanner emulator
-- On start create the project and database on the emulator instanc
-- Expose an execute ddl method to simplify defining the datamodel

### Benefit

Other developers in need of testing locally or in CI processes against a spanner will need less code and time to set up their testcontainer. The testcontainers module set will be bigger, increasing the maturity of the suite.

### Alternatives

No alternatives that I'm aware of

Yes

TwanVB:
## What does this PR do?
Add spanner (emulator) module:
- Created a new module based on the template
- Chose to use rest for internal setup of the container as to not create dependency issues when a client under test would depend upon a different version of the google spanner client Google.Cloud.Spanner.Data

## Why is it important?
Other developers in need of testing locally or in CI processes against a spanner will need less code and time to set up their testcontainer. The testcontainers module set will be bigger, increasing the maturity of the suite.

## Related issues

- Closes #862


src/Testcontainers.Spanner/.editorconfig
```diff
@@ -0,0 +1,241 @@
+root = true
```
HofmeisterAn, line 1: Could we please replace the configuration in the `.editorconfig` file with the configuration used by the other modules and stick to a plain code style (same as all the other modules)? Not all IDEs respect these files or configurations. While the indentation is nice, experience has shown that it does not work well in OSS projects.

src/Testcontainers.Spanner/SpannerBuilder.cs
```diff
@@ -0,0 +1,126 @@
+namespace Testcontainers.Spanner;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class SpannerBuilder : ContainerBuilder<SpannerBuilder, SpannerContainer, SpannerConfiguration>
+{
+  private const string DefaultProjectId = "my-project";
+  private const string DefaultInstanceId = "my-instance";
+  private const string DefaultDatabaseId = "my-database";
+  private const string SpannerEmulatorImage = "gcr.io/cloud-spanner-emulator/emulator";
```
HofmeisterAn: Please pin the version.

src/Testcontainers.Spanner/SpannerContainer.cs
```diff
@@ -0,0 +1,153 @@
...
+using Testcontainers.Spanner.RestApi.v1;
+
+namespace Testcontainers.Spanner;
+
+/// <inheritdoc cref="DockerContainer" />
+[PublicAPI]
+public sealed class SpannerContainer : DockerContainer, IDisposable
+{
+  internal const int InternalGrpcPort = 9010;
+  internal const int InternalRestPort = 9020;
```
HofmeisterAn: Please move the default ports to the builder (like the other modules are configured).

src/Testcontainers.Spanner/SpannerContainer.cs
```diff
@@ -0,0 +1,153 @@
...
+      .ValidateAsync($"create {resourceDescription}", ct: ct);
+  }
+
+  public void Dispose() => _webClient.Dispose();
+
+  protected override async ValueTask DisposeAsyncCore()
+  {
+    _webClient.Dispose();
+    await base.DisposeAsyncCore();
+  }
```
HofmeisterAn: Are these configurations at all necessary to start a Spanner container? Usually, we do not include client tasks into Testcontainers. Developers should use the corresponding client dependency. It looks like Spanner has a dedicated dependency to create an [instance](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.admin.instance.v1#instanceadmin) or [database](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.Spanner.Data/latest#creating-a-database-and-table)?

> The [Google.Cloud.Spanner.Admin.Instance.V1](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.Spanner.Admin.Instance.V1/latest) package should be used for Cloud Spanner instance administration, such as creating or deleting instances.

If necessary at all, can we move the container configuration to `WithStartupCallback`, e.g. like [Couchbase](https://github.com/testcontainers/testcontainers-dotnet/blob/629aedac0559346bc954c1c59881d4b75a8481cf/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#LL145C14-L145C33)? We try to configure the entire container until it is ready in the builder.

src/Testcontainers.Spanner/SpannerContainer.cs
```diff
@@ -0,0 +1,153 @@
...
+  private async Task AwaitOperationAsync(SpannerOperationDto operation)
+  {
+    int delayInSeconds = 1;
+    const int maxDelayInSeconds = 10;
+
+    while (!operation.Done)
+    {
+      await Task.Delay(TimeSpan.FromSeconds(delayInSeconds));
+      string operationName = operation.Name;
+      var operationResponse = await _webClient.GetAsync($"v1/{operationName}").ValidateAsync("get operation");
```
HofmeisterAn: You can use the `HttpWaitStrategy` to [wait](https://github.com/testcontainers/testcontainers-dotnet/blob/629aedac0559346bc954c1c59881d4b75a8481cf/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#L183-L184) for a specific HTTP response.

src/Testcontainers.Spanner/Testcontainers.Spanner.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+        <LangVersion>latest</LangVersion>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="JetBrains.Annotations" Version="2022.3.1"/>
+        <PackageReference Include="System.Net.Http.Json" Version="7.0.1" />
```
HofmeisterAn: We try to avoid 3rd-party dependencies under all circumstances in our modules. Otherwise, developers may run into dependency clashes. The Couchbase module is a good example of how to deal with HTTP endpoints to configure a service during the container start. I suggest to follow a similar implementation.

tests/Testcontainers.Spanner.Tests/.editorconfig
```diff
@@ -0,0 +1,241 @@
+root = true
```
HofmeisterAn, line 1: Here applies the same as for the `.editorconfig` in `src`.

tests/Testcontainers.Spanner.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1 @@
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
HofmeisterAn: Why should we disable parallelization?

HofmeisterAn (changes requested):
Hi Twan, thank you for submitting the pull request for the Testcontainers Spanner module. I appreciate the effort you've put into it so far.

However, I have some suggestions that I believe could help improve the module implementation. Specifically, I noticed that there are some parts of the implementation that are not aligned with the other modules.

Additionally, I am wondering if the code that creates instances, databases, etc. (in `StartAsync(CancellationToken)`) is necessary for a successful container start (running the Spanner service). It seems like we are including tasks that should be the responsibility of the `Google.Cloud.Spanner.*` NuGet dependencies. This relates to https://github.com/testcontainers/testcontainers-dotnet/pull/865#discussion_r1157322554.

Could you please provide some clarification on whether this code is necessary, or if it can be removed to make the module implementation more streamlined and consistent with other Testcontainers modules?

src/Testcontainers.Spanner/SpannerBuilder.cs
TwanVB: will do so

src/Testcontainers.Spanner/.editorconfig
TwanVB: will do so

src/Testcontainers.Spanner/SpannerContainer.cs
TwanVB: ok

src/Testcontainers.Spanner/SpannerContainer.cs
TwanVB: It was inspired by what I saw in the postgresql, where you can include a script for initial setup. In our use cases it made sense to always have this available, but for m ore general purpose i can see it malkes sense to also have the emulator running without this, so I'll remove it from startup.

tests/Testcontainers.Spanner.Tests/.editorconfig
TwanVB: will adapt this

src/Testcontainers.Spanner/Testcontainers.Spanner.csproj
TwanVB: I understand the point of that. For my implementation i considered this to be ok as in this case as it is a package owned by microsoft and it's .net framework team, but of course the issue can still occur then. I have no problem changing this to an implementation without this nuget package, and will adapt.

tests/Testcontainers.Spanner.Tests/Properties/AssemblyInfo.cs
TwanVB: There is a nasty thing in the implementation for the spanner .net client that requires an environment variable to be set in case of using the emulator: https://cloud.google.com/spanner/docs/emulator#cs 
So when running multiple tests in parrallel, they override eachother environment variable. 
So this environment variable is now also set in the startup, As i intended to hide this emulator specific from the client, i want to do a little investigation if there is a better place/way.

src/Testcontainers.Spanner/SpannerContainer.cs
TwanVB: It is now also used by the execute ddl, but if you prefer not to have anything in here that can also be done with a client library, we should remove that part i think.

tests/Testcontainers.Spanner.Tests/Properties/AssemblyInfo.cs
TwanVB: To be more specific the detection setting in the string will make the client require this environment variable, it will not take the one from the connection string.

src/Testcontainers.Spanner/SpannerContainer.cs
HofmeisterAn: > It was inspired by what I saw in the postgresql, where you can include a script for initial setup.

But this is part (supported) by the PostgreSQL image, right? As previously mentioned, while the intention is nice, we try to avoid incorporating client-side tasks into modules. This is because it does not make sense for us to maintain vendor-specific functionalities when there is already a suitable client that can handle them. Instead, we strive to offer best practice configurations for common test use cases.

src/Testcontainers.Spanner/Testcontainers.Spanner.csproj
HofmeisterAn: Yes, the package is generally fine. However, it may potentially force other developers to also use the specified version (7.0.1) in their own projects, leading to version conflicts. By avoiding "unnecessary" dependencies in modules, we can prevent developers from encountering this issue.

tests/Testcontainers.Spanner.Tests/Properties/AssemblyInfo.cs
HofmeisterAn: Oh, that is unfortunate. Is there an upstream issue? It seems weird that we are unable to connect to multiple instances. Is it possible that we could achieve it by configuring a session pool manager (https://github.com/googleapis/google-cloud-dotnet/issues/4970#issuecomment-631376145)?

TwanVB: As you can see from the failing tests on ubuntu only, there is a difference in how the dependency on the environment variable is treated in linux and windows, I'll need to gain a more thorough understanding of this before completing this so I'll be trying to find resources/ threads /issues or source code reference on this first.

HofmeisterAn: What do you think about utilizing an xUnit.net [collection fixture](https://xunit.net/docs/shared-context) instead of disabling parallelization? This creates a single Cloud Spanner instance for the tests. It would not be necessary to update the environment variable anymore.

HofmeisterAn:
Your PR hasn't seen any recent activities, I will close it. Please, do not hesitate to reopen it again if you continue to work on it.

PR #881, opened 2023-05-05
fix: Remove trailing slash from rootless Docker socket path

HofmeisterAn:
## What does this PR do?

The pull request removes a trailing slash that prevents Testcontainers for .NET to auto-detect rootless Docker environment.

## Why is it important?

It looks like the [UriBuilder](https://learn.microsoft.com/de-de/dotnet/api/system.uribuilder) appends a trailing slash to the Docker daemon socket path which breaks the auto discovery.

## Related issues

\-

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,7 +5,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <PackageId>$(AssemblyName)</PackageId>
-    <Version>3.1.0</Version>
+    <Version>3.1.1</Version>
     <AssemblyVersion>$(Version)</AssemblyVersion>
     <FileVersion>$(Version)</FileVersion>
     <Product>Testcontainers</Product>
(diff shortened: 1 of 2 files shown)


HofmeisterAn:
I have double-checked the configuration for Docker Desktop on macOS and noticed that TC is no longer working properly after the Homebrew update. Is this setting new? It looks like another attempt, I think we saw something similar in previous releases.

236433326-256ea545-a2f8-43e6-bcb6-309eefeaf4d7

Since we extract the Docker daemon socket path from the endpoint and mount it into the Ryuk container, TC cannot start Ryuk anymore. I am wondering whether we should stop extracting the Docker daemon socket path for now and use `/var/run/docker.sock` instead. Or keep it and tell users to enable the setting? WDYT @cristianrgreco @eddumelendez @kiview @mdelapenya?

**Workaround**, more information about the [symlink](https://docs.docker.com/desktop/mac/permission-requirements/#installing-symlinks). The behavior has changed in 4.18.

    sudo ln -sf "$HOME/.docker/run/docker.sock" "/var/run/docker.sock"

cristianrgreco:
@HofmeisterAn I will take a proper look soon, but I think it's related to https://github.com/testcontainers/testcontainers-java/issues/6426#issuecomment-1535473398

HofmeisterAn:
> but I think it's related to [testcontainers/testcontainers-java#6426 (comment)](https://github.com/testcontainers/testcontainers-java/issues/6426#issuecomment-1535473398)

Yep, it does.

PR #885, opened 2023-05-09
feat: Add TC host strategy

HofmeisterAn:
## What does this PR do?

Adds a new strategy for auto-discovering container runtimes.

## Why is it important?

?

## Related issues

\-


src/Testcontainers/Builders/TestcontainersCloudEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// <inheritdoc />
+    public Uri GetDockerHost()
+    {
+      return _customConfiguration.GetDockerHost();
+    }
+
+    /// <inheritdoc />
+    public string GetDockerHostOverride()
+    {
+      return _customConfiguration.GetDockerHostOverride();
```
cristianrgreco, line 71: I see `_customConfiguration` implements  `PropertiesFileConfiguration`, does this also take into account environment variable configurations?

HofmeisterAn: No, it does not. It purely relies on the properties file. I asked myself the same question, if it should. Until now, TC for .NET use either of both ways to configure TC.

kiview: I don't fully understand the question @cristianrgreco. 

We effectively want `TestcontainersHostConfiguration` to only be configured through the `testcontainers.host` property in the property file, right? And this implementation follows this and hence is correct I assume.

src/Testcontainers/Builders/TestcontainersCloudEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+{
+  using System;
+  using System.Text.Json;
+  using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Images;
+  using JetBrains.Annotations;
+
+  /// <inheritdoc cref="IDockerRegistryAuthenticationProvider" />
+  [PublicAPI]
+  internal sealed class TestcontainersHostEndpointAuthenticationProvider : DockerEndpointAuthenticationProvider, ICustomConfiguration
```
kiview, line 19:
```suggestion
  internal sealed class TestcontainersCloudEndpointAuthenticationProvider : DockerEndpointAuthenticationProvider, ICustomConfiguration
```

To be in line with the file name, right?

src/Testcontainers/Builders/TestcontainersCloudEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+    public string GetHubImageNamePrefix()
+    {
+      return _customConfiguration.GetHubImageNamePrefix();
+    }
+
+    private sealed class TestcontainersHostConfiguration : PropertiesFileConfiguration
+    {
+      protected override Uri GetDockerHost(string propertyName)
+      {
+        return base.GetDockerHost("testcontainers.host");
```
kiview:
```suggestion
        return base.GetDockerHost("tcc.host");
```

Haven't yet discussed, but maybe better to explicitly use a `tcc` namespace, to keep the `testcontainers` namespace potentially free.

src/Testcontainers/Builders/TestcontainersCloudEndpointAuthenticationProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+      return _customConfiguration.GetRyukContainerImage();
+    }
+
+    /// <inheritdoc />
+    public string GetHubImageNamePrefix()
+    {
+      return _customConfiguration.GetHubImageNamePrefix();
+    }
+
+    private sealed class TestcontainersHostConfiguration : PropertiesFileConfiguration
```
kiview, line 128:
```suggestion
    private sealed class TestcontainersCloudConfiguration : PropertiesFileConfiguration
```

kiview (dismissed):
Discussed internally, but added here for transparency:

Since this is effectively the integration with [Testcontainer Cloud](https://www.testcontainers.cloud), it's clearer if we rename the strategy accordingly.

src/Testcontainers/Builders/TestcontainersCloudEndpointAuthenticationProvider.cs
HofmeisterAn: Yes, this part was not finally done. The names are now aligned.

src/Testcontainers/Builders/TestcontainersCloudEndpointAuthenticationProvider.cs
kiview: Sorry for the confusion and to revert this. Going forward, this abstraction seems to be useful (and there can be a Testcontainers host that is `!=` running in TCC), so using `tc.host` might be the better way (see tc-node: https://github.com/testcontainers/testcontainers-node/pull/564/files#diff-a216e10441367a5e1ff574ed8c1b678a8a80886b31c6029231cdc291cb3e8f10R34)

PR #910, opened 2023-06-01
feat: Add UsingHttpMessageHandler(HttpMessageHandler) to HttpWaitStrategy

mkstephenson:
## What does this PR do?

This PR adds the ability to pass a custom HttpClientHandler to the HttpWaitStrategy.

## Why is it important?

One reason for passing a custom HttpClientHandler would be the ability to tell the health checks to ignore or deviate from local proxy settings (often useful if tests are being executed on a build agent which sits behind a proxy, where the settings for that proxy get in the way of communicating with the container).

### DIFF
--- a/src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
+++ b/src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
@@ -36,6 +36,8 @@ namespace DotNet.Testcontainers.Configurations
 
     private ushort? _portNumber;
 
+    private HttpClientHandler _httpClientHandler;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="HttpWaitStrategy" /> class.
     /// </summary>
@@ -64,7 +66,7 @@ namespace DotNet.Testcontainers.Configurations
         return false;
       }
 
-      using (var httpClient = new HttpClient())
+      using (var httpClient = _httpClientHandler == null ? new HttpClient() : new HttpClient(_httpClientHandler))
       {
         using (var httpRequestMessage = new HttpRequestMessage(_httpMethod, new UriBuilder(_schemeName, host, port, _pathValue).Uri))
         {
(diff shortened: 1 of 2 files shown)


src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
```diff
@@ -36,6 +36,8 @@ public sealed class HttpWaitStrategy : IWaitUntil
 
     private ushort? _portNumber;
 
+    private HttpClientHandler _httpClientHandler;
```
HofmeisterAn: Should we use the base class `HttpMessageHandler` instead?

src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
```diff
@@ -64,7 +66,7 @@ public async Task<bool> UntilAsync(IContainer container)
         return false;
       }
 
-      using (var httpClient = new HttpClient())
+      using (var httpClient = _httpClientHandler == null ? new HttpClient() : new HttpClient(_httpClientHandler))
```
HofmeisterAn:
```suggestion
      using (var httpClient = new HttpClient(_httpClientHandler ?? new HttpClientHandler()))
```

HofmeisterAn (commented):
Thank you for your pull request and contribution. The pull request looks good overall, I have two minor suggestions. WDYT?

src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
mkstephenson: That would make sense, it should allow for more flexibility.

mkstephenson: I have adjusted the code to use the base class instead.

HofmeisterAn: Thanks 👍

PR #913, opened 2023-06-06
feat: Add support for copying directories and files to a container

### LINKED ISSUE #811: [Enhancement]: Better method for sending multiple files to the conainer
### Problem

`.WithResourceMapping()` Allows us to send files to the container. When there are many files to be sent calling `StartAsync()` on the container can timeout. Probably as files are sent in parallel https://github.com/testcontainers/testcontainers-dotnet/blob/6688b92d217e710a773e87126fc98c78b27de253/src/Testcontainers/Clients/TestcontainersClient.cs#L294-L298

### Solution

To allow sending of archives that are then automatically unzipped in the container. Or this could be implemented as providing a list of files to the container builder so that the builder creates the archive and then unzips into a specified target directory.

https://github.com/testcontainers/testcontainers-dotnet/blob/6688b92d217e710a773e87126fc98c78b27de253/src/Testcontainers/Clients/TestcontainersClient.cs#L166-L206

### Benefit

When using sqitch for managing changes to database schemas the schema repo can grow very large resulting in many files needing to be moved.

### Alternatives

Using volume mounts which is not advised.

Yes

HofmeisterAn:
## What does this PR do?

The pull request extends the `IContainer` interface and adds the member `CopyAsync(...)` to support copying directories and files to a container.

## Why is it important?

The previous implementation only allowed copying single individual files. In cases where developers needed to copy entire directories, it became unnecessarily complicated. The new API is more convenient and easier for copying directories.

## Related issues

- Closes #811


src/Testcontainers/Configurations/Unix.cs
```diff
@@ -10,6 +10,60 @@ namespace DotNet.Testcontainers.Configurations
...
+    /// Represents the Unix file mode 644, which grants read and write permissions to the user and read permissions to the group and others.
+    /// </summary>
+    public const UnixFileMode FileMode644 =
+      UnixFileMode.UserRead |
+      UnixFileMode.UserWrite |
+      UnixFileMode.GroupRead |
+      UnixFileMode.OtherRead;
+
+    /// <summary>
+    /// Represents the Unix file mode 666, which grants read and write permissions to the user and group, and read and write permissions to others.
```
HofmeisterAn:
```suggestion
    /// Represents the Unix file mode 666, which grants read and write permissions to the user, group, and others.```

File: src/Testcontainers/Configurations/Unix.cs [hunk 2]
```diff
@@ -10,6 +10,60 @@ namespace DotNet.Testcontainers.Configurations
...
+    /// Represents the Unix file mode 644, which grants read and write permissions to the user and read permissions to the group and others.
+    /// </summary>
+    public const UnixFileMode FileMode644 =
+      UnixFileMode.UserRead |
+      UnixFileMode.UserWrite |
+      UnixFileMode.GroupRead |
+      UnixFileMode.OtherRead;
+
+    /// <summary>
+    /// Represents the Unix file mode 666, which grants read and write permissions to the user, group, and others.```
```
HofmeisterAn:
```suggestion
    /// Represents the Unix file mode 666, which grants read and write permissions to the user, group, and others.
```

PR #923, opened 2023-06-15
feat: Change access modifier of OracleBuilder.WithUsername(string) to public

0xced:
Contrary to what the remark says, the user is configurable.


HofmeisterAn (approved):
Thanks

PR #942, opened 2023-06-29
fix: Replace mysql with mariadb binary in MariaDB module

### LINKED ISSUE #941: [Bug]: MariaDB WaitUntil expects mysql command, replaced by mariadb in newer version
### Testcontainers version

3.3.0

Yes

### Host OS

Linux / Windows

### Host arch

amd64

### .NET version

7.0.101

### Docker version

```shell
Client:
 Cloud integration: v1.0.33
 Version:           24.0.2
 API version:       1.43
 Go version:        go1.20.4
 Git commit:        cb74dfc
 Built:             Thu May 25 21:53:15 2023
 OS/Arch:           windows/amd64
 Context:           default

Server: Docker Desktop 4.20.1 (110738)
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:52:17 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### Docker info

```shell
Shouldn't matter
```

### What happened?

Testcontainers hangs while it's waiting for the MariaDB container to start when the newest image is used. The reason seems to be that the command line utility has been renamed from `mysql` to `mariadb`. Older versions of MariaDB apparently had a symlink for `mysql`, but it looks like that has also been removed now:

```bash
> docker run --rm -it mariadb mysql
/usr/local/bin/docker-entrypoint.sh: line 617: exec: mysql: not found
```

it however still works for the version that is currently used by default in Testcontainers (`10.10`):

```bash
> docker run --rm -it mariadb:10.10 mysql
ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)
```

Since `mariadb` is already available in older versions of the image, I think it makes sense to just use that instead of `mysql`:

```bash
> docker run --rm -it mariadb:10.10 mariadb
ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)
> docker run --rm -it mariadb mariadb
ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)
```

### Relevant log output

```shell
[testcontainers.org 00:00:10.47] Docker image remote-docker-hub.jfrog.gdata.de/mariadb:latest created
[testcontainers.org 00:00:10.58] Docker container 43190b981419473c8ea9711c8c0f21d16f05568ee29bc4175ff4dc7de1e6b725 created
[testcontainers.org 00:00:10.68] Start Docker container 43190b981419473c8ea9711c8c0f21d16f05568ee29bc4175ff4dc7de1e6b725
[testcontainers.org 00:00:11.22] Execute "mysql --protocol=TCP --port=3306 --user=testuser --password=Password123 testdb --wait --silent --execute=SELECT 1;" at Docker container 43190b981419473c8ea9711c8c0f21d16f05568ee29bc4175ff4dc7de1e6b725
[testcontainers.org 00:00:12.35] Execute "mysql --protocol=TCP --port=3306 --user=testuser --password=Password123 testdb --wait --silent --execute=SELECT 1;" at Docker container 43190b981419473c8ea9711c8c0f21d16f05568ee29bc4175ff4dc7de1e6b725
[testcontainers.org 00:00:13.45] Execute "mysql --protocol=TCP --port=3306 --user=testuser --password=Password123 testdb --wait --silent --execute=SELECT 1;" at Docker container 43190b981419473c8ea9711c8c0f21d16f05568ee29bc4175ff4dc7de1e6b725
[...]
```

### Additional information

_No response_

FlorianHockmann:
The command has been renamed from `mysql` to `mariadb` and it looks like the symlink that has been in place for many versions has also been removed in newer MariaDB versions.

## What does this PR do?

`mysql` command changed to `mariadb` for the MariaDB module.

## Why is it important?

Wait policy fails otherwise for newer MariaDB images.

## Related issues

- Closes #941

## How to test this PR

Try to use the MariaDB image with a new tag, like `latest`.

### DIFF
--- a/src/Testcontainers.MariaDb/MariaDbBuilder.cs
+++ b/src/Testcontainers.MariaDb/MariaDbBuilder.cs
@@ -135,7 +135,7 @@ public sealed class MariaDbBuilder : ContainerBuilder<MariaDbBuilder, MariaDbCon
         /// <param name="configuration">The container configuration.</param>
         public WaitUntil(MariaDbConfiguration configuration)
         {
-            _command = new List<string> { "mysql", "--protocol=TCP", $"--port={MariaDbPort}", $"--user={configuration.Username}", $"--password={configuration.Password}", configuration.Database, "--wait", "--silent", "--execute=SELECT 1;" };
+            _command = new List<string> { "mariadb", "--protocol=TCP", $"--port={MariaDbPort}", $"--user={configuration.Username}", $"--password={configuration.Password}", configuration.Database, "--wait", "--silent", "--execute=SELECT 1;" };
         }
 
         /// <inheritdoc />
--- a/src/Testcontainers.MariaDb/MariaDbContainer.cs
+++ b/src/Testcontainers.MariaDb/MariaDbContainer.cs
@@ -45,7 +45,7 @@ public sealed class MariaDbContainer : DockerContainer
         await CopyAsync(Encoding.Default.GetBytes(scriptContent), scriptFilePath, Unix.FileMode644, ct)
             .ConfigureAwait(false);
 
-        return await ExecAsync(new[] { "mysql", "--protocol=TCP", $"--port={MariaDbBuilder.MariaDbPort}", $"--user={_configuration.Username}", $"--password={_configuration.Password}", _configuration.Database, $"--execute=source {scriptFilePath};" }, ct)
+        return await ExecAsync(new[] { "mariadb", "--protocol=TCP", $"--port={MariaDbBuilder.MariaDbPort}", $"--user={_configuration.Username}", $"--password={_configuration.Password}", _configuration.Database, $"--execute=source {scriptFilePath};" }, ct)
             .ConfigureAwait(false);
     }
 }


HofmeisterAn (commented):
Thank you for raising the issue and submitting the PR. I think we need to update the builder too:

https://github.com/testcontainers/testcontainers-dotnet/blob/c8a44a793039cb8bedb0b379226789a2a8e22b8b/src/Testcontainers.MariaDb/MariaDbBuilder.cs#L138

Do you know if removing the symlink is intended (are there any release notes)? Maybe it is an issue in the latest version of the image. I assume this will break many configurations, see: https://github.com/MariaDB/mariadb-docker/issues/519.

FlorianHockmann:
> I think we need to update the builder too:

Oh right, I just updated the commit to also change it there.

> Do you know if removing the symlink is intended (are there any release notes)?

I only found this part in [the release notes](https://mariadb.com/kb/en/mariadb-11-0-0-release-notes/):

> [MDEV-29582](https://jira.mariadb.org/browse/MDEV-29582) deprecate mysql* names

but looks like they didn't only deprecate them...

It however sounds me like they don't want to revert this change:

> I haven't see a major case for re-adding mysql commands, we where going to do it eventually, maybe a depreciation would have been nicer, but it seems to be done and the major use case of healthchecks has been improved.

([Source](https://github.com/MariaDB/mariadb-docker/issues/512#issuecomment-1608744838))

But at least for Testcontainers it makes sense in my opinion to switch to the `mariadb` command even if they would be adding the symlinks back as `mariadb` is the correct name and the symlink was only for backwards compatibility.

HofmeisterAn (approved):

PR #962, opened 2023-07-29
feat: Refactor `IImage` properties to align with DSL

### LINKED ISSUE #740: [Enhancement]: Refactor the `IImage` properties (align with DSL)
### Problem

The `IImage` interface does not follow the correct naming conventions.

### Solution

The interface should use the following properties: `Registry`, `Repository`, `Tag`, `FullName`.

### Benefit

Align with the DSL.

### Alternatives

\-

Yes

jacobjmarks:
## What does this PR do?

- Modifies `IImage` properties to align with DSL
  - Adds `string Registry`
  - Removes `string Name`
- Modifies `DockerImage` constructor to align with new properties
- Modifies `MatchImage.Match(string)` method to appropriately extract new properties
  - I used regex to do so. While the "requirements" shouldn't change, I'm happy to refactor to something a little more maintainable.

## Why is it important?

Aligns terminology used with that of the general Docker/container DSL.

## Related issues

- Resolves #740 

## Follow-ups

- Does this new support for registries make the [`hubImageNamePrefix`](https://github.com/testcontainers/testcontainers-dotnet/blob/ad14e246349c6af6e6c5e19039074a74c22660a5/src/Testcontainers/Images/DockerImage.cs#L48) parameter obsolete?

### DIFF
--- a/src/Testcontainers/Builders/ContainerBuilder`3.cs
+++ b/src/Testcontainers/Builders/ContainerBuilder`3.cs
@@ -80,7 +80,7 @@ namespace DotNet.Testcontainers.Builders
         return Clone(new ContainerConfiguration(image: image));
       }
 
-      return Clone(new ContainerConfiguration(image: new DockerImage(image.Repository, image.Name, image.Tag, TestcontainersSettings.HubImageNamePrefix)));
+      return Clone(new ContainerConfiguration(image: new DockerImage(image.Registry, image.Repository, image.Tag, TestcontainersSettings.HubImageNamePrefix)));
     }
 
     /// <inheritdoc />
--- a/src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
+++ b/src/Testcontainers/Builders/ImageFromDockerfileBuilder.cs
@@ -109,7 +109,7 @@ namespace DotNet.Testcontainers.Builders
     /// <inheritdoc />
     protected sealed override ImageFromDockerfileBuilder Init()
     {
-      return base.Init().WithImageBuildPolicy(PullPolicy.Always).WithDockerfile("Dockerfile").WithDockerfileDirectory(Directory.GetCurrentDirectory()).WithName(new DockerImage("localhost/testcontainers", Guid.NewGuid().ToString("D"), string.Empty));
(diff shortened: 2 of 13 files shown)


HofmeisterAn (commented):
Thanks for the PR once again 😃. I have not had a chance to thoroughly review it yet, but I would like to give some quick feedback on something I noticed.

I believe we should try to introduce as few breaking changes as possible to ensure a smooth transition. For the replaced members, I suggest using the obsolete flag (although I do not think a lot of developers are using them). Also, I want to mention that I forgot to include the `GetHostname()` in the issue, but have updated it.

> * Does this new support for registries make the [`hubImageNamePrefix`](https://github.com/testcontainers/testcontainers-dotnet/blob/ad14e246349c6af6e6c5e19039074a74c22660a5/src/Testcontainers/Images/DockerImage.cs#L48) parameter obsolete?

No, it does not. Developers have the option to configure it using the [custom configurations](https://dotnet.testcontainers.org/custom_configuration/) to replace the default registry (host) from which they want to pull the images.

jacobjmarks:
In this case, should the `hubImageNamePrefix` simply be considered the registry? I assume it will take precedence over any registry defined in the `repository` argument?

How do you want to handle the `DockerImage` constructor? I would look to do something like the following however it will clash with existing `DockerImage(string, string, string)` references.

``` csharp
// current constructor
[Obsolete]
public DockerImage(string repository, string name, string tag, string hubImageNamePrefix = null)
    : this(hubImageNamePrefix, $"{repository}/{name}", tag) { } // <-- forward to new constructor

// new constructor
public DockerImage(string registry, string repository, string tag) { }

HofmeisterAn:
Sorry for the late response, I was not able to reply earlier.

> In this case, should the `hubImageNamePrefix` simply be considered the registry? I assume it will take precedence over any registry defined in the `repository` argument? Like we are doing in `GetHostname()`.

Yes, I think that make sense. Do you think we can remove the registry in case the `repository` argument (obsolete constructor) contains one? This will probably support most backwards compatibility.

> How do you want to handle the `DockerImage` constructor? I would look to do something like the following however it will clash with existing `DockerImage(string, string, string)` references.

I have thought about that too. I do not think we can avoid a "breaking change", but I think the impact is very small.

- Not many developers will use this constructor.
- Internally, we use only the full name. I do not expect any issues 😬 (besides some existing tests).
- The old properties will still be available and marked as obsolete.

PR #963, opened 2023-08-02
feat: Add Azure Data Explorer Kusto emulator module

JonasBenz:
## What does this PR do?

Introduces a **Azure Data Explorer Kusto emulator** module.
The documentation about it can be found here: https://learn.microsoft.com/en-us/azure/data-explorer/kusto-emulator-install

## Why is it important?

Simplifies the setup of a Kusto testcontainer with all necessary parameters and a robust Wait strategy.
Having a specific KustoContainer also enables custom extension methods, strong typing, ...


src/Testcontainers.Kusto/KustoBuilder.cs
```diff
@@ -0,0 +1,100 @@
+namespace Testcontainers.Kusto;
+
+using System.Threading.Tasks;
```
HofmeisterAn: Move to global usings.

src/Testcontainers.Kusto/KustoBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /// <inheritdoc />
+    protected override KustoBuilder Init()
+    {
+        return base.Init()
+            .WithImage(KustoImage)
+            .WithPortBinding(KustoPort, true)
+            .WithEnvironment("ACCEPT_EULA", "Y")
+            .WithWaitStrategy(Wait.ForUnixContainer().AddCustomWaitStrategy(new WaitUntil()));
```
HofmeisterAn: Can't you use the  HTTP wait strategy? E.g. like

    _ = Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request => request.ForPort(KustoPort).ForPath("/v1/rest/mgmt"));
    
(Probably won't work if the HTTP request requires a payload)

src/Testcontainers.Kusto/KustoContainer.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+    }
+
+    /// <summary>
+    /// Gets the Kusto connection string.
+    /// </summary>
+    /// <returns>The Kusto connection string.</returns>
+    public string GetConnectionString()
+    {
+        return $"http://{Hostname}:{GetMappedPublicPort(KustoBuilder.KustoPort)}";
```
HofmeisterAn:
```suggestion
        return new UriBuilder(Uri.UriSchemeHttp, Hostname, GetMappedPublicPort(KustoBuilder.KustoPort)).ToString();
```

HofmeisterAn (commented):
Thanks for your PR and contribution. I have some minor suggestions. Maybe we can use the HTTP wait strategy implementation instead of the custom wait strategy (dunno if we need the JSON payload). Happy to merge the PR afterward.

src/Testcontainers.Kusto/KustoBuilder.cs
JonasBenz: fixed

src/Testcontainers.Kusto/KustoBuilder.cs
JonasBenz: Cool 😎, thanks for pointing that out.

You are right about the payload, so I extended the `HttpWaitStrategy`: https://github.com/testcontainers/testcontainers-dotnet/pull/963/commits/23f3c7526af05582e4f7a2f51ce94de77d93c9f3
With this extension I can use the same request (which I took from the offical documentation): https://github.com/testcontainers/testcontainers-dotnet/pull/963/commits/ad9635f67a4ecd4da2751593b148f86510efdec0

What do you think about this?

JonasBenz:
> Thanks for your PR and contribution. I have some minor suggestions. Maybe we can use the HTTP wait strategy implementation instead of the custom wait strategy (dunno if we need the JSON payload). Happy to merge the PR afterward.

Thanks for the fast feedback 👍 
After you merged, do we have to wait for the next release until the *.Kusto NuGet will be available? Or when can we expect it be published?

src/Testcontainers.Kusto/KustoBuilder.cs
HofmeisterAn: Exactly what I thought. Unfortunately, it looks like we cannot do it like that. The HTTP client disposes the HTTP content if the first request fails. Afterward, we cannot use the content again 😕.

HofmeisterAn:
> After you merged, do we have to wait for the next release until the *.Kusto NuGet will be available? Or when can we expect it be published?

I have planned a new release, end of this week. I can wait until this PR is ready and do it afterward.

src/Testcontainers.Kusto/KustoBuilder.cs
HofmeisterAn: I've added a suggestion that uses the container's console output to  detect readiness. Please share your thoughts on this. If you find it unreliable or prefer the curl approach, we can also go with that.

JonasBenz: > Exactly what I thought. Unfortunately, it looks like we cannot do it like that. The HTTP client disposes the HTTP content if the first request fails. Afterward, we cannot use the content again 😕.

Oh you are right (for .NET Core <= 3.0) according to this: https://makolyte.com/csharp-disposing-the-request-httpcontent-when-using-httpclient/
Not the first time I fall into this trap 😅

I changed the `HttpWaitStrategy` to take a factory for the `HttpContent`, so that we can handle the disposal. What do you think about this?: https://github.com/testcontainers/testcontainers-dotnet/pull/963/commits/09b0b0692b61470f62b5f93bf415ea9642f484a4 (including this fix: https://github.com/testcontainers/testcontainers-dotnet/pull/963/commits/55018dc890f70ca70c0440002c0561c5bb122175)

JonasBenz: > I've added a suggestion that uses the container's console output to detect readiness. Please share your thoughts on this. If you find it unreliable or prefer the curl approach, we can also go with that.

Thanks for the proposal 👍. I changed it back to the `HttpWaitStrategy` which now should be able to handle the diposal of the content correctly.

I thought it would be more stable to use the HTTP call which they also use in the documentation here: https://learn.microsoft.com/en-us/azure/data-explorer/kusto-emulator-install#install-the-kusto-emulator

Probably it is unlikely that they will ever change the console output and your proposal is much simpler, so I have no strong opinion on that. If you like checking the console output better, I can also revert the changes in the `HttpWaitStrategy` and go for your proposal.

HofmeisterAn: I think that's a good idea. I will take a closer look later the day.

src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
```diff
@@ -38,6 +38,8 @@ public sealed class HttpWaitStrategy : IWaitUntil
 
     private HttpMessageHandler _httpMessageHandler;
 
+    private Func<HttpContent> _contentFactory = () => null;
```
HofmeisterAn: Initializing the default value is done via the default constructor and the corresponding with member: https://github.com/testcontainers/testcontainers-dotnet/blob/55018dc890f70ca70c0440002c0561c5bb122175/src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs#L48

It would be nice if we can use the same approach for the HTTP content too.

src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
```diff
@@ -75,51 +77,54 @@ public async Task<bool> UntilAsync(IContainer container)
...
-          }
-          catch (HttpRequestException)
-          {
-            return false;
-          }
-
-          Predicate<HttpStatusCode> predicate;
-
-          if (!_httpStatusCodes.Any() && _httpStatusCodePredicate == null)
+          using (httpRequestMessage.Content = _contentFactory())
```
HofmeisterAn: Do we need the `using` block here? I think disposing the HTTP client will dispose the message anyway, that was our initial issue, right?

JonasBenz: yes you are right... but according to the follwing blog, they changed the behaviour from .NET core 3.0 onwards, so that it will not be disposed implicitly anymore with newer .NET (core) versions: https://makolyte.com/csharp-disposing-the-request-httpcontent-when-using-httpclient/

Because this is a .NET standard project and we don't really know which version the consumer is running, I thought it is better to dispose it explicitly and live with it that it gets disposed twice in older versions.

src/Testcontainers/Configurations/WaitStrategies/HttpWaitStrategy.cs
JonasBenz: sure 👍 , I aligned it: https://github.com/testcontainers/testcontainers-dotnet/pull/963/commits/244a873750a660102a0393423e1d924f07f43940

HofmeisterAn (dismissed):
Thanks again! PR looks good.

HofmeisterAn (approved):

PR #976, opened 2023-08-15
change private accesss modifier to public modifier in MsSqlBuilder File

ismaeel-ch:
## What does this PR do?

In `MsSqlBuilder` class, there are `WithDatabase` and `WithUsername` methods that have `private access modifiers`. These should be `public access modifiers`. 

## Why is it important?

There is no way to change `Username `and `Database `rather than reflection. By changing `private `to `public `access modifier, we can easily change Username and Database name.

## Related issues

N/A

## How to test this PR

N/A

## Follow-ups

N/A


HofmeisterAn (commented):
Please see my reply from the following discussion: https://github.com/testcontainers/testcontainers-dotnet/discussions/828#discussioncomment-5253889. There are a couple of issues and discussions about similar topics. The MSSQL image does not allow you to change the database or username (via env variables). Even with a public modifier you won't be able to set the database or username. Relates to: https://github.com/testcontainers/testcontainers-dotnet/pull/855#issuecomment-1491389100.

PR #979, opened 2023-08-19
fix: Do not pre pull Dockerfile build stages that do not correspond to base images

### LINKED ISSUE #972: [Bug]: ImageFromDockerfileBuilder cannot build `Dockerfile` with intermediate layer (regression works with `3.3.0`)
### Problem

I cannot give the docker directory to the `ImageFromDockerfileBuilder`, only the dockerfile is possible:
```C#
public ImageFromDockerfileBuilder WithDockerfile(string dockerfile)
    {
      return Merge(DockerResourceConfiguration, new ImageFromDockerfileConfiguration(dockerfile: dockerfile));
    }
```
I need to set the root to a certain directory so that the Copy instructions in my Dockerfile work.

### Solution

I am not sure if this is enough, but consider this:

Add another method accepting the dockerfile and the directory.

```C#
public ImageFromDockerfileBuilder WithDockerfile(string dockerfile, string dockerfileDirectory)
    {
      return Merge(DockerResourceConfiguration, new ImageFromDockerfileConfiguration(dockerfile: dockerfile,
                                                                                     dockerfileDirectory: dockerfileDirectory));
    }
```

### Benefit

This change could enable my use case for using the `ImageFromDockerfileBuilder`.

### Alternatives

I am not sure if there are other alternatives, I don't understand Docker well enough yet.

Yes

HofmeisterAn:
## What does this PR do?

This pull request excludes Dockerfile build stages that do not correspond to base images.

## Why is it important?

This pull request addresses an issue related to building Docker images that have dependencies on base images from private registries. Currently, Testcontainers encounters an issue when attempting to pull actual images. Testcontainers does not distinguish between base images and intermediate build stages. Intermediate build stages cannot be pulled; Testcontainers must ignore these stages. E.g. the second line of the following Dockerfile:

```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build
FROM build
```

## Related issues

- Closes #972
- Relates #980

## Follow-ups

Testcontainers for .NET cannot resolve variables (`ARG`s) during the Docker image build process. Currently, Testcontainers ignores all base images that utilize variables. To pull these base images in advance too, Testcontainers needs to have the capability to resolve the default values or build arguments of variables.

### DIFF
--- a/src/Testcontainers/Images/DockerfileArchive.cs
+++ b/src/Testcontainers/Images/DockerfileArchive.cs
@@ -97,7 +97,7 @@ namespace DotNet.Testcontainers.Images
 
       var stages = lines
         .Select(line => line.Value)
-        .Select(line => line.Split(new [] { " AS " }, StringSplitOptions.RemoveEmptyEntries))
+        .Select(line => line.Split(new [] { " AS ", " As ", " aS ", " as " }, StringSplitOptions.RemoveEmptyEntries))
         .Where(substrings => substrings.Length > 1)
         .Select(substrings => substrings[substrings.Length - 1])
         .Distinct()
--- a/src/Testcontainers/Images/DockerfileArchive.cs
+++ b/src/Testcontainers/Images/DockerfileArchive.cs
@@ -72,7 +72,7 @@ namespace DotNet.Testcontainers.Images
     /// <remarks>
     /// This method reads the Dockerfile and collects a list of base images. It
     /// excludes stages that do not correspond to base images. For example, it will not include
-    /// the second line in the Dockerfile below:
+    /// the second line from the following Dockerfile configuration:
     /// <code>
     ///   FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build
     ///   FROM build
--- a/src/Testcontainers/Images/DockerfileArchive.cs
+++ b/src/Testcontainers/Images/DockerfileArchive.cs
@@ -17,6 +17,8 @@ namespace DotNet.Testcontainers.Images
   /// </summary>
   internal sealed class DockerfileArchive : ITarArchive
   {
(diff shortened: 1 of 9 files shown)


src/Testcontainers/Images/DockerfileArchive.cs
```diff
@@ -64,6 +66,54 @@ public DockerfileArchive(DirectoryInfo dockerfileDirectory, FileInfo dockerfile,
...
+        .Where(match => match.Success)
+        // Until now, we are unable to resolve variables within Dockerfiles. Ignore base
+        // images that utilize variables. Expect them to exist on the host.
+        .Where(match => !match.Groups[imageGroup].Value.Contains('$'))
+        .Where(match => !match.Groups[imageGroup].Value.Any(char.IsUpper))
+        .ToArray();
+
+      var stages = lines
+        .Select(line => line.Value)
+        .Select(line => line.Split(new [] { " AS " }, StringSplitOptions.RemoveEmptyEntries))
```
jasdefer: Are lower cases of  `" AS "`somehow considered too?

HofmeisterAn: Yes, you are correct. While I was working on the fix, I observed that Rider was complaining about lowercase usage. This led me to think that this might be an invalid configuration. I double-checked it, and it is working with the Docker CLI (`as`, `aS`, etc.).

HofmeisterAn:
```suggestion
        .Select(line => line.Split(new [] { " AS ", " As ", " aS ", " as " }, StringSplitOptions.RemoveEmptyEntries))
```

goldsam:
This is still not working with 3.5.0. Getting the same results as OP.

PR #547, opened 2022-08-01
Feat: No Auth MongoDb

the-avid-engineer:
The MongoDb docker container supports a no-auth mode such that you don't need to specify a username or password. While it's simple enough to work around the current behavior in consumer code by using a string replacement, this change will just output a valid connection string from the beginning

Motivation comes from #538 - when running with no-auth in replica set mode, you don't need to specify a security key. Replica set mode is useful because it gives you access to Transactions.

NOTE - I was not able to run the tests locally. I got as far as `dotnet cake --target=Tests`, but I'm getting this exception:

```
Docker.DotNet.DockerImageNotFoundException : Docker API responded with status code=NotFound, response={"message":"No such image: testcontainers/ryuk:0.3.4"}
```

PR #983, opened 2023-08-28
feat: Allow MongoDb module configuration without credentials

### LINKED ISSUE #982: [Enhancement]: Explicit no-authentication mode for MongoDb
### Problem

The 3.0 release broke MongoDb no-authentication mode, which makes replica set mode easier to use (see https://github.com/testcontainers/testcontainers-dotnet/issues/538)

If you explicitly use `new MongoDbBuilder().WithUsername(null).WithPassword(null)`, the code will now default to using `mongo` for the username and password, and if you use empty string instead of null, a runtime error is throwing saying these fields are required, which means it is now impossible to use no-authentication mode out of the box.

### Solution

Provide a property on MongoDbConfiguration called `useDefaultCredentials` that defaults to `true` to preserve current behavior, but can be set to `false` to get the desired behavior.

### Benefit

This allows for running mongodb in replica set mode without having to set up the extra bits that are required when authentication is enabled, and having an explicit feature to not use the default credentials will hopefully prevent a regression

### Alternatives

I can only think of workarounds, not alternatives

Yes

the-avid-engineer:
## What does this PR do?

1. Do not require a username/password for mongodb
2. Allow for disabling default credentials for mongodb

## Why is it important?

This addresses a regression on https://github.com/testcontainers/testcontainers-dotnet/pull/547 as well as adding the ability to not use default credentials; when using credentials in replica set mode, there are additional set up pieces that are tedious; it's 1000x easier to run in replica set mode without credentials. Replica set mode allows users to use transactions

## Related issues

- Closes https://github.com/testcontainers/testcontainers-dotnet/issues/982

---

I would love to add a test to prevent another regression, but on my last PR I wasn't able to run the tests locally, and I suspect I still cannot run them

### DIFF
--- a/src/Testcontainers.MongoDb/MongoDbBuilder.cs
+++ b/src/Testcontainers.MongoDb/MongoDbBuilder.cs
@@ -17,6 +17,7 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
     /// <summary>
     /// Initializes a new instance of the <see cref="MongoDbBuilder" /> class.
     /// </summary>
+    /// <param name="useDefaultCredentials">When set to <c>false</c>, mongodb will run without authentication unless you explicitly specify credentials</param>
     public MongoDbBuilder(bool useDefaultCredentials = true)
         : this(new MongoDbConfiguration())
     {
--- a/src/Testcontainers.MongoDb/MongoDbBuilder.cs
+++ b/src/Testcontainers.MongoDb/MongoDbBuilder.cs
@@ -70,19 +70,21 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
     /// <inheritdoc />
     protected override MongoDbBuilder Init()
     {
-        var builder = base.Init()
-            .WithImage(MongoDbImage)
-            .WithPortBinding(MongoDbPort, true)
-            .WithWaitStrategy(Wait.ForUnixContainer().AddCustomWaitStrategy(new WaitUntil()));
+        string username = null;
+        string password = null;
 
         if (_useDefaultCredentials)
         {
-            builder = builder
-                .WithUsername(DefaultUsername)
-                .WithPassword(DefaultPassword);
+            username = DefaultUsername;
+            password = DefaultPassword;
         }
 
-        return builder;
+        return base.Init()
+            .WithImage(MongoDbImage)
+            .WithPortBinding(MongoDbPort, true)
+            .WithUsername(username)
+            .WithPassword(password)
+            .WithWaitStrategy(Wait.ForUnixContainer().AddCustomWaitStrategy(new WaitUntil()));
     }
 
     /// <inheritdoc />
--- a/src/Testcontainers.MongoDb/MongoDbBuilder.cs
+++ b/src/Testcontainers.MongoDb/MongoDbBuilder.cs
@@ -80,11 +80,9 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
         base.Validate();
 
         _ = Guard.Argument(DockerResourceConfiguration.Username, nameof(DockerResourceConfiguration.Username))
-            .NotNull()
             .NotEmpty();
 
         _ = Guard.Argument(DockerResourceConfiguration.Password, nameof(DockerResourceConfiguration.Password))
-            .NotNull()
             .NotEmpty();
     }
 
--- a/src/Testcontainers.MongoDb/MongoDbBuilder.cs
+++ b/src/Testcontainers.MongoDb/MongoDbBuilder.cs
@@ -17,7 +17,7 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
     /// <summary>
     /// Initializes a new instance of the <see cref="MongoDbBuilder" /> class.
     /// </summary>
-    /// <param name="useDefaultCredentials">When set to <c>false</c>, mongodb will run without authentication unless you explicitly specify credentials</param>
+    /// <param name="useDefaultCredentials">When set to <c>false</c>, MongoDb will run without authentication unless you explicitly specify a username and password.</param>
     public MongoDbBuilder(bool useDefaultCredentials = true)
         : this(new MongoDbConfiguration())
     {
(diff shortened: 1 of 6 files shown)


src/Testcontainers.MongoDb/MongoDbBuilder.cs
```diff
@@ -12,12 +12,16 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
 
     public const string DefaultPassword = "mongo";
 
+    private readonly bool _useDefaultCredentials;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="MongoDbBuilder" /> class.
     /// </summary>
-    public MongoDbBuilder()
+    /// <param name="useDefaultCredentials">When set to <c>false</c>, mongodb will run without authentication unless you explicitly specify credentials</param>
```
HofmeisterAn:
```suggestion
    /// <param name="useDefaultCredentials">When set to <c>false</c>, MongoDb will run without authentication unless you explicitly specify a username and password.</param>
```

src/Testcontainers.MongoDb/MongoDbBuilder.cs
```diff
@@ -66,12 +70,19 @@ public override MongoDbContainer Build()
     /// <inheritdoc />
     protected override MongoDbBuilder Init()
     {
-        return base.Init()
+        var builder = base.Init()
```
HofmeisterAn: What are your thoughts on something like that:

```csharp
string username;
string password;

if (_useDefaultCredentials)
{
    username = DefaultUsername;
    password = DefaultPassword;
}
else
{
    username = null;
    password = null;
}

...
```

This approach keeps the builder configuration in one place. I don't have a strong opinion on this; I just considered it might be beneficial to maintain the configuration in a cohesive manner.

HofmeisterAn (commented):
Thanks for your PR 👍.

> This addresses a regression on https://github.com/testcontainers/testcontainers-dotnet/pull/547 as well as adding the ability to not use default credentials;

Oh, looks like I missed that requirement during the migration to dedicated modules, thanks.

src/Testcontainers.MongoDb/MongoDbBuilder.cs
the-avid-engineer: This seems okay to me!

The only thing is I'm not sure how mongodb will behave if it has the environment variables available with no value; there might need to be a not-null check inside the WithUsername and WithPassword methods

I will update the PR later today

src/Testcontainers.MongoDb/MongoDbBuilder.cs
the-avid-engineer: Should the other doc comments be updated to `MongoDb` as well?

HofmeisterAn:
We must ensure that the tests also pass (`_ = new MongoDbBuilder(false).Build()`). The container instance is capable of executing a MongoDB [shell command](https://github.com/testcontainers/testcontainers-dotnet/blob/27611d1967b9795cb457c0438d75ee3327163a39/src/Testcontainers.MongoDb/MongoDbShellCommand.cs#L27-L28) that relies on the credentials, too. Perhaps we can replace a null string with an empty one.

https://github.com/testcontainers/testcontainers-dotnet/blob/27611d1967b9795cb457c0438d75ee3327163a39/tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs#L36-L49

src/Testcontainers.MongoDb/MongoDbBuilder.cs
HofmeisterAn: The [Dockerfile](https://github.com/docker-library/mongo/blob/971f4ce1976457fddd5cf722268da9f7759d7fc8/6.0/docker-entrypoint.sh#L258), at the very least, verifies the presence of both environment variables (version 4.4, 5 and 6).

the-avid-engineer: I confirmed that `.WithEnvironment("MONGO_INITDB_ROOT_USERNAME", null)` and `.WithEnvironment("MONGO_INITDB_ROOT_PASSWORD", null)` cause no problems :)

the-avid-engineer:
>We must ensure that the tests also pass (_ = new MongoDbBuilder(false).Build()). The container instance is capable of executing a MongoDB [shell command](https://github.com/testcontainers/testcontainers-dotnet/blob/27611d1967b9795cb457c0438d75ee3327163a39/src/Testcontainers.MongoDb/MongoDbShellCommand.cs#L27-L28) that relies on the credentials, too. Perhaps we can replace a null string with an empty one.

I don't have a Linux box (I'm assuming it only works on Linux because the test has a Linux trait), so I went with the best bet that not passing those options when username/password are null will certainly work as expected, and I'm not opposed to passing empty strings if that works

HofmeisterAn:
I extended the tests and executed them against `MongoDbBuilder(false)`, resulting in failures. Afterward, I made some adjustments to the implementation. It appears that we can keep non-nullable while achieving a configuration with no authentication by utilizing empty strings. Kindly share your thoughts on this matter. I am not a MongoDB expert, but curious why MongoDB logs different messages according to the configuration (user vs. w/o user). Probably something you ran into.

We can add something like `WithNoAuth()` to offer a more explicit API.

the-avid-engineer:
I do know that `mongodb://:@localhost:12345` is not a valid connection string, so it won't work there

I will check the other changes against replica set mode

HofmeisterAn:
> I do know that `mongodb://:@localhost:12345` is not a valid connection string, so it won't work there

That's not what `GetConnectionString()` returns. If you run the `MongoDbNoAuthConfiguration` test you will see it returns

    mongodb://127.0.0.1:56789/
    
If the username and password is empty it will not be included in the connection string.

HofmeisterAn:
> I will check the other changes against replica set mode

@the-avid-engineer have you had the chance to test the changes in replica set mode? I would like to prepare a new release, and ideal include the changes from this pull request.

the-avid-engineer:
Sorry, I was preoccupied and not able to test yet. I will do so later tonight

That is good that the connection string is correct, I only assumed it would not handle empty string the same as null

the-avid-engineer:
@HofmeisterAn I can confirm the empty strings work fine with replica set mode :)

HofmeisterAn (approved):
Thanks for taking the time 🙏.

ConMD:
Hi everyone,

hope this is the right place to ask this question:
When will this PR be merged into main?

Would be highly appreciated, if it happens soon :)

HofmeisterAn:
This has already been merged into the main branch and is released.

https://github.com/testcontainers/testcontainers-dotnet/blob/c0932f2170cf53d15a69a52626b736606a4d28f8/tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs#L61-L67

PR #984, opened 2023-08-29
Change WithDatabase method to public

MarkusKeusch:
## What does this PR do?

Expose the `WithDatabase` method of `MsSqlBuilder`

## Why is it important?

By default the initial catalog of the connectionstring is set to master. In some cases we want to just delete and recreate the whole database which is not possible for the master database.

## Related issues

-

### DIFF
--- a/src/Testcontainers.MsSql/MsSqlBuilder.cs
+++ b/src/Testcontainers.MsSql/MsSqlBuilder.cs
@@ -104,7 +104,7 @@ public sealed class MsSqlBuilder : ContainerBuilder<MsSqlBuilder, MsSqlContainer
     /// </remarks>
     /// <param name="database">The MsSql database.</param>
     /// <returns>A configured instance of <see cref="MsSqlBuilder" />.</returns>
-    private MsSqlBuilder WithDatabase(string database)
+    public MsSqlBuilder WithDatabase(string database)
     {
         return Merge(DockerResourceConfiguration, new MsSqlConfiguration(database: database))
             .WithEnvironment("SQLCMDDBNAME", database);


HofmeisterAn (commented):
Unfortunately, the Microsoft SQL Server Docker image does not support the configuration of the database or username through environment variables during the container start. Changing the modifier to `public` will not create or configure the database. This limitation is from the image itself, rather than a constraint from Testcontainers.

https://github.com/testcontainers/testcontainers-dotnet/blob/b121ddebe1597f94f73e307a079bedd3d6762053/src/Testcontainers.MsSql/MsSqlBuilder.cs#L102-L104

Nevertheless, the image does allow the execution of customized SQL [scripts](https://github.com/microsoft/mssql-docker/tree/master/linux/preview/examples/mssql-customize) (with a couple of adjustments). Along with `WithResourceMapping(bytes[], string)`, we can utilize this to configure the database or username. Furthermore, this enhances the capabilities of the module (image) to similar functionalities like PostgreSQL image offers.

Another alternative is to use the EF or override the database in the connection string.

This issue relates:

- https://github.com/testcontainers/testcontainers-dotnet/discussions/828#discussioncomment-5577198
- https://github.com/testcontainers/testcontainers-dotnet/pull/855
- https://github.com/testcontainers/testcontainers-dotnet/pull/976

JonasBenz:
> Unfortunately, the Microsoft SQL Server Docker image does not support the configuration of the database or username through environment variables during the container start. Changing the modifier to `public` will not create or configure the database. This limitation is from the image itself, rather than a constraint from Testcontainers.
> 
> https://github.com/testcontainers/testcontainers-dotnet/blob/b121ddebe1597f94f73e307a079bedd3d6762053/src/Testcontainers.MsSql/MsSqlBuilder.cs#L102-L104
> 
> Nevertheless, the image does allow the execution of customized SQL [scripts](https://github.com/microsoft/mssql-docker/tree/master/linux/preview/examples/mssql-customize) (with a couple of adjustments). Along with `WithResourceMapping(bytes[], string)`, we can utilize this to configure the database or username. Furthermore, this enhances the capabilities of the module (image) to similar functionalities like PostgreSQL image offers.
> 
> Another alternative is to use the EF or override the database in the connection string.
> 
> This issue relates:
> 
> * [MsSqlBuilder - why is WithDatabase private? #828 (reply in thread)](https://github.com/testcontainers/testcontainers-dotnet/discussions/828#discussioncomment-5577198)
> * [Testcontainers.MsSql: WithDatabase to public #855](https://github.com/testcontainers/testcontainers-dotnet/pull/855)
> * [change private accesss modifier to public modifier in MsSqlBuilder File #976](https://github.com/testcontainers/testcontainers-dotnet/pull/976)

What @MarkusKeusch and me want is just to have the "initialCatalog" of the connection string set to a specific DB. Of course we can set the initalCatalog in our project manually or make our own extension method. We will care ourself about creating the DB.

But it would be nice to have the functionality directly from testcontainers. I understand that `WithDatabase` could lead to false assumptions, that the DB will be created for you.

@HofmeisterAn What do you think about adding an overload for GetConnectionString(), where  you can specify the DB, like this: `GetConnectionString(string database)` or maybe better: `GetConnectionString(string initialCatalog)` 

Or what if we make a builder method `WithInitialCatalog(string database)`? I would like this approach the best, because it would be obvious, that this will not create a DB and you can specify it once, when you by default not want to work with the "master" DB.

HofmeisterAn:
You are referring to the workaround mentioned above, in which Entity Framework creates the database if it does not already exist, right? As previously mentioned, I believe this can be quite misleading. If developers do not utilize EF to generate the non-existent database, they run into issues. I will close this issue. Let's continue the discussion in https://github.com/testcontainers/testcontainers-dotnet/issues/986 and see if we can find a suitable solution.

PR #985, opened 2023-08-29
chore: Update BouncyCastle.Cryptography to 2.2.1 (previous Portable.BouncyCastle)

jcmrva:
Dependency update.

## What does this PR do?

Moving back to the main distro of BouncyCastle.Crypto now that it supports netstandard etc.
https://www.nuget.org/packages/BouncyCastle.Cryptography

## Why is it important?

Portable.BouncyCastle hasn't been updated in almost 2 years, and I assume Claire didn't intend to own it forever.

## Related issues

n/a (that I know of)

## How to test this PR

This builds but the tests have some failures I'm not able to diagnose; also the test process stalled on the last step (see gist) for more than 15 minutes before I killed it.
https://gist.github.com/jcmrva/c83912eb5834b4c0d67f07671b0771d0

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -6,7 +6,7 @@
     <PackageReference Update="Docker.DotNet.X509" Version="3.125.15" />
     <PackageReference Update="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <PackageReference Update="Microsoft.Extensions.Logging.Abstractions" Version="6.0.4" />
-    <PackageReference Update="Portable.BouncyCastle" Version="1.9.0" />
+    <PackageReference Update="BouncyCastle.Cryptography" Version="2.2.1" />
     <PackageReference Update="SharpZipLib" Version="1.4.2" />
     <PackageReference Update="SSH.NET" Version="2020.0.2" />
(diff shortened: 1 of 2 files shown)


HofmeisterAn (commented):
Thank you for bringing this to my attention. I was unaware of the recent changes in the .NET Bouncy Castle project. It's great that we can revert to the main distribution, although Claire's version was running smoothly.

> This builds but the tests have some failures I'm not able to diagnose; also the test process stalled on the last step (see gist) for more than 15 minutes before I killed it.

I don't believe these issues are related to your changes. All the tests run smoothly on my Windows machine with Docker Desktop. However, running the tests simultaneously puts a lot of load on the system. Depending on the available resources, there might be cases where the containers fail to start correctly. If you're able to reproduce the issues, it would be great if you could provide additional information. This would help in determining whether these issues require a fix.

jcmrva:
Yeah, I'm fairly sure it has something to do with my setup, although I'm also on Windows with Docker Desktop. I'll see if I can discover more about what's going on.

HofmeisterAn (approved):

PR #1003, opened 2023-09-17
feat: Add NATS module

niklasfp:
## What does this PR do?

This pr adds support for the [official Nats Server image](https://hub.docker.com/_/nats/)
It should roughly align with the Go version  providing support for `.WithUsername` and `.WithPassword`

The container enables the http monitoring endpoints of Nats, and exposes the default Nats server ports:

* Client port 4222 
* Http monitoring endpoints on port 8222 
* Routes on port 6222

It also adds methods for getting a Nats connection string and the url for the monitoring endpoint.

## Why is it important?

I needed it for my own work, and thought I might as well share it.
And since Nats already has both Go and Node.js implementations, I guess others would benefit as well.

## Related issues

N/A

## How to test this PR

Tests has been added in this pr.
Run the tests in `./tests/Testcontainers.Nats.Tests`


src/Testcontainers.Nats/NatsBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// Sets the Nats config.
+    /// </summary>
+    /// <param name="config">The Nats config.</param>
+    /// <returns>A configured instance of <see cref="NatsBuilder" />.</returns>
+    public NatsBuilder WithNatsConfig(NatsConfiguration config)
+    {
+        // Extends the ContainerBuilder capabilities and holds a custom configuration in NatsConfiguration.
+        // In case of a module requires other properties to represent itself, extend ContainerConfiguration.
+        return Merge(DockerResourceConfiguration, new NatsConfiguration(config));
+    }
```
HofmeisterAn: What is the use case for this member? None of the other modules has anything like that. Typically, the configuration instance is created by the builder, a pattern I would like to keep consistent across the Testcontainers for .NET modules.

src/Testcontainers.Nats/NatsBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+        return base.Init()
+            .WithImage(NatsImage)
+            .WithPortBinding(ClientPort, true)
+            .WithPortBinding(MonitoringPort, true)
+            .WithPortBinding(RoutingPort, true)
+            .WithCommand("-m", MonitoringPort.ToString()) // Enable monitoring endpoint.
+            .WithCommand("-js") // Enable JetStream functionality.
+            .WithCommand("-DV") // Enable both debug and protocol trace messages 
+            .WithWaitStrategy(Wait.ForUnixContainer()
+                .UntilMessageIsLogged("Listening for client connections on 0.0.0.0:4222"));
```
HofmeisterAn: Should we set a default username and password and ensure that the configuration is not null (in `Validate()`)? IIRC, if the URI contains a username and password, they need to be escaped using `Uri.EscapeDataString(_configuration.Username)`, e.g. like:

https://github.com/testcontainers/testcontainers-dotnet/blob/53172407ffdf79dd8c83fe9dd370f1fb7baaa312/src/Testcontainers.MongoDb/MongoDbContainer.cs#L27-L30

src/Testcontainers.Nats/NatsBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+    protected override NatsBuilder Init()
+    {
+        return base.Init()
+            .WithImage(NatsImage)
+            .WithPortBinding(ClientPort, true)
+            .WithPortBinding(MonitoringPort, true)
+            .WithPortBinding(RoutingPort, true)
+            .WithCommand("-m", MonitoringPort.ToString()) // Enable monitoring endpoint.
+            .WithCommand("-js") // Enable JetStream functionality.
+            .WithCommand("-DV") // Enable both debug and protocol trace messages
```
HofmeisterAn: Could we please use the long version of the parameter? It's easier for developers who are not familiar with the CLI options.

src/Testcontainers.Nats/NatsBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// <summary>
+    /// Sets the Nats Server password.
+    /// </summary>
+    /// <param name="password">The Nats Server password.</param>
+    /// <returns>A configured instance of <see cref="NatsBuilder" />.</returns>
+    public NatsBuilder WithPassword(string password)
+    {
+        return Merge(DockerResourceConfiguration, new NatsConfiguration(password: password))
+            .WithCommand("-pass", password);
```
HofmeisterAn:
```suggestion
            .WithCommand("--pass", password);
```

According to the documentation, it is using double dashes.

src/Testcontainers.Nats/NatsBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// <summary>
+    /// Sets the Nats Server username.
+    /// </summary>
+    /// <param name="username">The Nats Server username.</param>
+    /// <returns>A configured instance of <see cref="NatsBuilder" />.</returns>
+    public NatsBuilder WithUsername(string username)
+    {
+        return Merge(DockerResourceConfiguration, new NatsConfiguration(username: username))
+            .WithCommand("-user", username);
```
HofmeisterAn:
```suggestion
            .WithCommand("--user", username);
```

According to the documentation, it is using double dashes.

src/Testcontainers.Nats/NatsContainer.cs
```diff
@@ -0,0 +1,44 @@
...
+        }.ToString();
+    }
+
+    /// <summary>
+    /// Gets the nats monitor url
+    /// </summary>
+    /// <returns>A url in the form: http://hostname:mappedPort/>.</returns>
+    public string GetMonitorUrl()
+    {
+        return new UriBuilder("http", Hostname, GetMappedPublicPort(NatsBuilder.MonitoringPort)).ToString();
```
HofmeisterAn: Is authentication also required when accessing the monitor endpoint?

HofmeisterAn (commented):
Thank you for your pull request 🙏. It looks good! I just have a few minor questions and some suggestions for improvements.

src/Testcontainers.Nats/NatsBuilder.cs
niklasfp: Let me check, I think this from one of the other modules, maybe mssql 😄 
Edit, I'll just remove it 👍

src/Testcontainers.Nats/NatsBuilder.cs
niklasfp: As a default, null is fine, nats default runs without the username / password.
About the escaping, I was betting on the UriBuilder to do the work, and I might be wrong, so I'll check that one as well 👍

src/Testcontainers.Nats/NatsBuilder.cs
niklasfp: Omg, so embarrassing, I'll fix right away, and do the same for --user

src/Testcontainers.Nats/NatsContainer.cs
niklasfp: Nopes, from: https://docs.nats.io/running-a-nats-service/configuration/monitoring
>Warning: nats-server does not have authentication/authorization for the monitoring endpoint. When you plan to open your nats-server to the internet make sure to not expose the monitoring port as well. By default monitoring binds to every interface 0.0.0.0 so consider setting monitoring to localhost or have appropriate firewall rules.

src/Testcontainers.Nats/NatsBuilder.cs
HofmeisterAn: > As a default, null is fine, nats default runs without the username / password.

Yes, I saw that according to the configuration. Setting a default non-null value make it just simpler in case the encoding is necessary.

niklasfp: I would, lean towards being as true to the image as possible, meaning no user and password except if explicitly set.
and check for nulls before encoding ?
Oh and yeah, UriBuilder does not encode 😄 

I can do the validation to make sure you don't end up setting only password but not user?

src/Testcontainers.Nats/NatsBuilder.cs
niklasfp: Added the longer ones.

src/Testcontainers.Nats/NatsBuilder.cs
niklasfp: Removed.

src/Testcontainers.Nats/NatsBuilder.cs
niklasfp: Added validation and escaping.

src/Testcontainers.Nats/NatsBuilder.cs
niklasfp: Now uses --user and --pass (I'll go hide in a bush for some time).

HofmeisterAn: > I'll go hide in a bush for some time

But not for this 😄, it is not a problem. I will take a look at the changes tomorrow, thanks already for addressing them.

src/Testcontainers.Nats/NatsBuilder.cs
```diff
@@ -0,0 +1,115 @@
...
+        base.Validate();
+        
+        if (DockerResourceConfiguration.Password != null || DockerResourceConfiguration.Username != null)
+        {
+            _ = Guard.Argument(DockerResourceConfiguration.Username, nameof(DockerResourceConfiguration.Username))
+                .NotNull();
+
+            _ = Guard.Argument(DockerResourceConfiguration.Password, nameof(DockerResourceConfiguration.Password))
+                .NotNull();
+        }
```
HofmeisterAn: QQ: What do you think about a similar approach we use in the MongoDB module and set `string.empty` as default values for `username` and `password` and add these validation checks:

https://github.com/testcontainers/testcontainers-dotnet/blob/0c6588ed28b2bc14c19373e029f3421949440247/src/Testcontainers.MongoDb/MongoDbBuilder.cs#L92-L99

It makes the subsequent null-checks unnecessary. Note, the `UriBuilder` returns a Uri without `:` if the `username` and `password` is empty, see: https://github.com/testcontainers/testcontainers-dotnet/pull/983#issuecomment-1700376825.

niklasfp: I'll ditch the null checks.

niklasfp:
@HofmeisterAn How should we handle this, you can rebase and merge if you see it fitting, or let me know what you need done from me 😄

HofmeisterAn:
> @HofmeisterAn How should we handle this, you can rebase and merge if you see it fitting, or let me know what you need done from me 😄

Ah, perhaps I misinterpreted it. I thought you were going to make changes according to the suggestion.

> I'll ditch the null checks.

I am fine with the pull request; it looks good 👍, thanks! I can merge it later today when I am back home. Rebase, etc. is not necessary.

niklasfp:
@HofmeisterAn cool!!

HofmeisterAn (approved):
Thanks, again 🙏.

PR #1009, opened 2023-10-02
feat: Extend the "wait until file exists" API to distinguish between the test host and container filesystem

maaex:
## What does this PR do?
The existing implementation of the UntilFileExists WaitStrategy seems to look for a file in the host's file system instead of the container's.
This PR changes that so UntilFileExists looks for the file in the container's file system instead.

## Why is it important?
Looking for a file on the host instead of the container seems like incorrect behavior.

### DIFF
--- a/src/Testcontainers/Configurations/FileSystem.cs
+++ b/src/Testcontainers/Configurations/FileSystem.cs
@@ -9,7 +9,7 @@ namespace DotNet.Testcontainers.Configurations
   public enum FileSystem
   {
     /// <summary>
-    /// The file system of the host machine.
+    /// The test host file system.
     /// </summary>
     Host = 0,
 
--- a/src/Testcontainers/Configurations/FileSystem.cs
+++ b/src/Testcontainers/Configurations/FileSystem.cs
@@ -1,0 +1,21 @@
+namespace DotNet.Testcontainers.Configurations
+{
+  using JetBrains.Annotations;
+
+  /// <summary>
+  /// Indicates the file system for file operations.
+  /// </summary>
+  [PublicAPI]
+  public enum FileSystem
+  {
+    /// <summary>
+    /// The file system of the host machine.
+    /// </summary>
+    Host = 0,
+
+    /// <summary>
+    /// The container file system.
+    /// </summary>
+    Container = 1,
+  }
+}
--- a/src/Testcontainers/Configurations/WaitStrategies/IWaitForContainerOS.cs
+++ b/src/Testcontainers/Configurations/WaitStrategies/IWaitForContainerOS.cs
@@ -55,7 +55,7 @@ namespace DotNet.Testcontainers.Configurations
     /// <param name="file">The file to be checked.</param>
     /// <returns>A configured instance of <see cref="IWaitForContainerOS" />.</returns>
     [PublicAPI]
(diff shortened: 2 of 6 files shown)


HofmeisterAn (changes requested):
Thank you for your pull request. This is some kind of on purpose, but I understand the confusion regarding the wait strategy. I would like to suggest the following in order to avoid breaking existing configurations: 

1. Enhance the `UntilFileExists(string)` documentation (interface) and make it more explicit that it checks for a file on the test host.
2. Add an overloaded method that checks for the file inside the container.

We can also introduce two explicit APIs (wait strategies): One that checks on the test host and another that checks in the container, while deprecating the old one. My goal is to avoid introducing breaking changes without informing the developers.

maaex:
> Thank you for your pull request. This is some kind of on purpose, but I understand the confusion regarding the wait strategy. I would like to suggest the following in order to avoid breaking existing configurations:
> 
> 1. Enhance the `UntilFileExists(string)` documentation (interface) and make it more explicit that it checks for a file on the test host.
> 2. Add an overloaded method that checks for the file inside the container.
> 
> We can also introduce two explicit APIs (wait strategies): One that checks on the test host and another that checks in the container, while deprecating the old one. My goal is to avoid introducing breaking changes without informing the developers.

Good point. Should I move the code into a `UntilFileExistsInContainer(string)` method (+ a related WaitStrategy class) maybe? That naming should make the distinction clearer between the two.

HofmeisterAn:
> Good point. Should I move the code into a `UntilFileExistsInContainer(string)` method (+ a related WaitStrategy class) maybe? That naming should make the distinction clearer between the two.

Yes, I think that would be a better and more explicit approach. I am currently thinking how we can enhance the wait strategy API in a broader sense, without the need for a complete overhaul. The wait strategy API has remained unchanged since its early days. Perhaps an approach similar to the `HttpWaitStrategy`, which allows more configurations, would be a more future-proof solution. We could consider implementing something like a `FileSystemWaitStrategy`, which accept a file path and a scope to determine whether it is a file within a container or a file on the test host.

maaex:
I see that the `HttpWaitStrategy` has different usage pattern to the 'normal' wait strategies. Do we need to go there with this? Maybe something like `Wait.ForUnixContainer().UntilFileExists("/foo/bar", FileSystem.Host)` would suffice? And have the default scope value as Host to avoid breaking any existing functionality.

And do you have any suggestions for what to call the scope enum?

HofmeisterAn:
> Do we need to go there with this? Maybe something like

No, I think your suggestion is reasonable.

> And do you have any suggestions for what to call the scope enum?

`FileSystem.Host`, `FileSystem.Container` sounds good 👍.

maaex:
I have updated the code as per our discussion. I was unsure of where to put the `FileSystem` enum but ended up putting it under `TestContainers.Configurations` as I felt it was a more generic thing than WaitStrategies. What do you think?

HofmeisterAn:
I took a brief look. It looks very good. Great first contribution, thanks. I won't be able to finalize the review today, but I can certainly do it tomorrow.

HofmeisterAn (approved):
Thanks for the PR and addressing the requests. Great first contribution, thanks :+1:.

PR #1044, opened 2023-11-08
feat: Add Papercut module

TechLiam:
## What does this PR do?
Add Testcontainers.Papercut

## Why is it important?
Papercut is a testing SMTP server that also has a nice, easy REST API to get the messages in the inbox.
When doing integration testing it is helpful to be able to send emails and have the ability to check their contents by having this container developers will be able to quickly spin up an SMTP server and check the messages delivered to it without having to send emails to a proper SMTP server and email address.

## How to test this PR
Unit/Integration tests have been added to the solution that checks the container starts and can have emails sent to it. It also checks the helper methods for getting the emails delivered back out of Papercut from the REST API.


src/Testcontainers.Papercut/PapercutContainer.cs
```diff
@@ -0,0 +1,76 @@
...
+        client.BaseAddress = new Uri(WebUrl);
+        return client.GetAsync($"/api/messages/{id}/sections/{index}");
+    }
+
+    public Task<HttpResponseMessage> DownloadSectionContent(string id, string contentId)
+    {
+        var client = new HttpClient();
+        client.BaseAddress = new Uri(WebUrl);
+        return client.GetAsync($"/api/messages/{id}/contents/{contentId}");
+    }
```
HofmeisterAn: Is there no Papercut client or, at the very least, a dependency that holds the domain models? Is there a publicly available API document we can refer to? The Testcontainers modules do not implement client functionalities; instead, we focus on spinning up dependencies for various use cases. Can we remove these client functionalities? We can move and keep the necessary domain models to the test and simply send the required HTTP requests from there.

src/Testcontainers.Papercut/PapercutBuilder.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        Validate();
+        return new PapercutContainer(DockerResourceConfiguration, TestcontainersSettings.Logger);
+    }
+
+    /// <inheritdoc />
+    protected override PapercutBuilder Init()
+    {
+        return base.Init()
+            .WithImage("jijiechen/papercut:latest")
```
HofmeisterAn: Please use a const and pin the image version, as is done with all the other modules.

src/Testcontainers.Papercut/PapercutBuilder.cs
```diff
@@ -0,0 +1,71 @@
...
+            .WithPortBinding(25,true)
+            .WithPortBinding(37408,true)
```
HofmeisterAn: Please use a const and a meaningful name that describes the port.

src/Testcontainers.Papercut/PapercutBuilder.cs
```diff
@@ -0,0 +1,71 @@
+using System.Net;
```
HofmeisterAn: Move all usings to the global use file.

tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return _papercutContainer.DisposeAsync().AsTask();
+    }
+
+    [Fact]
+    [Trait(nameof(DockerCli.DockerPlatform), nameof(DockerCli.DockerPlatform.Linux))]
+    public async Task SendingAnEmail()
+    {
+        //Given
+        var client = new SmtpClient(_papercutContainer.Hostname,_papercutContainer.GetMappedPublicPort(25));
```
HofmeisterAn: WDYT about a delegate to access the SmtpPort?

HofmeisterAn (changes requested):
Thanks for your PR. Definitely a cool module. Can we please remove the Papercut client functionalities? This isn't the responsibility of Testcontainers.

src/Testcontainers.Papercut/PapercutContainer.cs
TechLiam: Cool, yeah, that makes sense responsibility wise all the container package is doing is creating the container, and interacting with it is outside the scope of this package. I have removed the HTTP Client calls from the package code (moved some/most of it to the tests)

src/Testcontainers.Papercut/PapercutBuilder.cs
TechLiam: Moved it to a const not sure how to pin the version as it's not tagged in the docker hub had a look at how the other packages are doing this and could see that some of the packages e.g. Testcontainers.CosmosDb are using :latest

I'm reasonably new to docker stuff though so if there is a way to do this please let me know I'm happy to add it in

src/Testcontainers.Papercut/PapercutBuilder.cs
TechLiam: Done

src/Testcontainers.Papercut/PapercutBuilder.cs
TechLiam: Done

tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs
TechLiam: I agree this is better so have added it in

src/Testcontainers.Papercut/PapercutBuilder.cs
HofmeisterAn: > not sure how to pin the version as it's not tagged

👍 Okay, then `latest` is fine. I didn't check the image tags. Thanks for checking.

TechLiam:
@HofmeisterAn is there anything else or is this PR good to merge now?

HofmeisterAn:
> @HofmeisterAn is there anything else or is this PR good to merge now?

I've made some improvements. WDYT? If you're good with it, we are ready to merge.

tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs
```diff
@@ -16,33 +16,57 @@ public Task DisposeAsync()
...
+            messages = jsonDocument.RootElement.GetProperty("messages").Deserialize<Message[]>();
+        }
+        while (messages.Length == 0);
+
+        // Then
+        Assert.NotEmpty(messages);
+        Assert.Equal(subject, messages[0].Subject);
+    }
+
+    private readonly struct Message
```
TechLiam, line 49: I like this struct hadn't thought about doing it with that good call

HofmeisterAn (dismissed):
I guess you're good with the recent changes. Thanks for the pull request. Great first Testcontainers contribution 👍. I will merge it later the day.

PR #1051, opened 2023-11-14
feat: Add reuse support

### LINKED ISSUE #407: [Question] Use existing container if it exists
I'm using this library to setup containers for integration tests for postgres. It works very well.

However it is slow to start and stop, as to be expected. So what I'm trying to achieve:
- On my dev machine, I have another instance of postgres running (almost) all the time. If it is running then I'd like the library to use it.
- On the test server, no containers would be running, so I'd like the library to start/stop throwaway containers as usual. 

Is this possible somehow?

(I could add boilerplate code to detect if the container is running, etc,, but I'm hoping this library has functionality to do that?)

david-szabo97:
## Related issues

- Relates #407


src/Testcontainers/Builders/ContainerBuilder`3.cs
```diff
@@ -329,6 +330,16 @@ public TBuilderEntity WithAutoRemove(bool autoRemove)
       return Clone(new ContainerConfiguration(autoRemove: autoRemove));
     }
 
+    /// <inheritdoc />
+    public TBuilderEntity WithReuse(bool reuse)
+    {
+      return Clone(new ContainerConfiguration(reuse: reuse)).WithCreateParameterModifier(createContainerParameters =>
+      {
+        createContainerParameters.Labels.Remove(TestcontainersClient.TestcontainersSessionIdLabel);
+        createContainerParameters.Labels.Remove(ResourceReaper.ResourceReaperSessionLabel);
```
david-szabo97: These labels are changing on each run. I'm not sure why they are in the `DefaultLabels` though. We need to remove them to keep the hash consistent.

They should be added via `WithCleanUp()` and `WithAutoRemove()`.

src/Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -208,11 +211,41 @@ public async Task<string> RunAsync(IContainerConfiguration configuration, Cancel
...
+            },
+          Limit = 1,
+        }, ct)
+          .ConfigureAwait(false);
+
+        var container = containers.FirstOrDefault();
+        if (container != default) {
+          return container.ID;
+        }
+      }
```
david-szabo97: It feels like there are better locations for this. But since this is the only location where we can access the `CreateContainerParams` used to create the container, I put the reuse logic here.

src/Testcontainers/Clients/DockerContainerOperations.cs
```diff
@@ -208,11 +211,41 @@ public async Task<string> RunAsync(IContainerConfiguration configuration, Cancel
...
     }
+
+    static string CreateReuseHash(CreateContainerParameters createContainerParameters)
+    {
+      var createContainerParametersJson = System.Text.Json.JsonSerializer.Serialize(createContainerParameters);
+
+      using (var sha1 = SHA1.Create())
+      {
+        return Convert.ToBase64String(sha1.ComputeHash(Encoding.Default.GetBytes(createContainerParametersJson)));
+      }
```
david-szabo97: Is there any preference for the hash algorithm and string format? Wanted to keep it simple, that's why `Convert.ToBase64String` is used here.

david-szabo97:
Currently, it's possible to use `WithAutoRemove(true)` and `WithCleanUp(true)` together with `WithReuse(true)`. Should we add some restrictions in there to prevent incorrect configuration?

src/Testcontainers/Clients/DockerContainerOperations.cs
david-szabo97: Maybe this should belong to `DockerContainer.UnsafeCreateAsync()`?

src/Testcontainers/Builders/ContainerBuilder`3.cs
HofmeisterAn: > These labels are changing on each run. I'm not sure why they are in the `DefaultLabels` though. We need to remove them to keep the hash consistent.

Exactly. We need to ignore some properties while generating the hash value. Please see my more detailed answer below.

src/Testcontainers/Clients/DockerContainerOperations.cs
HofmeisterAn: Interesting idea. I have not thought about using the `createParameters` instance. I was thinking about extending the `ResourceConfiguration` class and its implementations. I believe this has several advantages:

1. We need to calculate the hash value for the other resources (like network and volume) too. Each resource configuration derives from this class.
2. It is simpler to test. We can expose generating the hash value for each resource configuration through the public abstract `Resource` class (the resource interfaces). Therefore we do not need to create the resource at all.

    ```csharp
    _ = new ContainerBuilder().Build().GetHash();
    _ = new VolumeBuilder().Build().GetHash();
    ```
3. This should cover all resources at once. We probably "just" need to ignore individual properties of each resource configuration from being considered in calculating the hash value if they change (like the session ID).

src/Testcontainers/Clients/DockerContainerOperations.cs
HofmeisterAn: I have no preference. Maybe we can switch to SHA256, but it probably does not really matter.

HofmeisterAn:
Thanks for bringing this up and looking into it!

I like the idea of utilizing JSON serialization. Maybe we can build on top of that and ignore specific properties using JSON paths or at least implement something that removes configurations like the session ID, etc. before calculating the hash (the final create resource object that is sent to the Docker API requires those labels).

> Currently, it's possible to use `WithAutoRemove(true)` and `WithCleanUp(true)` together with `WithReuse(true)`. Should we add some restrictions in there to prevent incorrect configuration?

Yes, we should throw an exception for invalid configurations. I think we should do this in one of the `Validate()` base implementations.

david-szabo97:
Thanks @HofmeisterAn for all your inputs! Rewrote the PR with those in mind.

In the end, the implementation:
- `GetHash()` added to `IResourceConfiguration`; each `ResourceConfiguration` must implement this function to provide reusability.
- `WithReuse(true)` only works if it's the last method called in the builder. `WithReuse` adds the label to the resource containing the hash. Therefore, if `WithReuse` is not the last method, the hash might change before calling `Build()`. An exception is thrown if the hash mismatches to help in debugging.
- I replaced `System.Text.Json` with Newtonsoft, which supports JSONPath and is also a mutable object. Which we could use to build a modifier pipeline for the JSON object.  `WithGetHashModifier(Action<JObject>)` ?

Please let me know what you think about this implementation. Once we agree on an implementation, I'll implement the remaining `GetHash()` methods.

Does it make sense to have reusability for `FutureDockerImage`?

HofmeisterAn:
Thank you for responding and addressing the input so quickly. I believe the overall implementation represents well what I had in mind 👍. We just need to improve a few things and probably find the right places for some other parts.

I have not tested the changes locally or debugged them to gain a better understanding yet. I will do that in the next few days. My suggestions are not set in stone; they are just inputs and ideas to iterate over. I am certain we will find better solutions together.

> `GetHash()` added to `IResourceConfiguration`; each `ResourceConfiguration` must implement this function to provide reusability.

That makes sense; it is along the lines of what I had in mind too.

> `WithReuse(true)` only works if it's the last method called in the builder. `WithReuse` adds the label to the resource containing the hash. Therefore, if `WithReuse` is not the last method, the hash might change before calling `Build()`. An exception is thrown if the hash mismatches to help in debugging.

We probably cannot rely on the order. Referring to some modules, it is necessary to adjust the configuration before returning the built resource. Usually, this is done to properly preconfigure the [wait strategies](https://github.com/testcontainers/testcontainers-dotnet/blob/ab2c7c9e6802ac6df1694f103187536a39bd2ed3/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#L124), but I cannot promise it will not be used for other properties in the future. However, we can sort that out later; we will figure something out to append the reuse hash.

> I replaced `System.Text.Json` with Newtonsoft, which supports JSONPath and is also a mutable object.

I would prefer to avoid adding another JSON library or additional dependencies in general. I try to minimize dependencies to make developers' lives easier, ensuring they do not encounter dependency clashes or similar issues. 

Perhaps we can utilize annotations to configure JSON serialization, like `JsonIgnore` and `JsonConverter` (to ignore certain keys, etc.). Then we can simply serialze the configuration instance. I believe this has another advantage. It allows us to simply create the reuse hash for module configurations too.

> Which we could use to build a modifier pipeline for the JSON object. `WithGetHashModifier(Action<JObject>)` ?

Yes, I like that idea as a follow-up. Allowing developers to provide their own hash implementation if necessary sounds good. Maybe we can use an interface for that.

> Does it make sense to have reusability for `FutureDockerImage`?

I do not think it is necessary for images. Developers can already reuse them.

kiview:
How is the reuse of volumes intended to work? By volume name? Or will volumes implement a `getHash()` function that takes their content into account?

It might be better to out-scope volumes for the initial implementation.

HofmeisterAn:
> How is the reuse of volumes intended to work? By volume name? Or will volumes implement a `getHash()` function that takes their content into account?
> 
> It might be better to out-scope volumes for the initial implementation.

If we follow the approach, TC for .NET will basically identify networks and volumes based on their names (not many other properties are included here in calculating the hash). We do not offer many other public APIs to create and configure volumes and networks. Essentially, the reuse hash prevents Ryuk from deleting them.

kiview:
In general when using names for volumes and networks, as soon as reuse disabled, don't we run into potential issues with conflicting names?

Independent of this topics, can we please add the global opt-in mechanism for the reuse feature according to how tc-java [does it](https://java.testcontainers.org/features/reuse/#how-to-use-it)? This mechanism would automatically integrate with Testcontainers Desktop's `Enable Reusable Containers` feature, which will set this flag.

HofmeisterAn:
> In general when using names for volumes and networks, as soon as reuse disabled, don't we run into potential issues with conflicting names?

Yep, it does. There are several ways we can deal with it, even make it configurable (.NET can partially already reuse existing resources), but for now, I would like to focus on generating and assigning the hash consistently. Anything afterward is just agreeing on the workflow.



> Independent of this topics, can we please add the global opt-in mechanism for the reuse feature according to how tc-java [does it](https://java.testcontainers.org/features/reuse/#how-to-use-it)?

👍

david-szabo97:
@HofmeisterAn Removed Made a couple of changes. I'm not quite sure what we should `JsonIgnore` yet, but so far it seems to work fine with my simple test case. Added volumes and networks to the use case to cover more.

I've been thinking a lot about how to avoid relying on the order of the `WithReuse`. I figured out we could use a create parameter modifier for this. https://github.com/testcontainers/testcontainers-dotnet/pull/1051/commits/7994956dabc6e6a44c7ec577f62ea930f5c78ead let me know what you think about this, we can always revert it and figure out something better. TBH, I don't really like the implementation because of the reflection, but we can get rid of the reflection and just cast it to the right type. The current implementation is enough as a proof-of-concept.

HofmeisterAn:
> I'm not quite sure what we should `JsonIgnore` yet, but so far it seems to work fine with my simple test case.

I think the current configuration looks ok (we probably can ignore more). I will look into it more closely. Initially, we should only consider a few properties to keep it simple and gradually add support for more properties. For instance, enabling the `ResourceMappings` configuration would necessitate generating a hash value for the interface implementations as well (which I would like to avoid at the beginning).

> I figured out we could use a create parameter modifier for this. [7994956](https://github.com/testcontainers/testcontainers-dotnet/commit/7994956dabc6e6a44c7ec577f62ea930f5c78ead) let me know what you think about this, we can always revert it and figure out something better.

I will take a look at it 👍.

tests/Testcontainers.Platform.Linux.Tests/ReusableResourceTest.cs
```diff
@@ -0,0 +1,80 @@
...
+                .WithReuse(true)
+                .Build();
+
+            var volume = new VolumeBuilder()
+                .WithName(_labelKey)
+                .WithLabel(_labelKey, _labelValue)
+                .WithReuse(true)
+                .Build();
+
+            await Task.WhenAll(container.StartAsync(), network.CreateAsync(), volume.CreateAsync())
```
HofmeisterAn, line 44: The only thing I'm not really happy with is that _reuse_ doesn't support or might run into race conditions when creating and starting the same configuration in parallel. Since reuse shouldn't be used within the same processes, this isn't a huge issue, I guess.

HofmeisterAn (commented):
I reviewed the PR, tested a couple of things, and applied some changes and improvements. Overall, it looks very good already, just a few minor things left (I think).

I agree, it's not that easy to decide which properties should be ignored and which should not. As mentioned, I would prefer to opt-in supported properties, but unfortunately, this is not supported by _System.Text.Json_ (compared to _Json.NET_). I prefer opt-in because I am certain that not all properties will support the reuse feature without custom serialization. Not being able to opt-in is especially unfortunate for modules (luckily, most of their properties are not composed types nor very complex). I ignored most of the properties for the generic builders, which we can undo soon as we are sure that they won't cause an issue in generating the hash (all necessary properties available (serializable) to calculate the hash).

What do you think about the latest changes?

> TBH, I don't really like the implementation because of the reflection, but we can get rid of the reflection and just cast it to the right type.

I moved this part to the classes that create the actual resource. Then no reflection is necessary. It's a limitation of `Docker.DotNet` and the provided types. There's no inheritance; we don't need to create it artificially through reflection.

HofmeisterAn (approved):
@david-szabo97, thank you for the contribution and the efforts you put into this feature. In my opinion, it is a really great contribution. I did a few more underlying tests, and I believe it works very well already. It serves as a good foundation to continue adding new features (use cases). The next reasonable follow-ups will be:

- Adding the custom property `testcontainers.reuse.enable` (as Kevin mentioned).
- Including further builder configuration properties in calculating the hash value (using JSON custom serializers).

Thanks again 🙏.

david-szabo97:
Thanks for finishing it up! Looking forward to the next release 🚀

304NotModified:
When could we expect a release for this great feature? :)

HofmeisterAn:
No ETA yet. I would like to include the upcoming changes regarding the logging builder API in the next version as well. If the remaining PRs take longer, I think I can publish a pre-release, if that helps. Please keep in mind that this feature is experimental, and not all modules may be compatible or work out of the box.

mkedrzynskiAndea:
Having pre-release would be great

304NotModified:
Pre-release would be great indeed!

HofmeisterAn:
[Here](https://www.nuget.org/packages/Testcontainers/3.8.0-beta.8238206979) you are. Sorry for the delay; I wanted to ship the logging API changes in the beta version too. Before using the reuse feature, please double-check the [docs](https://github.com/testcontainers/testcontainers-dotnet/blob/36758c84856af442ef1f8d9f3562d8236a461a0d/docs/api/resource_reuse.md) and make sure you are familiar with the limitations.

304NotModified:
Works great!

(Tested it with Testcontainers.MsSql and a fixed container name)

PR #1053, opened 2023-11-22
Make ESDB container implement IDatabaseContainer

alexeyzimarev:
## What does this PR do?

`EvebtStoreDbContainer` class implements the `GetConnectionString` method so it can inherit from `IDatabaseContainer`.

## Why is it important?

When creating an abstraction for tests, I can't currently use `IDatabaseContainer` and its `GetConnectionString` because `EventStoreDbContainer` doesn't implement the interface.

### DIFF
--- a/src/Testcontainers.EventStoreDb/EventStoreDbContainer.cs
+++ b/src/Testcontainers.EventStoreDb/EventStoreDbContainer.cs
@@ -2,7 +2,7 @@ namespace Testcontainers.EventStoreDb;
 
 /// <inheritdoc cref="DockerContainer" />
 [PublicAPI]
-public sealed class EventStoreDbContainer : DockerContainer
+public sealed class EventStoreDbContainer : DockerContainer, IDatabaseContainer
 {
     /// <summary>
     /// Initializes a new instance of the <see cref="EventStoreDbContainer" /> class.


HofmeisterAn (commented):
The interface represents a database container instance that can be accessed with an ADO.NET provider. Does ESDB support ADO.NET? We [test](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs#L5-L17) against the db client.

alexeyzimarev:
I already figured it out by looking at failing test, and I am not exactly following that decision. The interface doesn't suggest anything like that, so it's unclear from the API surface what is the intention behind that interface.

From the test perspective, and looking at other container implementations, it looks like MonoDB for example, is _not_ a database when we all know it definitely _is_ a database. Or Neo4j, or ESDB in that regard. I personally find it confusing.

HofmeisterAn:
> The interface doesn't suggest anything like that, so it's unclear from the API surface what is the intention behind that interface.

Indeed, the interface itself does not, but the description does. IIRC, the initial idea was `IAdoNetDatabaseContainer` or something similar.

> From the test perspective, and looking at other container implementations, it looks like MonoDB for example, is not a database when we all know it definitely is a database.

No one is saying that MongoDB or ESDB are not databases, but abstraction becomes significant when commonality exists. You cannot easily swap MongoDB with ESDB as you can with ADO.NET compatible containers (AFAIK). But it is definitely ambiguous. Such as the publicly available methods to retrieve the connection string, endpoint, base address, etc. Typically, I aim to correspond to the client library.

> When creating an abstraction for tests, I can't currently use `IDatabaseContainer` and its `GetConnectionString` because `EventStoreDbContainer` doesn't implement the interface.

Could you provide more details about what you are trying to abstract? I would like to get a better understanding of the use cases.

alexeyzimarev:
I have a base class that is used across different implementation, which are database-specific. In my case, the abstraction is on a high enough level to just need the `GetConnectionString` be available across all those containers so I can pass the connection string to downstream implementations without exposing much other details about the container implementation.

I ended up making the base class generic, where the generic parameter is constrained to `DockerContainer`, and all inheritors use their own container types, and be able to call `Container.GetConnectionString()`. So, I had to expose container as a protected property, which I didn't want to.

HofmeisterAn:
Maybe we can add support together with this [issue](https://github.com/testcontainers/testcontainers-dotnet/issues/1060#issuecomment-1845687358). Right now, I am considering a connection string provider or something similar, which is capable of resolving the correct connection strings. This would allow developers to override the default connection string and could be used for the abstraction too.

HofmeisterAn:
As mentioned, the interface is exclusively for ADO.NET compatible containers. I've generated a follow-up [issue](https://github.com/testcontainers/testcontainers-dotnet/issues/1074) that includes a proposal for implementing a connection string provider across all modules. I'll close this PR. Let's carry on the discussion in the referenced issue. Thank you for bringing this to my attention.

PR #1054, opened 2023-11-22
chore: Update .NET SDK to version 8 (LTS)

HofmeisterAn:
## What does this PR do?

Updates the .NET SDK to version 8 (latest LTS).

## Why is it important?

We are trying to run our tests utilizing the latest .NET SDK (LTS). Testcontainers itself, including the modules, still targets `netstandard2.0` and `netstandard2.1` to provide support for most .NET Framework and .NET versions.

## Related issues

\-

### DIFF
--- a/Packages.props
+++ b/Packages.props
@@ -9,9 +9,9 @@
     <PackageReference Update="BouncyCastle.Cryptography" Version="2.2.1" />
     <PackageReference Update="SharpZipLib" Version="1.4.2" />
     <PackageReference Update="SSH.NET" Version="2023.0.0" />
-    <PackageReference Update="System.Text.Json" Version="6.0.8" />
+    <PackageReference Update="System.Text.Json" Version="6.0.9" />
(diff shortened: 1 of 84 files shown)


0xced:
Note that SourceLink is now [included in the .NET SDK](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#source-link):

> [Source Link](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink) is now included in the .NET SDK. The goal is that by bundling Source Link into the SDK, instead of requiring a separate `<PackageReference>` for the package, more packages will include this information by default. That information will improve the IDE experience for developers.

Since the `global.json` file requires the .NET 8 SDK, all the package references to `Microsoft.SourceLink.GitHub` could be removed instead of updated to version 8.0.0.

HofmeisterAn:
Thanks for sharing. I will remove it in the next few days 🧹.

PR #1063, opened 2023-12-01
feat(Azurite): Add in-memory support

### LINKED ISSUE #1056: [Enhancement]: Upgrade to Azurite >= 3.28.0 and support in-memory persistence
### Problem

> The current usage pattern of Azurite assumes that the caller is interested in persisting stored data beyond the lifetime of the node process. This certainly is useful for extended testing scenarios or where Azurite is actually used for production-like scenarios (perhaps a best-effort approach of free, on-prem Azure Storage). However, some Azurite use cases are more ephemeral in nature. For example, if you simply want to use Azurite for a single test run in a CI pipeline, you don't need the storage to persist beyond the lifetime of the node process.

Quote from: https://github.com/Azure/Azurite/blob/105a9cd217b881922016edb7a683ebea1ab3b28a/docs/designs/2023-10-in-memory-persistence.md

Depending on the amount and size of for example a blob there is an additional overhead on the test environment when Azurite has to write it to disk. In most scenarios, test containers won't survive more than one test run. So the overhead of writing to disk could be removed.

### Solution

Upgrade to Azurite 3.28.0 or later and add support for the `--inMemoryPersistence` argument

### Benefit

Using in-memory persistence may have a positive effect on overall performance of Azurite during tests, as it does not need to store stuff on disk.

### Alternatives

Leave the current implementation as is

Yes

### LINKED ISSUE #1060: [Enhancement]: Add method to return CosmosDb connection string for internal network
### Problem

Right now, `CosmosDbContainer` has a `string GetConnectionString()` method which returns a connection string for use on the host. It would be useful if `CosmosDbContainer` also had a method returning a connection string usable within a network the container is attached to. Ideally, this method would work before the container is started so that it can be used to build other containers which rely on this connection string. 

### Solution

I propose adding a method to class `CosmosDbContainer` which is functionally equivalent to:

```csharp
private string GetInternalConnectionString(string networkAlias)
{
    var properties = new Dictionary<string, string>
    {
        ["AccountEndpoint"] = new UriBuilder(Uri.UriSchemeHttps, networkAlias, CosmosDbBuilder.CosmosDbPort).ToString(),
        ["AccountKey"] = CosmosDbBuilder.DefaultAccountKey
    };
    return string.Join(";", properties.Select(kvp => string.Join("=", kvp.Key, kvp.Value)));
}
```

### Benefit

This method will make it easier to create connection strings used to build other containers which communicate with the cosmos container.

### Alternatives

This connection string can be created using the code snippet proposed above, but results in duplicated code.

Yes

Xzelsius:
## What does this PR do?

Updates the default Azurite image to v3.28.0
Additionally, adds support for the new in-memory persistence flag `--inMemoryPersistence`

The new in-memory persistence is opt-in and needs to be enabled using `.WithInMemoryPersistence()` on the Azurite builder.
As default, the memory limit is 50% of the available Docker host RAM, but it can be overridden by specifying it in the `.WithInMemoryPersistence(<memory-limit-in-mb>)` call.

## Why is it important?

Using in-memory persistence may have a positive effect on overall performance of Azurite during tests, as it does not need to store stuff on disk.

## Related issues

- Closes #1056


src/Testcontainers.Azurite/AzuriteBuilder.cs
```diff
@@ -40,13 +40,47 @@ private AzuriteBuilder(AzuriteConfiguration resourceConfiguration)
...
+        {
+            "azurite",
+            "--blobHost",
+            "0.0.0.0",
+            "--queueHost",
+            "0.0.0.0",
+            "--tableHost",
+            "0.0.0.0",
+            "--inMemoryPersistence"
+        };
```
HofmeisterAn: Can we please move this part to the builder `Init()` method?

```csharp
.WithEntrypoint("azurite")
.WithCommand("--blobHost", "0.0.0.0", "--queueHost", "0.0.0.0", "--tableHost", "0.0.0.0");
```

src/Testcontainers.Azurite/AzuriteBuilder.cs
```diff
@@ -40,13 +40,47 @@ private AzuriteBuilder(AzuriteConfiguration resourceConfiguration)
...
+
+        if (memoryLimit.HasValue)
+        {
+            command.Add("--extentMemoryLimit");
+            command.Add(memoryLimit.ToString());
+        }
+
+        return Merge(DockerResourceConfiguration, new AzuriteConfiguration(inMemoryPersistence: true, extentMemoryLimit: memoryLimit))
+            .WithCommand(command.ToArray());
```
HofmeisterAn: Then we can replace this lines with:

```csharp
if (memoryLimit.HasValue)
{
    return WithCommand("--inMemoryPersistence", "--extentMemoryLimit", memoryLimit.ToString());
}
else
{
    return WithCommand("--inMemoryPersistence");
}
```

src/Testcontainers.Azurite/AzuriteBuilder.cs
```diff
@@ -40,13 +40,47 @@ private AzuriteBuilder(AzuriteConfiguration resourceConfiguration)
...
     {
         Validate();
 
         var waitStrategy = Wait.ForUnixContainer();
 
+        if (DockerResourceConfiguration.InMemoryPersistence)
+        {
+            waitStrategy = waitStrategy.UntilMessageIsLogged("In-memory extent storage is enabled");
+        }
+
```
HofmeisterAn: I do not expect we need this wait strategy. I expect Azurite to be ready when the services are available.

src/Testcontainers.Azurite/AzuriteConfiguration.cs
```diff
@@ -7,8 +7,12 @@ public sealed class AzuriteConfiguration : ContainerConfiguration
     /// <summary>
     /// Initializes a new instance of the <see cref="AzuriteConfiguration" /> class.
     /// </summary>
-    public AzuriteConfiguration()
+    /// <param name="inMemoryPersistence">Indicates whether in-memory persistence is enabled.</param>
+    /// <param name="extentMemoryLimit">Optional memory-limit when using in-memory persistence.</param>
+    public AzuriteConfiguration(bool inMemoryPersistence = false, int? extentMemoryLimit = null)
```
HofmeisterAn: I do not think we need to keep the values. As long as we do not need them later (e.g. in the container instance) it is not necessary to keep them in the configuration instance. At least that's how we do it in other modules.

HofmeisterAn (changes requested):
Thanks for the PR 👍 I have some minor change requests, but overall, it looks really good. I will be happy to merge it once the requested changes are addressed.

HofmeisterAn:
I made slight changes to the tests. Instead of creating a new class, we can reuse the previous test class. Does the PR look good to you? What do you think? I am ready to merge it. Thanks for your contribution.

Xzelsius:
@HofmeisterAn That's actually a way better solution than what I did with the test :)
Looks good to me 👍

HofmeisterAn (dismissed):
Thanks for creating the issue and PR 🙏.

Issue #1066, opened 2023-12-07
[Enhancement]: Renaming a file (very minor)


mgroves:
### Problem

In the tests/Testcontainers.Couchbase.Tests folder, there is a file named CouchDbContainerTest.cs. The code itself is fine, there's a class called `CouchbaseContainerTest`. However, I think that class names and file names should match. Further, this might cause confusion (of which there is already a considerable amount) that CouchDb and Couchbase are the same thing.

### Solution

File rename from CouchDbContainerTest.cs -> CouchbaseContainerTest.cs

### Benefit

Reduction of potential confusion

### Alternatives

I can't think of an alternative, other than leaving it the way it is.

### Would you like to help contributing this enhancement?

Yes

HofmeisterAn:
Thanks for the PR.

PR #1067, opened 2023-12-07
chore: Rename the file CouchDbContainerTest.cs to CouchbaseContainerTest.cs

mgroves:
## What does this PR do?

Changes a filename

## Why is it important?

Reduces confusion by matching the filename to the class within it. The confusion between CouchDb and Couchbase has plagued both databases for years, so this is yet another small, but meaningful, step towards clarity.

## Related issues

This is a proposed fix for #1066

## How to test this PR

I don't think any testing is necessary, it's just a filename in a test project.


HofmeisterAn (approved):
😅 Thanks, this looks like a leftover from migrating to the new modules API.

PR #1073, opened 2023-12-12
feat: Add Firebird SQL module

park-jasper:
## What does this PR do?

Adding Testcontainers.FirebirdSql project that allows to use Firebird SQL with Testcontainers

## Why is it important?

Firebird SQL is not supported for Testcontainers .NET yet.
There is support for Testcontainers Java on the Firebird Repository (https://github.com/FirebirdSQL/firebird-testcontainers-java) which these changes are based on.

## Related issues

none

## How to test this PR

I created unit tests that test against the four tagged versions 2.5-ss, 2.5-sc, 3.0, 4.0
This is because the 2.5 versions behave differently and as such I wanted to ensure all the currently available versions will work.

## Comment

The code is based on the implementation of the MsSql project and the mentioned firebird-testcontainers-java repository. I am open to discuss if this is the right place or rather the firebird repository.


HofmeisterAn:
Thank you for the PR. I haven't had the time to review it yet, but I will do so in the coming days.

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
```diff
@@ -0,0 +1,170 @@
+namespace Testcontainers.FirebirdSql;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public class FirebirdSqlBuilder : ContainerBuilder<FirebirdSqlBuilder, FirebirdSqlContainer, FirebirdSqlConfiguration>
+{
+    public const string FirebirdSqlImage = "jacobalberty/firebird";
```
HofmeisterAn: Please pin the version.

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
```diff
@@ -0,0 +1,170 @@
...
+    {
+        var builder = Merge(DockerResourceConfiguration, new(username: username));
+        if (username.ToLower() != FirebirdSysdba)
+        {
+            builder = builder
+                .WithEnvironment("FIREBIRD_USER", username);
+        }
+
+        return builder;
+    }
```
HofmeisterAn: If a developer uses `sysdba`, the method will ignore it. Shouldn't we check this in the `Validation()` method? Or, we could use a similar approach as we are using in the [MySqlBuilder](https://github.com/testcontainers/testcontainers-dotnet/blob/e3be24f8256c21be4525e8abae623695c8bb7fb2/src/Testcontainers.MySql/MySqlBuilder.cs#L50-L71). I think I favor this. It makes the configuration simpler. WDYT?

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    private string GetDatabaseName()
+    {
+        if (State == TestcontainersStates.Running && IsFirebird2_5Image)
+        {
+            var dbPath = _configuration.Environments.TryGetValue("DBPATH", out var path)
+                ? path
+                : "/firebird/data";
+            return $"{dbPath}/{_configuration.Database}";
+        }
```
HofmeisterAn: Do we really need to support an outdated (EOL) version?

HofmeisterAn (changes requested):

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
park-jasper: I see your point, I think the way it is handled in the MySqlBuilder works great, I will apply that.

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
park-jasper: I appreciate you taking the time to look this over!
I actually have the use case where our customer is (stuck) using a firebird v2.5.9 database, and we want to test specifically against that EOL version.
My case is that
a) this is also part of the java testcontainers implementation that comes from the firebird repo, so the idea isn't too farfetched and,
b) if it becomes more hassle to support v2.5, support can be dropped in a future version, but devs may still use the older version to use whatever functionality it offers.

JMarianczuk: I am not sure about this section. In our project it _only_ works with v2.5 without settings the "/firebird/data" extra in front of the database, when using EntityFramework to access the database. I will do some more evaluation, but this is quite awkward.

JMarianczuk (changes requested):

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
```diff
@@ -0,0 +1,153 @@
...
+    public override FirebirdSqlContainer Build()
+    {
+        Validate();
+
+        var compoundWaitStrategy = Wait.ForUnixContainer()
+            .UntilContainerIsHealthy()
+            .AddCustomWaitStrategy(new WaitUntil(DockerResourceConfiguration));
+        return new FirebirdSqlContainer(
+            WithWaitStrategy(compoundWaitStrategy).DockerResourceConfiguration,
+            TestcontainersSettings.Logger);
```
HofmeisterAn: Are both wait strategies really necessary?

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
```diff
@@ -0,0 +1,153 @@
...
+
+    public const string DefaultDatabase = "test";
+
+    public const string DefaultUsername = "test";
+
+    public const string DefaultPassword = "test";
+
+    public const string DefaultSysdbaPassword = "masterkey";
+
+    private const string TestQueryString = "SELECT 1 FROM RDB$DATABASE;";
```
HofmeisterAn, line 19: We need to take a look at this statement. It fails for the version 2.5. While the exit code is `0` the error output contains:

    Statement failed, SQLSTATE = 08001
    I/O error during "open" operation for file "test"
    -Error while trying to open file
    -No such file or directory
    After line 0 in file /tmp/325ddf2a-2eff-4c45-af40-e80dd082d587/p3traelx.r1t
    Use CONNECT or CREATE DATABASE to specify a database
    Use CONNECT or CREATE DATABASE to specify a database

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+        string database;
+
+        if (_configuration.Image.Tag.StartsWith("2.5") || _configuration.Image.Tag.StartsWith("v2.5"))
+        {
+            database = string.Join("/", [_configuration.Environments.TryGetValue("DBPATH", out var dbPath) && !string.IsNullOrEmpty(dbPath) ? dbPath : "/firebird/data", _configuration.Database]);
+        }
+        else
+        {
+            database = _configuration.Database;
+        }
```
HofmeisterAn: I do not understand which part sets the env variable? The configuration never contains `DBPATH`. How does it work?

HofmeisterAn:
Thank you for addressing the comments. I have made some minor changes to ensure the module aligns with the other modules. I left three additional comments, and happy to merge the PR soon, once those comments are addressed. Thank you again.

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
park-jasper: Good point. The custom one alone is not sufficient, which is why I added the healthy check. It may be sufficient by itself

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
park-jasper: This is from the testcontainers-java repo. I guess it is not needed here.

park-jasper:
> Thank you for addressing the comments. I have made some minor changes to ensure the module aligns with the other modules. I left three additional comments, and happy to merge the PR soon, once those comments are addressed. Thank you again.

Thank you for your commits. I ran into an issue where the path to the database does not need to be modified with the /firebird/data path when using EntityFramework instead of the raw SqlConnection. I am still trying to figure out why that is. Until I can figure that out we should leave the PR open, it seems I prematurely opened it.
I may ask/involve someone from firebird on that topic and see what is up with the different ways the provider behaves.

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
HofmeisterAn: Ah, it is likely that this is something the user can override. The default value is probably `/firebird/data`. Additionally, I think that Java's implementation reads the value from the running container, while .NET does not. `_configuration` holds the information to start the container, not the information about the actual running container.

HofmeisterAn:
> > Thank you for addressing the comments. I have made some minor changes to ensure the module aligns with the other modules. I left three additional comments, and happy to merge the PR soon, once those comments are addressed. Thank you again.
> 
> Thank you for your commits. I ran into an issue where the path to the database does not need to be modified with the /firebird/data path when using EntityFramework instead of the raw SqlConnection. I am still trying to figure out why that is. Until I can figure that out we should leave the PR open, it seems I prematurely opened it. I may ask/involve someone from firebird on that topic and see what is up with the different ways the provider behaves.

If we figure out the correct wait strategy, I think we can merge the PR in the meantime (if you like). We can then update the configuration to work with EF afterward.

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
HofmeisterAn: Or does the configuration (connection string) depent on the actual client? Maybe the absolute path is only necessary for the Java client?

park-jasper:
> > > Thank you for addressing the comments. I have made some minor changes to ensure the module aligns with the other modules. I left three additional comments, and happy to merge the PR soon, once those comments are addressed. Thank you again.
> > 
> > 
> > Thank you for your commits. I ran into an issue where the path to the database does not need to be modified with the /firebird/data path when using EntityFramework instead of the raw SqlConnection. I am still trying to figure out why that is. Until I can figure that out we should leave the PR open, it seems I prematurely opened it. I may ask/involve someone from firebird on that topic and see what is up with the different ways the provider behaves.
> 
> If we figure out the correct wait strategy, I think we can merge the PR in the meantime (if you like). We can then update the configuration to work with EF afterward.

That sounds good, I will take a look at open comments and issues (next week) and we can incrementally improve this.

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
HofmeisterAn: I looked into this issue and indeed `/firebird/data/` is required to connect to the database for version 2.5 (isql CLI) too. Changing the database in the isql wait strategy command to `/firebird/data/test` works fine.

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
park-jasper: Yes, the Java version reads from the containers's ENV. DBPATH is also default set to /firebird/data. I think it is not necessary here, so the way it is left out now is good.

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
park-jasper: I saw the changes you did, looks good to me now the way it is.

src/Testcontainers.FirebirdSql/FirebirdSqlContainer.cs
HofmeisterAn: Did you see my latest changes? I moved it to the configuration class.

park-jasper: I did, and I think it is a good place for that logic to be.

HofmeisterAn: > Yes, the Java version reads from the containers's ENV. DBPATH is also default set to /firebird/data. I think it is not necessary here, so the way it is left out now is good.

Ok, then I think we can remove the part related to `DBPATH` since it does not read the variable from the container.

HofmeisterAn: Ah, sorry - it is already addressed and outdated.

src/Testcontainers.FirebirdSql/FirebirdSqlBuilder.cs
HofmeisterAn: I will do some tests to verify if the health check API alone is sufficient. I think the pull request looks good and is ready for merging. WDYT? Thank you for the pull request and your contribution ✌️.

park-jasper: I am content with the state it is right now. Thanks you for taking the time to look it over and putting in so much work yourself, too. For me it is fine to merge it.

HofmeisterAn (approved):

PR #1079, opened 2023-12-26
chore: Invert #if NETSTANDARD* conditional compilation conditions

0xced:
While conceptually .NET 5, 6, 7, 8 are greater than .NET Standard 2.1 it does not apply to the `NETSTANDARD2_1_OR_GREATER` macro.

So it's safer to conditionally compile on NETSTANDARD2_0 and assume that the #else branch targets a newer framework where the new bits are available in order not to use the old code path on the newest frameworks.

### DIFF
--- a/src/Testcontainers/Clients/TestcontainersClient.cs
+++ b/src/Testcontainers/Clients/TestcontainersClient.cs
@@ -271,10 +271,11 @@ namespace DotNet.Testcontainers.Clients
 
 #if NETSTANDARD2_0
         _ = await tarInputStream.ReadAsync(readBytes, 0, readBytes.Length, ct)
+          .ConfigureAwait(false);
 #else
         _ = await tarInputStream.ReadAsync(readBytes, ct)
-#endif
           .ConfigureAwait(false);
+#endif
 
         return readBytes;
       }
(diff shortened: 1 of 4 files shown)


HofmeisterAn (commented):
Hmm, I am curious, how did you run into this issue? If you consume the NuGet dependency, for example, in a `net8.0` project, it will use the `netstandard2.1` version. In the IDE, for instance in Rider, you can select the target framework (I think its default is `netstandard2.0`).

Thank you for bringing this to my attention. It is not something I would have expected, TBH, especially considering when I first used the macro years ago (which probably was not mentioned in the MSDN back then).

0xced:
> Hmm, I am curious, how did you run into this issue?

Sorry I should have been more clear in my description. This is _currently_ not an issue. This pull request is solely future-proofing the conditional compilation so that if a new target framework is added in the future (I have a plan 😉) then the _correct_ (newest) code will be compiled. This stuff is easy to get wrong, I already fixed it in https://github.com/spectreconsole/spectre.console/pull/563 too.

> If you consume the NuGet dependency, for example, in a `net8.0` project, it will use the `netstandard2.1` version.

Yes, absolutely! This is not an issue for consumers of the Testcontainers packages.

HofmeisterAn (approved):
Thanks ✌️

PR #1081, opened 2023-12-28
chore: Improve the Base64Provider resilience to malformed configuration files

0xced:
Also add some tests to cover different pathological scenarios.


tests/Testcontainers.Tests/Unit/Builders/Base64ProviderTest.cs
```diff
@@ -0,0 +1,65 @@
+using System.Text.Json;
+using DotNet.Testcontainers.Builders;
+using Microsoft.Extensions.Logging.Abstractions;
+using Xunit;
+
+namespace DotNet.Testcontainers.Tests.Unit
+{
+  public class Base64ProviderTest
```
HofmeisterAn: Can we  please move these tests to the already existing `Base64ProviderTest` test class? It is nested in `DockerRegistryAuthenticationProviderTest`. Probably we can reuse the existing test methods `ShouldGetAuthConfig` and `ShouldGetNull`.

src/Testcontainers/Builders/Base64Provider.cs
```diff
@@ -88,5 +96,18 @@ public IDockerRegistryAuthenticationConfiguration GetAuthConfig(string hostname)
...
+    private static byte[] DecodeBase64String(string base64)
+    {
+      try
+      {
+        return Convert.FromBase64String(base64);
+      }
+      catch (FormatException)
+      {
+        return null;
+      }
```
HofmeisterAn: Should we write a log message in case of an exception? It is likely a misconfiguration?

0xced: Good point! I added logs for every code path where the `auth` part can't be decoded in f79827f2ba376b4e29be8327d08f89961252442f.

tests/Testcontainers.Tests/Unit/Builders/Base64ProviderTest.cs
0xced: Absolutely, I got confused by the fact the existing `Base64ProviderTest` does not follow the same directory structure as in `src`. I have moved all the tests into the existing class in acd009ecea51ea89c9cbd232006a6bd426cd730c.

HofmeisterAn (approved):
Thanks 🎉.

PR #1082, opened 2024-01-02
feat: Add docker compose support (#122)

### LINKED ISSUE #122: Add missing function `WithDockerCompose`
I love the approach of having a single container scenario for a test case but there are also scenarios where I'd like to spin up a whole set of containers and therefore use docker-compose as declaration. 

I'd like to spin up a set of containers by specifying a `docker-compose.yml` similar to the Java version  https://www.testcontainers.org/modules/docker_compose/ 

**Describe alternatives you've considered**
Not sure If there is one
- configuring the things as in the compose in code seems odd to me and leads to duplication of configuration and most of the times `docker-compose.yml` will also be used to deploy the application under test this would be strange. 

**Additional context**
Also other library authors could benefit from such an approach, I happen to know that https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks is using compose and I think it would be great improvement on there development experience (since you don' t have to remember starting the containers in advance ;) ) 

Cheers & let me know what you think of the idea

nkz-soft:
## What does this PR do?

Adds docker compose support

## Why is it important?

We have docker compose support for java implementations, but that's not here yet. This PR does not contain a complete implementation of all methods in the java library, but it is a starting point from which you can begin to port the rest of the methods.

## Related issues

- Closes #122

## Additional information

At this time, there are only two methods in the `DockerComposeBuilder` class:

- `WithLocalCompose` - Specifies the `docker-compose` file that needs to be executed.
- `WithComposeFile` - Whether or not to execute `docker-composer` locally or in a container.

Additional methods are available for implementing in Java, listed in order of importance (to me):

- `withOptions` - Adds options to the docker-compose command, e.g. docker-compose --compatibility.
- `withRemoveImages` - Remove images after containers shutdown.
- `withRemoveVolumes` - Remove volumes after containers shut down.
- `withBuild` - Whether to always build images before starting containers.
- `withExposedService` - Specifies the port for current service. According to the documentation, Testcontainers will spin up a small 'ambassador' container, which will proxy between the Compose-managed containers and ports that are accessible to your tests. This is done using a separate, minimal container that runs socat as a TCP proxy. I don't fully understand the necessity of creating a proxy (ambassador) container here, perhaps to avoid port conflicts or to allow the port to listen.
- `getServiceHost` - Returns the IP address where the container is listening (via an ambassador container).
- `getServicePort(serviceName, servicePort)` - Returns the Docker mapped port for a port that has been exposed (via an ambassador container).

It's my opinion that we should start with the first four methods that cover 90% of the functionality.
The implementation of proxy and port forwarding functions in docker-compose can be done directly through a yaml file.


HofmeisterAn:
Thanks for the initial PR. Could you please provide additional information in the PR description, specifying which parts are addressed by the PR and which parts are not covered, in comparison to the features provided by Java? Additionally, what aspects do we need to take care of afterward? What follow-ups are necessary, etc.?

nkz-soft:
> Thanks for the initial PR. Could you please provide additional information in the PR description, specifying which parts are addressed by the PR and which parts are not covered, in comparison to the features provided by Java? Additionally, what aspects do we need to take care of afterward? What follow-ups are necessary, etc.?

I have added some additional information, but I am not sure that this is enough. Please take a look at this.

HofmeisterAn:
> I have added some additional information, but I am not sure that this is enough. Please take a look at this.

Thanks. I will need some days to review it. A lot of other topics have accumulated over the holidays.

dmytro-dovhan-gl:
Hi, Any updates here?

HofmeisterAn:
> Hi, Any updates here?

Sorry, I haven't had the time yet, but it is the next item in my GH inbox. I cannot promise it, but I will try to look at it sometime next week.

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+            !dockerComposeContainer.RuntimeConfiguration.LocalCompose )
+        {
+            var fileInfo = new FileInfo(dockerComposeContainer.RuntimeConfiguration.ComposeFile);
+            if (!fileInfo.Exists)
+            {
+                throw new FileNotFoundException(NoComposeFile, fileInfo.Name);
+            }
+
+            await container.CopyAsync(fileInfo, ".", Unix.FileMode644, ct)
+                .ConfigureAwait(false);
```
HofmeisterAn: Instead of copying the file here, could we please move it to `WithComposeFile(string)` and utilize the `WithResourceMapping(FileInfo, FileInfo)` builder method? For example, similar to [this implementation](https://github.com/testcontainers/testcontainers-dotnet/blob/7f41b7a7f5dbf371e54dee54df62cd2644b90818/src/Testcontainers.WebDriver/WebDriverBuilder.cs#L65-L68). You can keep the reference to the file in addition.

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+    private async Task ConfigureDockerComposeAsync(IContainer container, CancellationToken ct = default)
+    {
+        if (container is DockerComposeRemote dockerComposeContainer && 
+            !dockerComposeContainer.RuntimeConfiguration.LocalCompose )
+        {
+            var fileInfo = new FileInfo(dockerComposeContainer.RuntimeConfiguration.ComposeFile);
+            if (!fileInfo.Exists)
+            {
+                throw new FileNotFoundException(NoComposeFile, fileInfo.Name);
+            }
```
HofmeisterAn: Can we please move this check to `Validate()`.

```csharp
_ = Guard.Argument(DockerResourceConfiguration.ComposeFile, nameof(DockerResourceConfiguration.ComposeFile))
    .ThrowIf(argument => !File.Exists(argument.Value), argument => new FileNotFoundException("Compose file not found.", argument.Name));
```

Probably we will no longer need `ConfigureDockerComposeAsync(IContainer, CancellationToken)`.

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    
+    /// <inheritdoc />
+    protected override DockerComposeBuilder Init()
+    {
+        return base.Init()
+            .WithImage(DockerComposeImage)
+            .WithEntrypoint(CommonCommands.SleepInfinity)
+            .WithBindMount(DockerSocketPath, DockerSocketPath, AccessMode.ReadWrite)
```
HofmeisterAn:
```suggestion
            .WithBindMount(DockerSocketPath, DockerSocketPath, AccessMode.ReadOnly)
```

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (!fileInfo.Exists)
+            {
+                throw new FileNotFoundException(NoComposeFile, fileInfo.Name);
+            }
+
+            await container.CopyAsync(fileInfo, ".", Unix.FileMode644, ct)
+                .ConfigureAwait(false);
+
+            await container.ExecAsync(dockerComposeContainer.BuildStartCommandLine(),  ct)
+                .ConfigureAwait(false);
```
HofmeisterAn: What do you think about configuring the entrypoint and command as other modules do in `Init()`, like:

```csharp
.WithEntrypoint("docker")
.WithCommand("compose", "up")
```

This will keep the command, including its arguments, in the container configuration. Starting the remote container will automatically pick it up. The local proxy container can also use these properties and append file `--file` arg.

Then, `WithComposeFile(string)` can simply copy the Compose file using the default name to the container:

```csharp
return Merge(DockerResourceConfiguration, new DockerComposeConfiguration(composeFile: composeFile))
    .WithResourceMapping(new FileInfo(composeFile), new FileInfo("docker-compose.yml"));
```

Which makes `.WithEntrypoint(CommonCommands.SleepInfinity)` obsolete too (running not detached).

src/Testcontainers.DockerCompose/DockerCompose.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Initializes a new instance of the <see cref="DockerCompose" /> class.
+    /// </summary>
+    protected DockerCompose(IContainerConfiguration configuration, ILogger logger) : base(configuration, logger)
+    {
+    }
+
+    /// <summary>
+    /// Gets the runtime configuration.
+    /// </summary>
+    public DockerComposeConfiguration RuntimeConfiguration => _configuration as DockerComposeConfiguration;
```
HofmeisterAn: Do we need that property publicly available? We can keep a reference for the derived classes if necessary, then we do not need to cast the field all the time and can keep it private.

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+    
+    /// <summary>
+    /// Remove images after containers shutdown.
+    /// </summary>
+    /// <param name="removeImages"></param>
+    /// <returns>A configured instance of <see cref="DockerComposeBuilder" />.</returns>
+    public DockerComposeBuilder WithRemoveImages(RemoveImages removeImages) {
+        return Merge(DockerResourceConfiguration, new DockerComposeConfiguration
+            (removeImages: removeImages));
+    }
```
HofmeisterAn: Is this a Java feature?

HofmeisterAn (changes requested):
Thank you for the PR. Sorry for the late response, I hadn't found the time earlier. I still need to run and test it locally. Meanwhile, I have a couple of suggestions and improvements. It would be great if we could discuss and address them in the meantime. Thanks again for your contribution.

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
nkz-soft: Yep, you can look it up [here](https://github.com/testcontainers/testcontainers-java/blob/c210afb0b0944d1a96b7a760f93d6601a9058173/core/src/main/java/org/testcontainers/containers/ComposeContainer.java#L329).

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
nkz-soft: Yes, it's a prettier implementation. I'll give it a try.

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
nkz-soft: The main idea here was to keep the container running until the docker finishes, without the -d option this should work.

src/Testcontainers.DockerCompose/DockerCompose.cs
nkz-soft: Do you mean something like this?
```
    /// <summary>   
    /// Initializes a new instance of the <see cref="DockerCompose" /> class.
    /// </summary>
    protected DockerCompose(IContainerConfiguration configuration, ILogger logger) : base(configuration, logger)
    {
        RuntimeConfiguration = _configuration as DockerComposeConfiguration;
    }

    /// <summary>
    /// Gets the runtime configuration.
    /// </summary>
    protected DockerComposeConfiguration RuntimeConfiguration { get; }
```

nkz-soft:
> Thank you for the PR. Sorry for the late response, I hadn't found the time earlier. I still need to run and test it locally. Meanwhile, I have a couple of suggestions and improvements. It would be great if we could discuss and address them in the meantime. Thanks again for your contribution.

Thanks, I think next week, I'll check and fix all the issues.

src/Testcontainers.DockerCompose/DockerComposeBuilder.cs
nkz-soft: `.WithCommand("compose", "up")`
One more thing, the command line may change after initialization, additional arguments may be added. Where is the best place to add them? Through a method in `DockerComposeBuilder`, something like that? 
```
    public DockerComposeBuilder WithOptions(params string[] options)
    {
        return Merge(DockerResourceConfiguration, new DockerComposeConfiguration(options: options))
            .WithCommand(options);
    }
```

nkz-soft:
> Thank you for the PR. Sorry for the late response, I hadn't found the time earlier. I still need to run and test it locally. Meanwhile, I have a couple of suggestions and improvements. It would be great if we could discuss and address them in the meantime. Thanks again for your contribution.

I think it's done, could you please review the code again?

HofmeisterAn:
Please see my suggestions and comments here: https://github.com/nkz-soft/testcontainers-dotnet/pull/1. Overall, it looks good, but there are still a few things we need to discuss and sort out first.

dava2788:
Hello all.
Any updates of this PR.
Really interesting in using this docker compose functionality

Best Regards

kanpov:
It'd also be nice to land support for podman's k8s yaml files (`podman kube play`) as well once podman support is actually finished.

InspiringCode:
@nkz-soft Any updates on this? This would be really useful for our projects too!

nkz-soft:
> @nkz-soft Any updates on this? This would be really useful for our projects too!

@HofmeisterAn
Unfortunately, I don't have a clue. I'm still waiting for a code review.

JBastiaan:
Id also really like to see progress on this, would make setting up testcontainers so much easier in certain scenarios where you already have a docker-compose. So upvote from me :)

HofmeisterAn:
We need to merge the develop branch and address the build issues first. Additionally, I still believe this does not contain the necessary features developers expect for an MVP. We can split the Docker Compose support into multiple PRs, but we would need something like a feature flag for the current state (opt-in) incl. docs about the current capabilities. The implementation does not or cannot indicate the readiness of the containers (i.e., when the services are actually running, known as wait strategies), nor is it able to retrieve the Docker mapped ports for exposed ports. Which IMO are very basic and necessary features to write reliable tests.

deMD:
@nkz-soft @HofmeisterAn 

Any updates on this PR? I just noticed a [LinkedIn post about this feature](https://www.linkedin.com/posts/gferreira_add-missing-function-withdockercompose-activity-7272193124536381441-hkVP?utm_source=share&utm_medium=member_desktop) and it would be super useful, but this PR seems stale atm?

tuga001-sme:
@HofmeisterAn any news?

HofmeisterAn:
My previous comment still applies, the PR isn't up to date.

PureKrome:
@nkz-soft any chance you can come back to this PR and update it?

PR #1093, opened 2024-01-22
Use the `pg_isready` command to asses whether PostgreSQL is ready or not

### LINKED ISSUE #1092: [Bug]: The PostgreSqlContainer wait strategy is not reliable (non default configuration)
### Testcontainers version

3.7.0

Yes

### Host OS

Any

### Host arch

Any

### .NET version

8.0.100

### Docker version

```shell
Docker version 20.10.21, build baeda1f
```

### Docker info

```shell
Client:
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.9.1)
  compose: Docker Compose (Docker Inc., v2.13.0)
  dev: Docker Dev Environments (Docker Inc., v0.0.5)
  extension: Manages Docker extensions (Docker Inc., v0.2.16)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.22.0)

Server:
 Containers: 5
  Running: 0
  Paused: 0
  Stopped: 5
 Images: 78
 Server Version: 20.10.21
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 770bd0108c32f3fb5c73ae1264f7e503fe7b2661
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.15.49-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 7.772GiB
 Name: docker-desktop
 ID: NURJ:VHOB:QPS5:BPZK:FS36:CRYL:2G6T:42JR:44K4:6OVH:PXOO:WPRL
 Debug Mode: true
  File Descriptors: 70
  Goroutines: 96
  System Time: 2024-01-22T20:42:41.925461134Z
  EventsListeners: 27
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 No Proxy: hubproxy.docker.internal
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

### What happened?

As already briefly mentioned in https://github.com/testcontainers/testcontainers-dotnet/pull/920#issuecomment-1666129246, the default wait strategy for `PostgreSqlContainer` might hang forever.

There are several ways to trigger this faulty behaviour.
1. Reusing a volume with `WithVolumeMount("Testcontainers.PostgreSql.Data", "/var/lib/postgresql/data")`
2. Reusing the container with `WithReuse(true)`
3. Starting / stopping the container multiple times in a row:
```csharp
var postgreSqlContainer = new PostgreSqlBuilder().Build();
for (var i = 0; i < 3; i++)
{
    await postgreSqlContainer.StartAsync();
    await postgreSqlContainer.StopAsync();
}
```

The logs produced when reusing an existing volume looks like this:

```text
PostgreSQL Database directory appears to contain a database; Skipping initialization
2024-01-22 22:15:39.834 UTC [1] LOG:  starting PostgreSQL 15.1 (Debian 15.1-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
2024-01-22 22:15:39.834 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2024-01-22 22:15:39.834 UTC [1] LOG:  listening on IPv6 address "::", port 5432
2024-01-22 22:15:39.834 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2024-01-22 22:15:39.837 UTC [29] LOG:  database system was shut down at 2024-01-22 22:15:26 UTC
2024-01-22 22:15:39.845 UTC [1] LOG:  database system is ready to accept connections
```

We can see that `database system is ready to accept connections` appears only once instead of twice which is [expected by the wait strategy](https://github.com/testcontainers/testcontainers-dotnet/blob/3.7.0/src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs#L134-L137).

When reusing the container or starting/stopping it many times in a row, the logs will grow and `database system is ready to accept connections` will appear more than twice.

In both cases the `UntilAsync` method never returns `true` and starting the container hangs.

### Relevant log output

_No response_

### Additional information

I have already prepared a fix for this issue that I will submit shortly.

0xced:
## What does this PR do?

This pull request uses the builtin `pg_isready` command instead of searching the logs for `database system is ready to accept connections` in order to determine whether the PostgreSQL database is ready or not.

## Why is it important?

Under some circumstances (described in #1092), starting a `PostgreSqlContainer` might hang forever.

## Related issues

Closes #1092

## How to test this PR

A new test (`Testcontainers.PostgreSql.PostgreSqlContainerTest.StopAndStartMultipleTimes`) was added to demonstrates how the hang might happen and how using `pg_isready` instead addresses this issue.


src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs
```diff
@@ -123,18 +123,10 @@ protected override PostgreSqlBuilder Merge(PostgreSqlConfiguration oldValue, Pos
...
-            var (stdout, stderr) = await container.GetLogsAsync(timestampsEnabled: false)
-                .ConfigureAwait(false);
-
-            return 2.Equals(Array.Empty<string>()
-                .Concat(stdout.Split(LineEndings, StringSplitOptions.RemoveEmptyEntries))
-                .Concat(stderr.Split(LineEndings, StringSplitOptions.RemoveEmptyEntries))
-                .Count(line => line.Contains("database system is ready to accept connections")));
+            return ((PostgreSqlContainer)container).IsReadyAsync();
         }
     }
```
HofmeisterAn, line 156: Can we please follow the pattern we are already using in other moduls (e.g. MySQL [1](https://github.com/testcontainers/testcontainers-dotnet/blob/f4ee0b2d4cd0d02d11d1e1ecceb90c6d521728e7/src/Testcontainers.MySql/MySqlBuilder.cs#L78-L81), [2](https://github.com/testcontainers/testcontainers-dotnet/blob/f4ee0b2d4cd0d02d11d1e1ecceb90c6d521728e7/src/Testcontainers.MySql/MySqlBuilder.cs#L128-L148))? I prefer a consistent approach; it is easier to navigate and refactor the code.

tests/Testcontainers.PostgreSql.Tests/PostgreSqlContainerTest.cs
```diff
@@ -42,4 +42,33 @@ public async Task ExecScriptReturnsSuccessful()
...
+    {
+        // Given
+        var timeoutSource = new CancellationTokenSource(TimeSpan.FromSeconds(60));
+
+        // When
+        var exception = await RestartAsync(timeoutSource.Token);
+
+        // Then
+        Assert.Null(exception);
+    }
```
HofmeisterAn, line 58: While this will work for PostgreSQL, it won't work for other modules, specifically for those who use log messages to indicate readiness. I was thinking about a more general approach to utilize the existing module tests. What do you think about the following proposal? We can then update the module's tests over time.

```csharp
namespace DotNet.Testcontainers.Commons;

public class SharedContainerInstance<TContainer> : IAsyncLifetime
    where TContainer : IContainer
{
    public SharedContainerInstance(TContainer container)
    {
        Container = container;
    }

    public TContainer Container { get; }

    public Task InitializeAsync()
    {
        return Task.CompletedTask;
    }

    public Task DisposeAsync()
    {
        return Container.DisposeAsync().AsTask();
    }
}
```

```diff
diff --git a/tests/Testcontainers.PostgreSql.Tests/PostgreSqlContainerTest.cs b/tests/Testcontainers.PostgreSql.Tests/PostgreSqlContainerTest.cs
index 5433e8c..a966ca9 100644
--- a/tests/Testcontainers.PostgreSql.Tests/PostgreSqlContainerTest.cs
+++ b/tests/Testcontainers.PostgreSql.Tests/PostgreSqlContainerTest.cs
@@ -1,17 +1,22 @@
 namespace Testcontainers.PostgreSql;
 
-public sealed class PostgreSqlContainerTest : IAsyncLifetime
+public sealed class PostgreSqlContainerTest : IClassFixture<PostgreSqlContainerTest.SharedContainerInstance>, IAsyncLifetime
 {
-    private readonly PostgreSqlContainer _postgreSqlContainer = new PostgreSqlBuilder().Build();
+    private readonly PostgreSqlContainerTest.SharedContainerInstance _sharedContainerInstance;
+
+    public PostgreSqlContainerTest(PostgreSqlContainerTest.SharedContainerInstance sharedContainerInstance)
+    {
+        _sharedContainerInstance = sharedContainerInstance;
+    }
 
     public Task InitializeAsync()
     {
-        return _postgreSqlContainer.StartAsync();
+        return _sharedContainerInstance.Container.StartAsync();
     }
 
     public Task DisposeAsync()
     {
-        return _postgreSqlContainer.DisposeAsync().AsTask();
+        return _sharedContainerInstance.Container.StopAsync();
     }
 
     [Fact]
@@ -19,7 +24,7 @@ public sealed class PostgreSqlContainerTest : IAsyncLifetime
     public void ConnectionStateReturnsOpen()
     {
         // Given
-        using DbConnection connection = new NpgsqlConnection(_postgreSqlContainer.GetConnectionString());
+        using DbConnection connection = new NpgsqlConnection(_sharedContainerInstance.Container.GetConnectionString());
 
         // When
         connection.Open();
@@ -36,10 +41,19 @@ public sealed class PostgreSqlContainerTest : IAsyncLifetime
         const string scriptContent = "SELECT 1;";
 
         // When
-        var execResult = await _postgreSqlContainer.ExecScriptAsync(scriptContent)
+        var execResult = await _sharedContainerInstance.Container.ExecScriptAsync(scriptContent)
             .ConfigureAwait(true);
 
         // When
         Assert.True(0L.Equals(execResult.ExitCode), execResult.Stderr);
     }
+
+    [UsedImplicitly]
+    public sealed class SharedContainerInstance : SharedContainerInstance<PostgreSqlContainer>
+    {
+        public SharedContainerInstance()
+            : base(new PostgreSqlBuilder().Build())
+        {
+        }
+    }
 }
\ No newline at end of file
```

HofmeisterAn (changes requested):

src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs
0xced: Absolutely, I just addressed this in bf08a0ec9865fd1bdf23d34c5d20bca85cca9e19.

0xced: I added `.ConfigureAwait(false)` and force-pushed so that's now commit cf0fdf109940b6bd6df0c4467849da4284b2409c.

tests/Testcontainers.PostgreSql.Tests/PostgreSqlContainerTest.cs
0xced: > While this will work for PostgreSQL, it won't work for other modules, specifically for those who use log messages to indicate readiness. 

I think we are not on the same page on this topic. The purpose of the `StopAndStartMultipleTimes` that was added for the PostgreSQL module was in my mind a way to prove that the log-based wait strategy was incorrect, in test driven development style.

1. Run the `StopAndStartMultipleTimes` test
2. ❌ Test does not pass (times out after one minute)
3. Fix the wait strategy by using `pg_isready` instead of reading logs
4. Run the `StopAndStartMultipleTimes` test again
5. ✅ The test now passes indicating that the `pg_isready` strategy implementation is correct

If I understand your intention correctly, you'd like to introduce a way to start/stop containers before/after each tests. I think that could be a good addition but should probably be handled in a separate pull request.

HofmeisterAn: Yes, I understand your intention. I think we are talking about the same thing; my approach ensures that the service running inside the container is ready too.

If the container indicates readiness after a stop, that does not necessarily mean that the service running inside the container is ready. Consider log message wait strategies, for example. If we simply check for "log message contains x", this may indicate readiness too early, because the message already exists from a previous start.

My intention is to ensure that not only the wait strategy passes but also that the service running inside the container is ready from the user's perspective, utilizing our existing tests.

If we stop and start the container in between, and the second test fails or hangs, we now know the wait strategy is broken. I agree; your suggestion is more explicit but lacks real readiness.

0xced: OK, we are indeed on the same page! And going a bit further to ensure that the service is actually running is a good idea. 👍

I see you already started working on it in the `feature/set-created-started-stopped-container-timestamp-pass-wait-strategy` branch. I was hoping to be able to use `ContainerFixture` from my [Testcontainers.Xunit branch](https://github.com/0xced/testcontainers-dotnet/tree/Testcontainers.Xunit) which serves the exact same purpose as `SharedContainerInstance`. But that branches depends on #1100.

I'm still convinced that a dedicated `Testcontainers.Xunit` NuGet package that could be used both from inside Testcontainers test projects and for consumer integration tests is a great idea. 😉

HofmeisterAn: > I was hoping to be able to use `ContainerFixture` from my [Testcontainers.Xunit branch](https://github.com/0xced/testcontainers-dotnet/tree/Testcontainers.Xunit) which serves the exact same purpose as `SharedContainerInstance`. But that branches depends on #1100.

The PR does not contain the Xunit project, right? I think it is a good idea! A lot of developers will benefit from it. I have not taken a close look into the PR that involves the changes to the logging implementation yet. I took a quick look, and it is quite big. IIRC, I proposed to split it into two steps: one that provides an internal `WithLogger(ILogger)` builder method, and another that makes the necessary changes regarding logging the container runtime information and making the method public. After that, we can add the dedicated Xunit, package. WDYT? Splitting it makes reviewing much easier.

> I'm still convinced that a dedicated `Testcontainers.Xunit` NuGet package that could be used both from inside Testcontainers test projects and for consumer integration tests is a great idea. 😉

✅

BTW, did you notice the changes I made in the mentioned branch above to the PostgreSQL's wait strategy? Through the issue and the referred GitHub discussion, I noticed that PostgreSQL logs the interfaces to which it is listening. Maybe we do not need to invoke `pg_isready`.

0xced: I'll reply in the comments of #1100 for all the `ILogger` related stuff.

> BTW, did you notice the changes I made in the mentioned branch above to the PostgreSQL's wait strategy? Through the issue and the referred GitHub discussion, I noticed that PostgreSQL logs the interfaces to which it is listening. Maybe we do not need to invoke `pg_isready`.

I've seen you made some changes about the strategy. I still think that leveraging `pg_isready` is the _right_ thing to do. I never liked making decisions based on reading logs of another program. Those are just logs and — although unlikely — could change in the future. The whole purpose of the `pg_isready` command line tool is to test whether a PostgreSQL database is ready and that is pretty much guaranteed not to break in future versions. From an engineering standpoint it feels much better to me. Also, it's very close to the MySQL / MariaDb strategies.

HofmeisterAn: I agree. Logs are always a bit tricky, especially across different versions. As long as `pg_isready` is reliable, I am fine. Did you check other PostgreSQL images (versions) as well (I guess they contain `pg_isready` too)? Let's merge `develop` and complete this pull request (after #1110 if you do not mind).

0xced: Good point about older versions. The `pg_isready` command was introduced in PostgreSQL 9.3 so I will revisit this PR once #1110 is merged. I will test with PostgreSQL 8 (which is the oldest image available on Docker Hub) and see what happens.

HofmeisterAn: I guess if versions 14, 15, and 16 support it, it is good enough. Developers can still override the wait strategy or use the generic builder if an older version is really necessary.

0xced:
Superseded by #1111.

PR #1099, opened 2024-01-25
feat: Add Ollama module

### LINKED ISSUE #1097: [Enhancement]: Ollama module
### Problem

Its not having an Ollama module.

I realize that my POC here needs to have a discussion of design and features, but its a fun litte thing

### Solution

Make a module, here's a prototype of Ollama working with Testcontainers in LinqPad 8:

Deps the nugets Testcontainers and OllamaSharp

```c#

var container = new ContainerBuilder()
	.WithName("ollama-container")
	.WithImage("ollama/ollama")
	.WithPortBinding(11434, 11434)
	.WithVolumeMount("ollama", "/root/.ollama")
	.WithExposedPort(11434)
	.Build();
	
var keepAlive = true;

container.Started += async (sender, args) => 
{
	Console.WriteLine("Container started");
	try
	{
		Console.WriteLine("Start ollama service");
		await container.ExecAsync(new List<string>() {
			"ollama run llama2"
		});
		var result = await RunAsync();
		result.Dump();
	}
	catch (Exception ex)
	{
		Console.WriteLine(ex);
	}
	
	keepAlive = false;
};

await container.StartAsync();

while (!QueryCancelToken.IsCancellationRequested && keepAlive)
{
	await Task.Delay(1000);
}

var logs = await container.GetLogsAsync();

await container.StopAsync();
await container.DisposeAsync();

async Task<string> RunAsync()
{
	try
	{
		var client = new OllamaApiClient("http://localhost:11434/api/generate", "llama2");
		ChatRequest request = new ChatRequest();
		request.Messages ??= new List<Message>();
		request.Messages.Add(new Message(ChatRole.User, "Write me a dad-joke about cats please?", Enumerable.Empty<string>().ToArray()));
		request.Model = "llama2";
		request.Stream = false;
		var messages = new List<ChatResponseStream>();
		var responses = await client.SendChat(request, x => messages.Add(x));
		return string.Join("\n\n", responses.Select(r => r.Content));
	}
	catch (Exception ex)
	{
		return ex.ToString();
	}
}
```

### Benefit

Its really cool

### Alternatives

To be less cool

Yes

frankhaugen:
## What does this PR do?

Added a new Module project that uses OLLAMA docker images to give local LLM capabilities. This only supports CPU workload, but adding the env. variables to activate GPU if the prerequisites are there, can be done and so CUDA core based processing isn't precluded

## Why is it important?

This module is very powerful and can give a lot of value to those that are incorporating OLLAMA models in their projects

## Related issues

Link related issues below. Insert the issue link or reference after the word "Closes" if merging this should automatically close it.

- Closes #1097

## How to test this PR

Run the test added

## Follow-ups

- Make more docs to guide the user to be able to use GPU might be a future thing, but right now, no need.


src/Testcontainers.Ollama/OllamaContainer.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (State!= TestcontainersStates.Created && State != TestcontainersStates.Running)                {
+                throw new InvalidOperationException("Cannot start a container that has not been created.");
+            }
+            Task.WaitAll(ExecAsync(new List<string>()
+            {
+                "ollama", "run", ModelName,
+            }));
+            
+            await Task.CompletedTask;
+        }
```
HofmeisterAn: Does it make sense to pass the model name as an argument to the member? `Run(string modelName, CancellationToken ct = default)`? In addition, please do not block the thread; simply return `ExecAsync(["ollama", "run", modelName], ct);`.

src/Testcontainers.Ollama/OllamaConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+    
+        /// <summary>
+        /// Default volume path.
+        /// </summary>
+        public const string DefaultVolumePath = "/root/.ollama";
+    
+        /// <summary>
+        /// Default volume name.
+        /// </summary>
+        public const string DefaultVolumeName = "ollama-volume";
```
HofmeisterAn: Can you please move the default configurations to the builder, like all the other modules do?

src/Testcontainers.Ollama/OllamaContainer.cs
```diff
@@ -0,0 +1,52 @@
...
+    
+        /// <summary>
+        /// Starts the Ollama container.
+        /// </summary>
+        public async Task StartOllamaAsync()
+        {
+            if (State!= TestcontainersStates.Created && State != TestcontainersStates.Running)                {
+                throw new InvalidOperationException("Cannot start a container that has not been created.");
+            }
```
HofmeisterAn: We use `ThrowIfResourceNotFound();` to make sure the resources exists and is "running".

src/Testcontainers.Ollama/OllamaConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+    
+        /// <summary>
+        /// Gets the default port of the Ollama API.
+        /// </summary>
+        public const int DefaultPort = 11434;
+        
+        /// <summary>
+        /// Default image name.
+        /// </summary>
+        public const string ImageName = "ollama/ollama:latest";
```
HofmeisterAn: Please pin the image version.

tests/Testcontainers.Ollama.Tests/OllamaContainerTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    private OllamaContainer _ollamaContainer;
+
+    public OllamaContainerTests(ITestOutputHelper testOutputHelper)
+    {
+      _testOutputHelper = testOutputHelper;
+    }
+
+    public async Task InitializeAsync()
+    {
+      TestcontainersSettings.Logger = new TestOutputLogger(nameof(OllamaContainerTests), _testOutputHelper);
```
HofmeisterAn: Can we please keep the default logger configuration, etc. I am aware of the issues around xUnit.net's `ITestOutputHelper` and console logging, but I would prefer to keep the modules and tests consistent. This helps in applying future refactoring, etc.

HofmeisterAn (changes requested):
Thanks for the PR. I had a quick look and added some minor suggestions and improvements.

src/Testcontainers.Ollama/OllamaContainer.cs
frankhaugen: Agreed and resolved about the return. I added two methods one taking only the ct and one taking name and ct. Have a look if you agree

src/Testcontainers.Ollama/OllamaConfiguration.cs
frankhaugen: Done

src/Testcontainers.Ollama/OllamaConfiguration.cs
frankhaugen: Oops, that was an oversight

tests/Testcontainers.Ollama.Tests/OllamaContainerTests.cs
frankhaugen: I was a bit careless, and shouldn't have included it, its very much a "I like this to as a debug tool while working" so, removed

src/Testcontainers.Ollama/OllamaContainer.cs
frankhaugen: Done

frankhaugen:
@HofmeisterAn  I'm updating the branch with my current state as the tests were green on local, but I think it needs another pass. No Rush, I'm busy with my day-job so, no need for you to rush 😄

src/Testcontainers.Ollama/OllamaBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+    
+        /// <summary>
+        /// Default volume name.
+        /// </summary>
+        public const string DefaultVolumeName = "ollama-volume";
+
+        /// <summary>
+        /// The default model name for the OllamaBuilder.
+        /// </summary>
+        public const string DefaultModelName = OllamaModels.Llama2;
```
eddumelendez: Models are just like images and should be provided as a string rather than having an specific type. It will also reduce the amount of PRs just to keep it up-to-date

src/Testcontainers.Ollama/OllamaBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+        
+        /// <summary>
+        /// Default image name and version tag.
+        /// </summary>
+        public const string OllamaImage = "ollama/ollama:0.1.22";
+    
+        /// <summary>
+        /// Default volume path.
+        /// </summary>
+        public const string DefaultVolumePath = "/root/.ollama";
```
eddumelendez: Volume mounting should be optional rather than the default, at least for now. We can align along with Java and Go implementation.

eddumelendez (commented):
Thanks for your contribution. Testcontainers for Java and Go has recently release a new Ollama module, which allows to detect and enable GPU and allow to commit the container's state and create a new Image. It would be nice to align the .NET implementation along with the others. See [java implementation](https://github.com/testcontainers/testcontainers-java/pull/8369).

duranserkan:
Is there any update on this?

IRusio:
any update on that?

HofmeisterAn (commented):
I updated the PR to develop. I'll take a look at it sometime next week and address the necessary parts.

HofmeisterAn (approved):
I updated the module implementation and aligned it with the other Testcontainers module implementations. I also added a text example.

> Thanks for your contribution. Testcontainers for Java and Go has recently release a new Ollama module, which allows to detect and enable GPU and allow to commit the container's state and create a new Image. It would be nice to align the .NET implementation along with the others. See java implementation.

I didn't include detecting and setting the GPU configuration, as it's not that simple in .NET (but it's something we can address in the future). I'm also not sure if committing an image is supported in Docker.DotNet. I need to double-check that. It might be something we need to implement first. If someone wants to pick it up, that would be great, but we need to discuss how to address this in an issue first.

```csharp
// How to set the device requests using the container builder API:
.WithCreateParameterModifier(parameterModifier =>
    parameterModifier.HostConfig.DeviceRequests
        = new List<DeviceRequest> { new DeviceRequest { /* TODO: Set properties. */ } })
```

kiview:
I also like to reference the `useLocal` feature in the [tc-go implementation](https://golang.testcontainers.org/modules/ollama/#use-local). Pragmatically, it is the best way to get a good DX on macOS.

PR #1103, opened 2024-01-31
feat: Add Apache Pulsar module

entvex:
## What does this PR do?

This PR introduces Apache Pulsar support to testcontainers-dotnet, enabling users to easily spin up Pulsar containers for testing purposes. It also provides an option to enable TokenAuthentication, allowing for secure Pulsar interactions.

## Why is it important?

Apache Pulsar is a rapidly growing distributed messaging system gaining traction in the cloud-native ecosystem. Integrating Pulsar support into testcontainers-dotnet will make it easier for developers to test their applications using Pulsar. Additionally, providing TokenAuthentication support enhances security and privacy in Pulsar-based testing scenarios.

## How to test this PR

Run the XUnit tests in the testcontainers-dotnet project to ensure the Pulsar containerization and TokenAuthentication functionality work as expected.

## Follow-ups

@HofmeisterAn, I encountered a challenge while adding [WithTokenAuthentication ](https://github.com/entvex/testcontainers-dotnet/blob/db9156d6707a00bace05994b637a0a7647085e58/src/Testcontainers.Pulsar/PulsarBuilder.cs#L71)capabilities to PulsarBuilder. The current implementation of WithCommand appends commands instead of replacing them, making it difficult to modify container startup commands based on selected Pulsar features. In the testcontainers-java module, the container command is modified just before the container starts, enabling control over its execution. Could you provide suggestions on how to achieve a similar approach in testcontainers-dotnet?

In the java module they seem to have the control to edit it, [just before the container starts](https://github.com/testcontainers/testcontainers-java/blob/78d17685359de57dd770d3dbf5e24433d40698fb/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java#L96)

 Do you have a suggestion on a suggestion on how I should implement this for testcontainers-dotnet ?


HofmeisterAn:
> The current implementation of WithCommand appends commands instead of replacing them, making it difficult to modify container startup commands based on selected Pulsar features.

You can apply the final command at the end. We are doing this in other modules too; sometimes it is necessary to configure the [wait strategy](https://github.com/entvex/testcontainers-dotnet/blob/db9156d6707a00bace05994b637a0a7647085e58/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#L124-L125) at the end. You just need to keep a reference somewhere (in the configuration) to decide which command should be used. Or you can use a similar approach we are doing with [Kafka](https://github.com/entvex/testcontainers-dotnet/blob/db9156d6707a00bace05994b637a0a7647085e58/src/Testcontainers.Kafka/KafkaBuilder.cs#L66-L88) and upload a script that gets executed. Does that help?

entvex:
> > The current implementation of WithCommand appends commands instead of replacing them, making it difficult to modify container startup commands based on selected Pulsar features.
> 
> You can apply the final command at the end. We are doing this in other modules too; sometimes it is necessary to configure the [wait strategy](https://github.com/entvex/testcontainers-dotnet/blob/db9156d6707a00bace05994b637a0a7647085e58/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#L124-L125) at the end. You just need to keep a reference somewhere (in the configuration) to decide which command should be used. Or you can use a similar approach we are doing with [Kafka](https://github.com/entvex/testcontainers-dotnet/blob/db9156d6707a00bace05994b637a0a7647085e58/src/Testcontainers.Kafka/KafkaBuilder.cs#L66-L88) and upload a script that gets executed. Does that help?

Yes! Thanks 👍

entvex:
Hi @HofmeisterAn ,
Testcontainers.Pulsar is now ready for a code review.

entvex:
Dear @HofmeisterAn.
I can see you have started looking at my PR.
Please let me know if you need any help getting it across the finish line 💪.

src/Testcontainers.Pulsar/PulsarBuilder.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// <summary>
+    ///
+    /// </summary>
+    /// <returns></returns>
+    public PulsarBuilder WithAuthentication()
+    {
+        return Merge(DockerResourceConfiguration, new PulsarConfiguration(authenticationEnabled: true))
+            .WithEnvironment(AuthenticationEnvVars);
+    }
```
HofmeisterAn: Do we really need this API? It looks like Java does not support it either, and it prevents us from using the same [wait strategy](https://github.com/testcontainers/testcontainers-java/blob/13c2641ff1ff8e74f37594bff96275811a7d809b/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java#L109) Java uses. I have serious issues with the current wait strategy. Running `bin/pulsar-admin clusters list` seems not reliable. Tests are flaky locally.

src/Testcontainers.Pulsar/PulsarBuilder.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// <summary>
+    ///
+    /// </summary>
+    /// <returns></returns>
+    public PulsarBuilder WithFunctionsWorker(bool functionsWorkerEnabled = true)
+    {
+        // TODO: When enabled we need to adjust the wait strategy.
+        return Merge(DockerResourceConfiguration, new PulsarConfiguration(functionsWorkerEnabled: functionsWorkerEnabled));
+    }
```
HofmeisterAn: We will need to adjust the default wait strategy if [functions worker](https://github.com/testcontainers/testcontainers-java/blob/13c2641ff1ff8e74f37594bff96275811a7d809b/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java#L118-L120) is enabled (we can use a similar approach like we are doing in [Couchbase](https://github.com/testcontainers/testcontainers-dotnet/blob/9156593723eb0c083d9de2741c954806e164061f/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#L81)).

src/Testcontainers.Pulsar/PulsarBuilder.cs
entvex: @HofmeisterAn In our company's workflow, we heavily depend on WithAuthentication, especially during testing. Bringing this functionality into TestContainers would greatly help facilitate the adoption of TestContainers to all our engineers. While examining the Java version's withStrategy, I encountered some challenges in implementing the wait strategy for token authentication. Nonetheless, I'm determined to explore potential solutions and see if I can successfully integrate it

HofmeisterAn: If it helps, we can split the PR into two separate PRs. One that adds the module configuration, similar to Java, and a follow-up that specifically adds the authentication support.

src/Testcontainers.Pulsar/PulsarBuilder.cs
entvex: Very good feedback. I'll use the same approach as them 👍

entvex: I have used the approach you suggested. 😄

entvex:
@HofmeisterAn,
I have successfully integrated the approach you suggested from Couchbase into the pulsar module. During my individual testing, the tests passed without any issues. However, when I queued them to run consecutively, I noticed that the test without authentication was hanging. I am wondering if there might be an issue with the [UsedImplicitly] after you refactored the tests. Could you please take a closer look and let me know if you spot anything?

Thank you for your valuable contribution to the open-source community.

src/Testcontainers.Pulsar/PulsarBuilder.cs
```diff
@@ -0,0 +1,207 @@
+namespace Testcontainers.Pulsar;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class PulsarBuilder : ContainerBuilder<PulsarBuilder, PulsarContainer, PulsarConfiguration>
+{
+    public const string PulsarImage = "apachepulsar/pulsar:3.0.4";
```
HofmeisterAn: Support the latest version `3.2` too.

HofmeisterAn (commented):
> I am wondering if there might be an issue with the [UsedImplicitly] after you refactored the tests.

No, this controls the usage check (code smell detection).

The issue was with the `EnabledServices` implementation. You cannot store values in the builder because the builder returns new instances and loses the information without further adjustments (we made this change to support A/B testing). Therefore, the authentication use case was not properly configured. Please take a look at the latest adjustments and let me know what you think. IMO the implementation looks good. However, we should take a final look at the configuration and figure out why the latest version is incompatible. Version `3.2.2` does not start with enabled authentication.

src/Testcontainers.Pulsar/PulsarBuilder.cs
entvex: Thanks for catching the issue with Pulsar 3.2.2! We'll investigate this further, in the Pulsar community.

In the meantime, using Pulsar version 3.0.4 for TestContainers is a good choice. It's an LTS (Long-Term Support) version, which means it will be supported for a longer period compared to Pulsar 3.2.2 [(feature release with shorter support ending in 3 months).](https://pulsar.apache.org/contribute/release-policy/)

The builder's design for A/B testing makes sense. My local TestContainers tests have been running stably as well.

If you think it's ready. Feel free to merge 👍

Loydik:
Can you please complete this? This library is needed for my project. Thank you

entvex:
> Can you please complete this? This library is needed for my project. Thank you

I am waiting for a reponse from @HofmeisterAn :)

src/Testcontainers.Pulsar/PulsarBuilder.cs
HofmeisterAn: I would prefer to address the issue right at the beginning of a new module. Fixing it afterwards is usually difficult (without a module owner). While I had a quick look at the issue, it seems like there are two issues:

1) The `/pulsar/secret.key` file is not created when the wait strategy tries to create the auth token. This is simple to fix. Either wait for the file until it exists or catch the exception and handle the error.

2) When I wait until the key exists, and the token gets created successfully, I still receive occasionally: `HTTP ERROR 401 Authentication required`. Once in a while, it works. ~~When the wait strategy does not cache the auth token, it seems less likely to fail. Of course, this is not a proper fix~~.

Both findings indicate some kind of race-condition, and I am not sure if this will happen with version `3.0.4` as well.

**Edit**: When I add a 5 second delay right at the beginning of the wait strategy (before creating the auth token), the tests seem to run reliable, which confirms the race-condition we need to fix first, I guess.

```diff
diff --git a/src/Testcontainers.Pulsar/PulsarBuilder.cs b/src/Testcontainers.Pulsar/PulsarBuilder.cs
index 7b9b835..c9aa35a 100644
--- a/src/Testcontainers.Pulsar/PulsarBuilder.cs
+++ b/src/Testcontainers.Pulsar/PulsarBuilder.cs
@@ -163,6 +163,9 @@ public sealed class PulsarBuilder : ContainerBuilder<PulsarBuilder, PulsarContai
             {
                 try
                 {
+                    await Task.Delay(TimeSpan.FromSeconds(5))
+                        .ConfigureAwait(false);
+
                     _authToken = await container.CreateAuthenticationTokenAsync(TimeSpan.FromHours(1))
                         .ConfigureAwait(false);
 
diff --git a/tests/Testcontainers.Pulsar.Tests/PulsarContainerTest.cs b/tests/Testcontainers.Pulsar.Tests/PulsarContainerTest.cs
index 27ae620..dd4def9 100644
--- a/tests/Testcontainers.Pulsar.Tests/PulsarContainerTest.cs
+++ b/tests/Testcontainers.Pulsar.Tests/PulsarContainerTest.cs
@@ -73,7 +73,7 @@ public abstract class PulsarContainerTest : IAsyncLifetime
     public sealed class PulsarAuthConfiguration : PulsarContainerTest
     {
         public PulsarAuthConfiguration()
-            : base(new PulsarBuilder().WithAuthentication().Build())
+            : base(new PulsarBuilder().WithAuthentication().WithImage("apachepulsar/pulsar:3.2.2").Build())
         {
         }
 
```

HofmeisterAn:
I stumbled across this [discussion](https://github.com/apache/pulsar/discussions/22718#discussioncomment-9454990) where the friends from Go are trying to achieve something similar. Maybe this can help to sort out the issue or align the configuration.

src/Testcontainers.Pulsar/PulsarBuilder.cs
entvex: Thanks for also looking into it, you provided some very useful suggestions. I'll also look into it now 😄

entvex:
> I stumbled across this [discussion](https://github.com/apache/pulsar/discussions/22718#discussioncomment-9454990) where the friends from Go are trying to achieve something similar. Maybe this can help to sort out the issue or align the configuration.

Very interesting! Thanks for the link!

entvex:
Hi @HofmeisterAn,

I've identified a critical bug in Apache Pulsar related to token creation. The issue stems from mishandling time units (specifically, treating seconds as milliseconds). In my latest commit to this module, I've adjusted the expiration time for tokens used in the authentication to be one year. This change mitigates the impact of the token creation bug, which affected versions 3.2.0 to 3.2.3.

entvex:
The tests now run reliably, so you are free to merge 😄 @HofmeisterAn

HofmeisterAn:
> specifically, treating seconds as milliseconds

I see. Do you have any additional information, such as official documentation, a related issue, etc., regarding the workaround? Do we set the `--expiry-time` correct? I can review the PR again sometime this week :v:.

entvex:
> > specifically, treating seconds as milliseconds
> 
> I see. Do you have any additional information, such as official documentation, a related issue, etc., regarding the workaround? Do we set the `--expiry-time` correct? I can review the PR again sometime this week ✌️.

Jep, we have set the `--expiry-time` correct. But as I mentioned before, the bug is in pulsar. I have already talked to another Apache Pulsar PMC. I'll create an issue and submit a PR for pulsar, but I have not created them yet. Therefore, I can't link to them here.

src/Testcontainers.Pulsar/PulsarContainer.cs
```diff
@@ -58,9 +58,14 @@ public async Task<string> CreateAuthenticationTokenAsync(TimeSpan expire = defau
...
             "--expiry-time",
-            $"{expire.TotalSeconds}s",
+            "",
         };
 
+        if (_configuration.Image.FullName is "apachepulsar/pulsar:3.2.0" or "apachepulsar/pulsar:3.2.1" or "apachepulsar/pulsar:3.2.2" or "apachepulsar/pulsar:3.2.3")
+            command[8] = $"{expire.TotalSeconds * 1000}s";
+        else
+            command[8] = $"{expire.TotalSeconds}s";
+
```
HofmeisterAn: QQ: Cannot we simply use something like: `$"{expire.TotalMilliseconds}ms"` for all versions?

entvex: > QQ: Cannot we simply use something like: `$"{expire.TotalMilliseconds}ms"` for all versions?

If we do that, when we will have issues in this module when the issue gets fixed in pulsar. We will also have issues in any other version that is not 3.2.X

HofmeisterAn: My assumption is that if the flag `--expiry-time` supports `y` and `s`, does it also support `ms`? Which units of time are supported? Does the mentioned bug only occur for `s`? What about days — can we use something like `0.0416d` (~ 1 hour)? I am not very happy with the current workaround because a new Pulsar version without "the fix" will fail again.

entvex: > My assumption is that if the flag `--expiry-time` supports `y` and `s`, does it also support `ms`?

As you can see in the [source code](https://github.com/apache/pulsar/blob/c39f9f82b425c66c899f818583714c9c98d3e213/pulsar-cli-utils/src/main/java/org/apache/pulsar/cli/converters/RelativeTimeUtil.java), milliseconds (ms) are currently not supported. I’m actively working on a pull request to address this issue. Once fixed, the lowest supported unit will remain seconds (s). Because it's a fix for a regression.

> Which units of time are supported?

If you examine the referenced Java file, you’ll find that the following time units are supported: seconds (s), minutes (m), hours (h), days (d), weeks (w), and years (y).

> Does the mentioned bug only occur for `s`? What about days — can we use something like `0.0416d` (~ 1 hour)?

The mentioned bug affects all time units, not just seconds

> I am not very happy with the current workaround because a new Pulsar version without "the fix" will fail again.

I agree that the current workaround isn’t ideal because a new Pulsar version without the fix will encounter the same issue. To mitigate this, we could consider setting the TimeSpan in the tests and the UntilAsync method to a sufficiently long duration. Additionally, we can add an explanatory note in the user-facing method CreateAuthenticationTokenAsync to inform users about this issue in versions 3.2.X.

Or as I suggested [here](https://github.com/testcontainers/testcontainers-dotnet/pull/1103#discussion_r1576063779) set the default version to a LTS release that does not have the bug.

Let’s collaborate to find the best path forward! 😊

You can find the Issue in the Apache Pulsar Repo [here](https://github.com/apache/pulsar/issues/22811)

HofmeisterAn: > The mentioned bug affects all time units, not just seconds

Oh, that's unfortunate.

> Let’s collaborate to find the best path forward! 😊

I propse the following (not sure about `latest` though):

```csharp
int secondsToMilliseconds;

if (_configuration.Image.Tag.StartsWith("3.2") || _configuration.Image.Tag.StartsWith("latest"))
{
    Logger.LogWarning("The 'apachepulsar/pulsar:3.2.?' image contains a regression. The expiry time is converted to the wrong unit of time: https://github.com/apache/pulsar/issues/22811.");
    secondsToMilliseconds = 1000;
}
else
{
    secondsToMilliseconds = 1;
}

var command = new[]
{
    "bin/pulsar",
    "tokens",
    "create",
    "--secret-key",
    PulsarBuilder.SecretKeyFilePath,
    "--subject",
    PulsarBuilder.Username,
    "--expiry-time",
    $"{secondsToMilliseconds * expire.TotalSeconds}s",
};
```

WDYT?

entvex: I think that works, I just checked and latest does indeed point to 3.2.3. So it should be good for now.
We can remove latest check, when the fix is released.
I think that this should unblock the test containers module?

Before you merge this PR in, will you make a GitHub do a git squash?

HofmeisterAn: > I think that this should unblock the test containers module?

👍 

> Before you merge this PR in, will you make a GitHub do a git squash?

Yes, I will take care of that.

HofmeisterAn (approved):
Thanks for the PR, the efforts, and the patience. Great first contribution 🙏.

entvex:
> Thanks for the PR, the efforts, and the patience. Great first contribution 🙏.

Thanks a lot! It was a pleasure working with you. I hope we can collaborate more modules soon 👍

smbecker:
I noticed that this merged. Do you have an estimate on when it will be available in [nuget](https://www.nuget.org/packages/Testcontainers.Pulsar)?

PR #1111, opened 2024-02-10
chore: Use the `pg_isready` command to assess whether PostgreSQL is ready or not

0xced:
This strategy aligns with the MySQL and MariaDb wait strategies.

Also, logs have changed over time and the current strategy doesn't work with older PostgreSQL Docker images (for example v9.6) which does not produce the same logs as recent versions of the PostgreSQL docker image.


HofmeisterAn (approved):
👍

PR #1113, opened 2024-02-11
chore: Always run the Upload Test And Coverage Results step

0xced:
Use `if: always() `instead of `if: true` in order to always run a step even if the previous one fails. This is surprising but that's how GitHub actions work, see https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f#comment112735158_58859404

Proof that `if: true` does not work.

Screenshot 2024-02-11 at 13 24 19


HofmeisterAn (commented):
Thanks

PR #1116, opened 2024-02-13
chore: Ensure that stderr is empty after executing scripts

0xced:
## What does this PR do?

This pull request add assertions to all tests that call `ExecScriptAsync` asserting that no error was written to stderr.

Adding these assertions found two issues, in the MySQL and MongoDB implementations.

MySQL would produce the following warning on stderr:
```
mysql: [Warning] Using a password on the command line interface can be insecure.
```

This has been addressed in 925f426081d8188dcc510898692f37db6e2933b2 by writing the configuration into `/etc/mysql/my.cnf` instead of passing it on the command line.

MongoDB would produce the following error on stderr (when using the `mongo:4.4` image in the `MongoDbV4Configuration` test):
```
/bin/sh: 1: mongosh: not found
```

This has been addressed in 3ceeb4b02272d1f9d444fbc35e408e4fc2127192 by verifying if the `mongosh` binary exists and choosing the proper binary instead of running `mongosh || mongo` through `/bin/sh`. This also simplified the mongosh command implementation.

Finally, I also fixed some typos: `// When` should be `// Then` in the assertion part of the tests.

## Why is it important?

Having stderr empty after successfully executing a script ensures that stderr will only contain actual errors and not spurious warnings or errors that are implementation detail errors that don't actually matter.


src/Testcontainers.MySql/MySqlContainer.cs
```diff
@@ -45,7 +46,20 @@ public async Task<ExecResult> ExecScriptAsync(string scriptContent, Cancellation
...
+            var config = new StringWriter { NewLine = "\n" };
+            config.WriteLine("[client]");
+            config.WriteLine("protocol=TCP");
+            config.WriteLine($"user={_configuration.Username}");
+            config.WriteLine($"password={_configuration.Password}");
+            await CopyAsync(Encoding.Default.GetBytes(config.ToString()), "/etc/mysql/my.cnf", UnixFileModes.UserRead | UnixFileModes.UserWrite, ct)
+                .ConfigureAwait(false);
+
+            _hasConfigFile = true;
+        }
```
HofmeisterAn: I was thinking of copying the file only once (without `_hasConfigFile` and "thread-safe"). We could make use of `WithStartupCallback(Func<IContainer, CancellationToken, Task>)`. However, this requires some minor adjustments I have been thinking about for a while.

If we change (overload) the signature to `WithStartupCallback(Func<TContainerEntity, CancellationToken, Task> startupCallback)`, we can access the actual container and its configuration. Then, we can simply upload the file once at the start. WDYT?

```csharp
_ = MySqlBuilder()
    .WithStartupCallback(Foo)
    .Build();

public static Task Foo(MySqlContainer container, CancellationToken ct)
{
    return container.CopyAsync(Encoding.Default.GetBytes("Hello, World!"), "/tmp/foo", ct: ct);
}
```

Unfortunately, this requires a cast to `TContainerEntity`. Even though I would prefer to avoid casts, I believe in this case, it is acceptable.

```csharp
public TBuilderEntity WithStartupCallback(Func<TContainerEntity, CancellationToken, Task> startupCallback)
{
    return Clone(new ContainerConfiguration(startupCallback: (container, ct) => startupCallback((TContainerEntity)container, ct)));
}
```

0xced: I did not know about `WithStartupCallback` and this is indeed a good solution! I applied your suggestion in 0fcffb1e1e362c3c0358320dd10746b24fa4ddfd with the cast from `IContainer` to `TContainerEntity`. (I tried to make `IContainerConfiguration` generic to avoid the cast but it did not end well at all so I gave up.)

I'm just a bit worried that the startup callback can be _hijacked_ by the user. Unlike wait strategies, there's only one startup callback and if the user decides to use `WithStartupCallback` then it will break the `ExecScriptAsync` feature since the username and password won't be provided.

I see that it's already the case for some modules. For example, I tried to modify the  Couchbase test like this:
```csharp
_couchbaseContainer = new CouchbaseBuilder().WithStartupCallback((_, _) => Task.CompletedTask).Build();
```
This breaks the test (the wait strategy hangs forever) since `ConfigureCouchbaseAsync` is not called anymore because the startup callback has been _hijacked_.

HofmeisterAn: > I see that it's already the case for some modules. For example, I tried to modify the Couchbase test like this:
> 
> ```cs
> _couchbaseContainer = new CouchbaseBuilder().WithStartupCallback((_, _) => Task.CompletedTask).Build();
> ```
> 
> This breaks the test (the wait strategy hangs forever) since `ConfigureCouchbaseAsync` is not called anymore because the startup callback has been _hijacked_.

The module configurations are quite opinionated. Developers can break modules easily by overriding certain default configurations, as you mentioned. The same applies to the entrypoint, command, etc., and this also applies to the Docker run command. Even here, developers can override, for example, the entrypoint and break the default configuration (container start). If developers really need to customize some default configurations, they can always fall back and use the generic builder approach.

HofmeisterAn (approved):

HofmeisterAn:
Thanks again.=

HofmeisterAn:
Thanks again.

PR #1117, opened 2024-02-14
chore: Ensure that ConfigureAwait is always called on awaited tasks

0xced:
## What does this PR do?

This pull requests enables [code quality analysis](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview#code-quality-analysis) and turns [CA2007: Do not directly await a Task](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007) into an error.

## Why is it important?

Testcontainers for .NET calls `ConfigureAwait(false)` everywhere so this pull request enforces this rule so that authors contributors can't forget about it.

## Follow-ups

Note that this was only enabled in the `src` directory and not at the root of the project to avoid enabling it for the test projects. It becomes problematic in `TestcontainersContainerTest` because of `await using` of IAsyncDisposable where it would generate an error.

And now the big question: is it really necessary to always call `.ConfigureAwait(true)` everywhere in the tests? I understand why calling `.ConfigureAwait(false)` is important in library code but I don't think it matters at all for the test project. Does it? I think we could suppress all `.ConfigureAwait(true)` in the tests and increase the readability of the tests.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,6 +25,7 @@
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
   <PropertyGroup>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <NoWarn>CS0618,CS1591,SA0001,SA1600,SA1633,SA1649,CA1859,CA1861</NoWarn>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -1,0 +1,2 @@
+[*.cs]
+dotnet_diagnostic.CA2007.severity = error


HofmeisterAn (commented):
I am not a huge fan of this PR and change. A couple of months ago (I think with the introduction of modules), I removed all of this stuff. It was just annoying to see that the output was inconsistent. Some configurations are not supported by all IDEs or are treated differently (maybe it got better with .NET 8). We have a code analysis tool enabled, and I am fine with Sonar, even though we cannot run it for PRs (forks). I will just do the cleanup afterwards, and that is fine.

> And now the big question: is it really necessary to always call `.ConfigureAwait(true)` everywhere in the tests?

I know it is unnecessary (because it is the default configuration), but it does not bother me. I prefer being explicit. That is all. I would like to stick to it for now.

> but I don't think it matters at all for the test project.

It depends on the test framework, see: [xUnit1030](https://github.com/xunit/xunit/blob/gh-pages/xunit.analyzers/_rules/xUnit1030.md).

0xced:
> I will just do the cleanup afterwards, and that is fine.

👍

> It depends on the test framework, see: [xUnit1030](https://github.com/xunit/xunit/blob/gh-pages/xunit.analyzers/_rules/xUnit1030.md).

Thanks for that interesting piece of information!

PR #1118, opened 2024-02-14
feat: Add URL-based resource mapping container builder API

0xced:
## What does this PR do?

This pull request introduces a new URL-based resource mapping, allowing to easily copying a file accessible over HTTP or HTTPS directly into a container before it starts, just like one could copy a local file.

## Why is it important?

The `UriResourceMapping` is straightforward to write and would not _need_ to be included in the Testcontainers library but I think having it built-in is a nice addition.

### DIFF
--- a/src/Testcontainers/Builders/ContainerBuilder`3.cs
+++ b/src/Testcontainers/Builders/ContainerBuilder`3.cs
@@ -197,7 +197,7 @@ namespace DotNet.Testcontainers.Builders
     /// <inheritdoc />
     public TBuilderEntity WithResourceMapping(string source, string target, UnixFileModes fileMode = Unix.FileMode644)
     {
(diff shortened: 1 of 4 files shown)


HofmeisterAn (dismissed):
Thanks! The PR looks good, and I think it is a nice enhancement. I have added additional tests to cover the implementation of `UriResourceMapping` as well. While adding the tests, I noticed a minor misunderstanding. The `WithResourceMapping(string, string, UnixFileModes)` always copies files to the target directory path, never to a file path. If you want to copy a file to a particular target file path, you need to use the overload `WithResourceMapping(FileInfo, FileInfo, UnixFileModes)` (this was necessary due to backward compatibility). I updated the docs and tried to explain the behavior better. I will merge it later. Thanks again.

HofmeisterAn (approved):

PR #1129, opened 2024-02-24
feat: Embed symbols and enable continuous integration builds (deterministic source paths)

### LINKED ISSUE #1128: [Enhancement]: Use deterministic builds with SourceLink support
### Problem

I had a hard time tracking down a crash (#1125) because the stack trace only showed "external code"

### Solution

Provide a better debugging experience via SourceLink 

### Benefit

A much better debugging experience

### Alternatives

None

Yes
```[tasklist]
### Tasks
- [ ] https://github.com/testcontainers/testcontainers-dotnet/pull/1129
```

tom-englert:
## What does this PR do?

Use embedded symbols instead of separate symbol package

## Why is it important?

Works better across all IDEs

## Related issues

- Closes #1128

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -31,6 +31,9 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
+  <PropertyGroup>
+    <DebugType>embedded</DebugType>
+  </PropertyGroup>
   <ItemGroup>
     <None Include="$(MSBuildThisFileDirectory)docs/banner.png" Visible="false" Pack="true" PackagePath="docs/" />
     <None Include="$(MSBuildThisFileDirectory)docs/logo.png" Visible="false" Pack="true" PackagePath="docs/" />
--- a/build.cake
+++ b/build.cake
@@ -67,6 +67,7 @@ Task("Build")
     Verbosity = param.Verbosity,
     NoRestore = true,
     ArgumentCustomization = args => args
+      .Append($"/p:ContinuousIntegrationBuild=true")
   });
 });
 
@@ -130,12 +131,8 @@ Task("Create-NuGet-Packages")
     Verbosity = param.Verbosity,
     NoRestore = true,
     NoBuild = true,
-    IncludeSymbols = true,
-    SymbolPackageFormat = "snupkg",
     OutputDirectory = param.Paths.Directories.NuGetDirectoryPath,
     ArgumentCustomization = args => args
-      .Append("/p:ContinuousIntegrationBuild=true")
-      .Append("/p:EmbedUntrackedSources=true")
       .Append($"/p:Version={param.Version}")
   });
 });
@@ -160,7 +157,7 @@ Task("Publish-NuGet-Packages")
   .WithCriteria(() => param.ShouldPublish)
   .Does(() =>
 {
-  foreach(var package in GetFiles($"{param.Paths.Directories.NuGetDirectoryPath}/*.(nupkg|snupkgs)"))
+  foreach (var package in GetFiles($"{param.Paths.Directories.NuGetDirectoryPath}/*.nupkg"))
   {
     DotNetNuGetPush(package.FullPath, new DotNetNuGetPushSettings
     {
--- a/src/Testcontainers/Builders/CommonDirectoryPath.cs
+++ b/src/Testcontainers/Builders/CommonDirectoryPath.cs
@@ -107,17 +107,17 @@ namespace DotNet.Testcontainers.Builders
 
     private static string GetDirectoryPath(string path, string searchPattern)
     {
-      return GetDirectoryPath(Directory.Exists(path) ? new DirectoryInfo(path) : null, searchPattern, path);
+      return GetDirectoryPath(Directory.Exists(path) ? new DirectoryInfo(path) : null, searchPattern);
     }
 
-    private static string GetDirectoryPath(DirectoryInfo path, string searchPattern, string originalPath)
+    private static string GetDirectoryPath(DirectoryInfo path, string searchPattern)
     {
       if (path != null)
       {
-        return path.EnumerateFileSystemInfos(searchPattern, SearchOption.TopDirectoryOnly).Any() ? path.FullName : GetDirectoryPath(path.Parent, searchPattern, originalPath);
+        return path.EnumerateFileSystemInfos(searchPattern, SearchOption.TopDirectoryOnly).Any() ? path.FullName : GetDirectoryPath(path.Parent, searchPattern);
       }
 
-      var message = $"Cannot find '{searchPattern}' and resolve the base directory in the directory tree of '{originalPath}'.";
+      var message = $"Cannot find '{searchPattern}' and resolve the base directory in the directory tree.";
       throw new DirectoryNotFoundException(message);
     }
   }
--- a/src/Testcontainers/Builders/CommonDirectoryPath.cs
+++ b/src/Testcontainers/Builders/CommonDirectoryPath.cs
@@ -107,17 +107,17 @@ namespace DotNet.Testcontainers.Builders
 
     private static string GetDirectoryPath(string path, string searchPattern)
     {
-      return GetDirectoryPath(Directory.Exists(path) ? new DirectoryInfo(path) : null, searchPattern);
+      return GetDirectoryPath(Directory.Exists(path) ? new DirectoryInfo(path) : null, searchPattern, path);
     }
 
-    private static string GetDirectoryPath(DirectoryInfo path, string searchPattern)
+    private static string GetDirectoryPath(DirectoryInfo path, string searchPattern, string originalPath)
     {
       if (path != null)
       {
-        return path.EnumerateFileSystemInfos(searchPattern, SearchOption.TopDirectoryOnly).Any() ? path.FullName : GetDirectoryPath(path.Parent, searchPattern);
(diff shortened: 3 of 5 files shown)


HofmeisterAn (commented):
According to this [comment](https://github.com/testcontainers/testcontainers-dotnet/pull/1054#issuecomment-1828082264) and this [section](https://github.com/dotnet/sourcelink#using-source-link-in-net-projects), Source Link is now included in the SDK. The added properties, among others, are set in our Cake build [file](https://github.com/testcontainers/testcontainers-dotnet/blob/8aee376aa98018d07d270264920161113dfd8490/build.cake#L138). According to the Source Link docs, they are not necessary anymore, either. Probably, we can even remove them.

Is something not working? I haven't had issues in the past debugging the library. The NuGet Package Explorer looks Ok too (although, we need to move `ContinuousIntegrationBuild` to Cake's build task).

image

tom-englert:
It should look like this:
- no symbol server
- deterministic, including repo ulr & commit
image

I'll have a look whats wrong with this...

HofmeisterAn:
> It should look like this:

I see, thanks! Maybe we just need to move the properties (args) to the Cake build task (the pack task does not rebuild the projects):

```diff
diff --git a/build.cake b/build.cake
index 1d2e4ae..0843d4e 100644
--- a/build.cake
+++ b/build.cake
@@ -67,6 +67,8 @@ Task("Build")
     Verbosity = param.Verbosity,
     NoRestore = true,
     ArgumentCustomization = args => args
+      .Append("/p:ContinuousIntegrationBuild=true")
+      .Append("/p:EmbedUntrackedSources=true")
   });
 });
 
@@ -134,8 +136,6 @@ Task("Create-NuGet-Packages")
     SymbolPackageFormat = "snupkg",
     OutputDirectory = param.Paths.Directories.NuGetDirectoryPath,
     ArgumentCustomization = args => args
-      .Append("/p:ContinuousIntegrationBuild=true")
-      .Append("/p:EmbedUntrackedSources=true")
       .Append($"/p:Version={param.Version}")
   });
 });
```

tom-englert:
Probably yes, and the symbols should be embedded rather than having a separate symbol package, from my experience the latter never works without friction.

tom-englert:
Also maybe .Append("/p:ContinuousIntegrationBuild=true") should not really be needed, SourceLink should determine this automatically

tom-englert:
https://github.com/dotnet/sourcelink/blob/a22426a46745b8fe77b9d306fa4bc1d03eff44ea/src/Microsoft.Build.StandardCI/build/Microsoft.Build.StandardCI.props#L15-L19

tom-englert:
This is what I got with the current version of TestContainers:
- listed in the "External Sources", but seems not to match the symbols found on the symbol server.
image

Directory.Build.props
```diff
@@ -31,6 +31,10 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
+  <PropertyGroup>
+    <DebugType>embedded</DebugType>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
```
HofmeisterAn:
```suggestion
```

I did some local tests, and it appears that this property is indeed not necessary.

build.cake
```diff
@@ -130,12 +130,9 @@ Task("Create-NuGet-Packages")
     Verbosity = param.Verbosity,
     NoRestore = true,
     NoBuild = true,
-    IncludeSymbols = true,
-    SymbolPackageFormat = "snupkg",
+    IncludeSymbols = false,
```
HofmeisterAn:
```suggestion
```

Its default value is `false`.

build.cake
```diff
@@ -160,7 +157,7 @@ Task("Publish-NuGet-Packages")
   .WithCriteria(() => param.ShouldPublish)
   .Does(() =>
 {
-  foreach(var package in GetFiles($"{param.Paths.Directories.NuGetDirectoryPath}/*.(nupkg|snupkgs)"))
+  foreach(var package in GetFiles($"{param.Paths.Directories.NuGetDirectoryPath}/*.nupkg"))
```
HofmeisterAn:
```suggestion
  foreach (var package in GetFiles($"{param.Paths.Directories.NuGetDirectoryPath}/*.nupkg"))
```

HofmeisterAn (changes requested):
> This is what I got with the current version of TestContainers:

This is strange. Sorry, I do not have access to Visual Studio. Rider seems to be able to download the symbols, etc. I can debug into the sources without any issues. However, we can merge the suggestions. The NuGet Package Explorer looks even better with them. Thanks.

HofmeisterAn:
I published the pipeline artifacts (NuGets) to ensure that the NuGets are built and configured properly, and everything is set up correctly for the next release. It appears that `ContinuousIntegrationBuild` is somehow not automatically applied to the build. We need to set it explicitly ([before](https://github.com/testcontainers/testcontainers-dotnet/actions/runs/8031926977) and [after](https://github.com/testcontainers/testcontainers-dotnet/actions/runs/8031957625)).

image

tom-englert:
Strange, I never had to set this - however I never used Cake, maybe that's special about this.

HofmeisterAn:
Probably, we cannot enable `ContinuousIntegrationBuild` that easily 😐. It obviously breaks our [CommonDirectoryPathTest](https://github.com/testcontainers/testcontainers-dotnet/runs/21941308921#r44s1) tests. Not sure how (or if) we can solve this, because we are using the `CallerFilePath` attribute to resolve common directory paths:

https://github.com/testcontainers/testcontainers-dotnet/blob/5c677d9ac043d414ecc3c165c265700c98668808/src/Testcontainers/Builders/CommonDirectoryPath.cs#L61-L65

tom-englert:
> It obviously breaks our [CommonDirectoryPathTest](https://github.com/testcontainers/testcontainers-dotnet/runs/21941308921#r44s1) tests

That's strange, how can this have such a side effect? And why only on linux?

HofmeisterAn:
> > It obviously breaks our [CommonDirectoryPathTest](https://github.com/testcontainers/testcontainers-dotnet/runs/21941308921#r44s1) tests
> 
> That's strange, how can this have such a side effect? And why only on linux?

I think I found the "reason" and have a potential working configuration, which also explains why you never had to set `ContinuousIntegrationBuild` explicitly. I need a few more minutes to validate my changes.

HofmeisterAn:
If we set `ContinuousIntegrationBuild` to `true` during the build, it will break the tests because `CallerFilePath` now returns a deterministic root path. We can simply disable `ContinuousIntegrationBuild` for the test projects. 

My [assumption](https://github.com/testcontainers/testcontainers-dotnet/commit/e6f1746956e165c1a7e569d55a17b63c833ed46d) was, if we allow Cake to rebuild the packages on `Pack` (if the packages are already built, we likely need to force a rebuild), the `ContinuousIntegrationBuild` is applied automatically, and it won't interfere with the tests. Unfortunately, the configuration is not applied automatically...

HofmeisterAn:
I do not think we need that many changes. The following [changes](https://github.com/tom-englert/testcontainers-dotnet/compare/%231128...testcontainers:testcontainers-dotnet:feature/tom-englert-%231128) work [fine](https://github.com/testcontainers/testcontainers-dotnet/actions/runs/8037822226) (incl. Sonar). We just need to make sure the pack target rebuilds the NuGets. I am still wondering why we need to set `ContinuousIntegrationBuild` explicitly, but maybe Source Link behaves differently if the source control provider is added explicitly.

image

tom-englert:
Just needed to turn off deterministic for the tests, they don't get deployed anyhow.

src/Testcontainers/Builders/CommonDirectoryPath.cs
```diff
@@ -107,17 +107,17 @@ public static CommonDirectoryPath GetCallerFileDirectory([CallerFilePath, NotNul
...
       if (path != null)
       {
-        return path.EnumerateFileSystemInfos(searchPattern, SearchOption.TopDirectoryOnly).Any() ? path.FullName : GetDirectoryPath(path.Parent, searchPattern);
+        return path.EnumerateFileSystemInfos(searchPattern, SearchOption.TopDirectoryOnly).Any() ? path.FullName : GetDirectoryPath(path.Parent, searchPattern, originalPath);
       }
 
-      var message = $"Cannot find '{searchPattern}' and resolve the base directory in the directory tree.";
+      var message = $"Cannot find '{searchPattern}' and resolve the base directory in the directory tree of '{originalPath}'.";
       throw new DirectoryNotFoundException(message);
     }
```
HofmeisterAn: Can we please revert these changes and only include changes that are necessary to embed the symbols / enable deterministic paths?

tests/Testcontainers.Tests/Testcontainers.Tests.csproj
```diff
@@ -9,6 +9,10 @@
   <PropertyGroup>
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
+  <PropertyGroup>
+    <EnableSourceLink>false</EnableSourceLink>
+    <Deterministic>false</Deterministic>
+  </PropertyGroup>
```
HofmeisterAn: Can we please move this to `tests/Directory.Build.props` so that the properties apply to the other test projects too?

tom-englert: Sure, just wasn't aware that this already exists.

HofmeisterAn (approved):
Thank you for the efforts, and apologies for the back and forth.

tom-englert:
Happy to contribute to such a great tool! Thanks for all YOUR efforts 😄

PR #1131, opened 2024-02-28
feat: Add Milvus module

roji:
This adds a test container for the [Milvus vector database](https://milvus.io).

This largely follows the existing Milvus Java [testcontainer](https://github.com/testcontainers/testcontainers-java/blob/main/modules/milvus/src/test/java/org/testcontainers/milvus/MilvusContainerTest.java); potentially more configuration knobs could be exposed, but we can do that if/when users request them.

Let me know about any adjustments you'd like made!

/cc @yhmo @weianweigan


src/Testcontainers.Milvus/MilvusBuilder.cs
```diff
@@ -0,0 +1,75 @@
...
+            builder = builder
+                .WithResourceMapping(Encoding.UTF8.GetBytes(etcdYaml), "/milvus/configs/embedEtcd.yaml")
+                .WithEnvironment("ETCD_USE_EMBED", "true")
+                .WithEnvironment("ETCD_DATA_DIR", "/var/lib/milvus/etcd")
+                .WithEnvironment("ETCD_CONFIG_PATH", "/milvus/configs/embedEtcd.yaml");
+        }
+        else
+        {
+            throw new NotImplementedException();
+        }
```
HofmeisterAn: Hi, I do not understand this part. `_etcdEndpoint` will always be `null` unless you store it in the container configuration object. Both tests use the variation that includes the `embedEtcd.yaml` file.

The builder API is immutable to support certain configurations for A/B testing. The library creates a new instance of the builder; hence, you cannot keep the reference to the field without storing it somewhere else.

I would expect (propose) the following default configuration:

```csharp
protected override MilvusBuilder Init()
{
    return base.Init()
        .WithImage(MilvusImage)
        .WithPortBinding(MilvusManagementPort, true)
        .WithPortBinding(MilvusGrpcPort, true)
        .WithCommand("milvus", "run", "standalone")
        .WithEnvironment("COMMON_STORAGETYPE", "local")
        .WithEnvironment("ETCD_CONFIG_PATH", EtcdConfigFilePath)
        .WithEnvironment("ETCD_DATA_DIR", "/var/lib/milvus/etcd")
        .WithEnvironment("ETCD_USE_EMBED", "true")
        .WithResourceMapping(DefaultEtcdConfig, EtcdConfigFilePath)
        .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request =>
            request.ForPort(MilvusManagementPort).ForPath("/healthz")));
}
```

Then override the default configuration from the other available public builder APIs, that is how we do it in other modules (_there might be other (better) ways to pass the endpoint to the container, but I think you know better_):

```csharp
public MilvusBuilder WithEtcdEndpoint(string etcdEndpoint)
{
    return WithEnvironment("ETCD_USE_EMBED", "false")
        .WithEnvironment("ETCD_ENDPOINTS", etcdEndpoint);
}
```

This will run both variations. One that uses the embedded etcd and one that uses the sidecar container `quay.io/coreos/etcd`. I think that is what you are trying to achieve?

If you need further guidance or help, just let me know. Happy to assist.

roji: Yeah, apologies, the part with an external etcd indeed wasn't finished.

I've made changes following your suggestion - that all makes sense to me. It's slightly odd to define the "embedded" settings as a default and then to override them when WithEtcdEndpoint is called - ideally the embedded settings wouldn't be set at all. For example, as the code is after my change, the embedEtcd.yaml file is always mapped in the container, though it's really needed only for the embedded config (and I'm not sure there's a way to unset environment variables and unmap a resource).

But that's not very important - I'm definitely happy with things as they are if you are. Re a better way for passing the endpoint to the container, I'm not expert here and am mainly following the lead of the [Java module](https://github.com/testcontainers/testcontainers-java/blob/main/modules/milvus/src/main/java/org/testcontainers/milvus/MilvusContainer.java) (we can always improve this in the future etc.)

src/Testcontainers.Milvus/Testcontainers.Milvus.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>net6.0;netstandard2.0;net462</TargetFrameworks>
```
HofmeisterAn: Regarding this, I have a question; maybe you know more about it. Are there significant benefits for users if a library targets the latest frameworks, in addition to .NET Standard, **especially** with transitive dependencies that still target .NET Standard?

I know, as the consumer (library maintainer) of the framework (SDK), I get access to more APIs, etc. But what is exactly the benefit for the Testcontainers user targeting `net6.0;netstandard2.0;net462` instead of `netstandard2.0;netstandard2.1`? I did some research and have not found much information.

src/Testcontainers.Milvus/MilvusBuilder.cs
HofmeisterAn: > Yeah, apologies, the part with an external etcd indeed wasn't finished.

No worries. That is what reviews are for.

> or example, as the code is after my change, the embedEtcd.yaml file is always mapped in the container, though it's really needed only for the embedded config (and I'm not sure there's a way to unset environment variables and unmap a resource).

Indeed. This is a bit tricky. Overriding environment variables works fine, but unmapping resources does not. Definitely something I would like to look at and support in the future. Some configurations cannot be unset.

> we can always improve this in the future etc.

I agree. In the future, we can provide, e.g. the sidecar container together with the module, something we are already doing in other modules like [FFmpeg](https://github.com/testcontainers/testcontainers-dotnet/blob/2131fbd27ef394e602b454793216136440584393/src/Testcontainers.WebDriver/WebDriverBuilder.cs#L74-L84). Anyway, the module configuration looks good. Thanks for the adjustments.

HofmeisterAn (commented):
Thanks for the PR. The module configurations look good 👍. Do you mind if I update some parts to align the module with all the others? Usually, this helps me to refactor stuff that involves changes in the modules too. Either way, I am happy to merge the PR afterwards.

src/Testcontainers.Milvus/Testcontainers.Milvus.csproj
roji: > Do you mind if I update some parts to align the module with all the others?

Absolutely - please feel free to make any adjustments necessary to align this with the repo standards or anything else!

> But what is exactly the benefit for the Testcontainers user targeting net6.0;netstandard2.0;net462 instead of netstandard2.0;netstandard2.1?

On this... For net462 specifically, the .NET Framework build of the Milvus client uses the .NET Framework build of Grpc.Net.Client, because that automatically uses WinHttpHandler; if you target netstandard2.0, that helpful feature isn't there. tl;dr it makes the experience easier on .NET Framework, where some extra config/wiring up  is necessary.

For net6.0, I think the main reason there's that TFM is to not bring in various dependencies that are necessarily on .NET Standard (e.g. Microsoft.Bcl.HashCode) but are already included in the box in net6.0. So you get cleaner dependency graph etc. Of course, in theory there may also be new APIs exposed that depend on newer .NET functionality, but specifically for Milvus.Client there aren't any IIRC.

Hopefully that all makes sense - let me know if you have any further questions!

HofmeisterAn: > I think the main reason there's that TFM is to not bring in various dependencies that are necessarily on .NET Standard (e.g. Microsoft.Bcl.HashCode) but are already included in the box in net6.0.

I see, I didn't know that the TFM uses its own implementation instead of the one from the dependency. Then it definitely makes sense to target TC for other TFMs too. Thanks 🙏.

roji: Yeah, this is basically for functionality that got added at some point (e.g. to net6.0), but a compat nuget was released to also support older versions. It's no big deal if you use netstandard2.0 and get that package, but slightly cleaner without it.

HofmeisterAn (approved):

roji:
Thanks for merging @HofmeisterAn! Looking forward to seeing this on nuget.org!

roji:
@HofmeisterAn let me know if I can give a hand in updating the [docs](https://testcontainers.com/modules/milvus/), or any other task needed to getting this published/doc'ed etc.

HofmeisterAn:
The module is published together with a new Testcontainers version/release. If you like, I can create a pull request and add the .NET implementation to the modules catalog after the release.

roji:
That would be great, thanks!

PR #1135, opened 2024-03-04
feat: Add Gitlab CE module

Peet-HD:
## What does this PR do?

Implemented a TestContainer for Gitlab. This can be used to test projects that use the API of gitlab without relying on an actual instance of gitlab.

## Why is it important?

This can be super useful to create integration tests for projects that need access to the Gitlab-API. For example this can be used to test a tool that generates an merge request after running a custom pipeline. 

## How to test this PR

This PR can be tested if someone generates a personal access token and uses it to authenticates against the gitlab testcontainer.

To run this test in your shell:
```bash
dotnet test "tests/Testcontainers.Gitlab.Tests/Testcontainers.Gitlab.Tests.csproj"
```


HofmeisterAn:
Sorry for the late response. I will try to review the module next week.

src/Testcontainers.Gitlab/GitlabBuilder.cs
```diff
@@ -0,0 +1,78 @@
+ namespace Testcontainers.Gitlab;
+ 
+ /// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+ [PublicAPI]
+ public sealed class GitlabBuilder : ContainerBuilder<GitlabBuilder, GitlabContainer, GitlabConfiguration>
+ {
+     /// <summary>
+     /// This is the default image for gitlab community edition.
+     /// </summary>
+     public const string GitlabImage = "gitlab/gitlab-ce";
```
HofmeisterAn: Can we please pin the version?

src/Testcontainers.Gitlab/GitlabBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+     {
+         return base.Init()
+             .WithImage(GitlabImage)
+             .WithPortBinding(GitlabHttpPort, true)
+             .WithPortBinding(GitlabSshPort, true)
+             .WithWaitStrategy(Wait.ForUnixContainer().UntilFileExists("/etc/gitlab/initial_root_password"))
+             .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(80))
+             .WithWaitStrategy(Wait.ForUnixContainer().UntilPortIsAvailable(22))
+             .WithWaitStrategy(Wait.ForUnixContainer().UntilContainerIsHealthy())
+             .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request => request.ForPath("/users/sign_in").ForStatusCode(HttpStatusCode.OK)));
```
HofmeisterAn: This overrides the previous wait strategies. You need to add the wait strategy once (you can chain the until members though).

src/Testcontainers.Gitlab/GitlabContainer.cs
```diff
@@ -0,0 +1,93 @@
...
+        });
+
+
+    /// <summary>
+    /// This method returns the initial root password for the gitlab root user.
+    /// </summary>
+    /// <returns>Returns the initial root password generated by gitlab.</returns>
+    private async Task<string> GetInitialRootPassword()
+    {
+        var byteArray = await ReadFileAsync("/etc/gitlab/initial_root_password");
```
HofmeisterAn: Testcontainers allows uploading files before the container starts. Can we upload the `initial_root_password` file to specify the password? Can we simply use basic authentication? This would make the module simpler and should be enough for testing purposes (I guess).

HofmeisterAn (commented):
Thanks for the PR. I have some questions before I continue the review.

src/Testcontainers.Gitlab/GitlabBuilder.cs
Peet-HD: Sure thing. I tested with the latest image, with is version 16.11.1 .

src/Testcontainers.Gitlab/GitlabBuilder.cs
Peet-HD: I've fixed the WaitStrategy so that the previous strategy will not be overwritten.

src/Testcontainers.Gitlab/GitlabContainer.cs
Peet-HD: Yeah I checked last night if there is some easier method to provide a root password and there a sereval ways. You can set the root password with an environment variable named `GITLAB_ROOT_PASSWORD` ([see here for dokumentation](https://docs.gitlab.com/ee/administration/environment_variables.html)). Another way would be to provide a file named `root_password.txt` in the root directory. This is documented [here](https://docs.gitlab.com/ee/install/docker.html#install-gitlab-using-docker-swarm-mode).

I would strongly prefer to set the root password via an environment variable, would either way would be valid. I try to find some time this week to resolve this. 

If you have an preferred implementation than just let me know.

Peet-HD: Should be fixed now. I've selected the approach via an environment variable just to keep it simple.

PR #1162, opened 2024-04-25
feat: Include the container name in the reuse hash

### LINKED ISSUE #1161: [Enhancement]: Make the container name part of the reuse hash
### Problem

Since the reuse feature became available I started using Testcontainers to automate setting up a development database. I have created a [hosted service](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services#ihostedservice-interface) that goes like this:

```csharp
using System;
using System.Threading;
using System.Threading.Tasks;
using Microsoft.EntityFrameworkCore;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;
using Testcontainers.MsSql;

namespace SampleCode;

internal class DockerDatabaseService : IHostedService
{
    private readonly IDbContextFactory<MyDbContext> _dbContextFactory;
    private readonly MsSqlContainer _container;

    public DockerDatabaseService(IHostEnvironment environment, IDbContextFactory<MyDbContext> dbContextFactory, ILoggerFactory loggerFactory)
    {
        if (!environment.IsDevelopment())
        {
            throw new NotSupportedException($"{nameof(DockerDatabaseService)} must only be used in the development environment but the current environment is \"{environment.EnvironmentName}\"");
        }

        _dbContextFactory = dbContextFactory;
        _container = new MsSqlBuilder()
            .WithLogger(loggerFactory.CreateLogger("Testcontainers"))
            .WithReuse(true)
            .WithName(environment.ApplicationName)
            .WithPortBinding(hostPort: 1433, containerPort: 1433)
            .Build();
    }

    public async Task StartAsync(CancellationToken cancellationToken)
    {
        await _container.StartAsync(cancellationToken);

        await using var dbContext = await _dbContextFactory.CreateDbContextAsync(cancellationToken);
        await dbContext.Database.MigrateAsync(cancellationToken);
    }

    public async Task StopAsync(CancellationToken cancellationToken)
    {
        await _container.StopAsync(cancellationToken);
    }
}
```

The `Testcontainers.MsSql` package is included conditionally for debug builds only:
```xml
<ItemGroup Condition="$(Configuration) == 'Debug'">
  <PackageReference Include="Testcontainers.MsSql" Version="3.8.0" />
</ItemGroup>
<ItemGroup Condition="$(Configuration) != 'Debug'">
  <Compile Remove="DockerDatabaseService.cs" />
</ItemGroup>
```

And the hosted service is also registered conditionally for debug builds in the development environment:

```csharp
#if DEBUG
    if (environment.IsDevelopment())
    {
        services.AddHostedService<DockerDatabaseService>();
    }
#endif
```

It worked so well that I used it in several projects. But then the _same_ container was reused across _different_ projects. Since I explicitly set a container name (`WithName(environment.ApplicationName)`) I assumed that the reuse mechanism would use it but it does not.

### Solution

I think the container name (just like network and volume names) should be part of the reuse hash.

### Benefit

Container reuse would work out of the box by just setting an explicit container name.

### Alternatives

Currently I'm working around this issue by configuring the container with a `reuse-id` label:

```csharp
_container = new MsSqlBuilder()
    .WithLogger(loggerFactory.CreateLogger("Testcontainers"))
    .WithReuse(true)
    .WithLabel("reuse-id", environment.ApplicationName) // 👈 goes into the reuse hash
    .WithName(environment.ApplicationName)
    .WithPortBinding(hostPort: 1433, containerPort: 1433)
    .Build();
```

Yes: #1162

0xced:
## What does this PR do?

This pull request removes the `[JsonIgnore]` attribute from the `ContainerConfiguration.Name` property so that the container name becomes part of the reuse hash.

## Why is it important?

The rationale behind this change is explained in #1161.

> [!IMPORTANT]  
> Changing the parts used in the reuse hash will break existing reused containers but the best time to do it is now, while the reuse feature is still at the experimental stage.

## Related issues

Closes #1161

## How to test this PR

A new unit test (`ContainersWithDifferentNamesShouldHaveDifferentHashes`) has been added.

### DIFF
--- a/tests/Testcontainers.Platform.Linux.Tests/ReusableResourceTest.cs
+++ b/tests/Testcontainers.Platform.Linux.Tests/ReusableResourceTest.cs
@@ -76,6 +76,15 @@ public sealed class ReusableResourceTest : IAsyncLifetime, IDisposable
         Assert.Single(response.Volumes);
     }
 
+    [Fact]
+    public void ContainersWithDifferentNamesShouldHaveDifferentHashes()
+    {
+        var hash1 = new ReuseHashContainerBuilder().WithName("Name1").GetReuseHash();
+        var hash2 = new ReuseHashContainerBuilder().WithName("Name2").GetReuseHash();
+
+        Assert.NotEqual(hash1, hash2);
+    }
+
     public static class UnsupportedBuilderConfigurationTest
     {
         private const string EnabledCleanUpExceptionMessage = "Reuse cannot be used in conjunction with WithCleanUp(true). (Parameter 'Reuse')";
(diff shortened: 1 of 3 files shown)


tests/Testcontainers.Platform.Linux.Tests/ReusableResourceTest.cs
```diff
@@ -143,4 +152,30 @@ public void EnabledCleanUpThrowsException()
             }
         }
     }
+
+    private sealed class ReuseHashContainerBuilder : ContainerBuilder<ReuseHashContainerBuilder, DockerContainer, ContainerConfiguration>
```
HofmeisterAn, line 156: Cool idea 🏅.

HofmeisterAn (commented):
I do not remember, but I thought this was on purpose. However, I think excluding it does not make much sense. IIRC when we initially discussed the feature, we were considering hashing the container response from the Docker Engine API (which contains random names). Nevertheless, LGTM. Could you please update the documentation accordingly?

https://github.com/testcontainers/testcontainers-dotnet/blob/f6c93d31531379ad06beab409ea92aa31dc408af/docs/api/resource_reuse.md?plain=1#L20-L29

0xced:
> Could you please update the documentation accordingly?

Of course! I just pushed d715a180a8e02a2b0cd056256c876f7947781f1e which updates the documentation.

HofmeisterAn (approved):

PR #1164, opened 2024-04-26
chore: Prevent ReusableResourceTest from leaking resources

### LINKED ISSUE #1157: [Bug]: ReusableResourceTest leaks docker resources
### Testcontainers version

3.8.0

Yes

### Host OS

Any

### Host arch

Any

### .NET version

8.0.204

### Docker version

```shell
Not relevant
```

### Docker info

```shell
Not relevant
```

### What happened?

Running the `ShouldReuseExistingResource` test in the `ReusableResourceTest` creates a reusable container, a reusable network and a reusable volume. They are properly disposed in the `DisposeAsync` method but since these three resources are configured as reusable they are not deleted.

This is not an issue when running on disposable virtual machines for continuous integration scenarios. This is problematic when running on our development machines where those resources with random GUIDs are piling up, though. The accumulation of networks is particularly problematic since it will eventually lead to this error: [could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network](https://stackoverflow.com/questions/43720339/docker-error-could-not-find-an-available-non-overlapping-ipv4-address-pool-am), rendering Docker unusable until running `docker network prune` manually.

Both `DockerNetwork` and `DockerVolume` have a public `DeleteAsync` method but those classes are internal. The `DockerContainer` class is public but has no `DeleteAsync` method so I'm not sure what's the best way to fix this issue.

Besides the issue about the test I had already wondered how a reusable Docker resource could be deleted programatically using the Testcontainers API.

### Relevant log output

_No response_

### Additional information

_No response_

0xced:
Fixes #1157

### DIFF
--- a/tests/Testcontainers.Platform.Linux.Tests/ReusableResourceTest.cs
+++ b/tests/Testcontainers.Platform.Linux.Tests/ReusableResourceTest.cs
@@ -83,13 +83,18 @@ public sealed class ReusableResourceTest : IAsyncLifetime, IDisposable
         Assert.Single(response.Volumes);
     }
 
-    [Fact]
-    public void ContainersWithDifferentNamesShouldHaveDifferentHashes()
+    public static class ReuseHash
     {
-        var hash1 = new ReuseHashContainerBuilder().WithName("Name1").GetReuseHash();
-        var hash2 = new ReuseHashContainerBuilder().WithName("Name2").GetReuseHash();
-
-        Assert.NotEqual(hash1, hash2);
+        public sealed class NotEqual
+        {
+            [Fact]
+            public void ForDifferentNames()
+            {
+                var hash1 = new ReuseHashContainerBuilder().WithName("Name1").GetReuseHash();
+                var hash2 = new ReuseHashContainerBuilder().WithName("Name2").GetReuseHash();
+                Assert.NotEqual(hash1, hash2);
+            }
+        }
     }
 
     public static class UnsupportedBuilderConfigurationTest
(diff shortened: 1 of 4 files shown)


tests/Testcontainers.Platform.Linux.Tests/ReusableResourceTest.cs
```diff
@@ -51,6 +51,9 @@ await Task.WhenAll(container.StartAsync(), network.CreateAsync(), volume.CreateA
 
     public Task DisposeAsync()
     {
+        // We don't want to leak resources, but .WithReuse(true).WithCleanUp(true) is not supported.
+        // So we cheat by setting a non-empty SessionId in order DisposeAsyncCore to delete the container, network and volume.
+        new List<int> { 0, 1, 2 }.ForEach(i => _disposables[i].AsDynamic()._configuration.SessionId = ResourceReaper.DefaultSessionId);
         return Task.WhenAll(_disposables.Select(disposable => disposable.DisposeAsync().AsTask()));
```
HofmeisterAn:
```suggestion
        return Task.WhenAll(_disposables.Select(disposable =>
        {
            // We do not want to leak resources, but `WithCleanUp(true)` cannot be used
            // alongside `WithReuse(true)`. As a workaround, we set the `SessionId` using
            // reflection afterward to delete the container, network, and volume.
            disposable.AsDynamic()._configuration.SessionId = ResourceReaper.DefaultSessionId;
            return disposable.DisposeAsync().AsTask();
        }));
```

HofmeisterAn (commented):
Thanks, good idea 💡. I think we can slightly improve setting the `SessionId`. Happy to merge it afterward.

HofmeisterAn (approved):

PR #1183, opened 2024-05-22
feat: Add Azure EventHubs module

workedbeforepush:
## What does this PR do?

Microsoft released an emulator for EventHubs yesterday, see https://github.com/Azure/azure-service-bus/issues/223#issuecomment-2123141525

This PR included a module for the emulator.

## Info

You need to provide an Azurite (emulator) instance/endpoint to start the EventHub emulator. This can be done using the `EventHubsConfiguration`.

You need to specify the following configuration file when starting the emulator.

```json
{
  "UserConfig": {
    "NamespaceConfig": [
      {
        "Type": "EventHub",
        "Name": "emulatorNs1",
        "Entities": [
          {
            "Name": "eh1",
            "PartitionCount": "2",
            "ConsumerGroups": [
              {
                "Name": "cg1"
              }
            ]
          }
        ]
      }
    ], 
    "LoggingConfig": {
      "Type": "File"
    }
  }
}
```

It will be dynamically mapped and i have built a floating builder so that the user can set the configuration as easily as possible.
The namespace "emulatorNs1" is mandatory and it is the only one currently supported by the emulator.

```cs
var configurationBuilder = ConfigurationBuilder
    .Create()
    .WithEventHub(EventHubName, "2", new[] { EventHubConsumerGroupName });

var builder = new EventHubsBuilder()
    .WithNetwork(_network)
    .WithConfigurationBuilder(configurationBuilder)
    .WithAzuriteBlobEndpoint(AzuriteNetworkAlias)
    .WithAzuriteTableEndpoint(AzuriteNetworkAlias);

_eventHubsContainer = builder.Build();

await _eventHubsContainer.StartAsync();
```

### DIFF
--- a/src/Testcontainers.EventHubs/Configuration/ConfigurationBuilder.cs
+++ b/src/Testcontainers.EventHubs/Configuration/ConfigurationBuilder.cs
@@ -1,0 +1,53 @@
+﻿namespace Testcontainers.EventHubs.Configuration
+{
+    public class ConfigurationBuilder
+    {
+        private const string DefaultNamespace = "emulatorns1";
+        
+        private readonly RootConfiguration _rootConfiguration = new RootConfiguration();
+            
+        private ConfigurationBuilder()
+        {
+            _rootConfiguration.UserConfig = new UserConfig
+            {
+                NamespaceConfig = new List<NamespaceConfig>()
+                {
+                    new NamespaceConfig
+                    {
+                        Type = "EventHub",
+                        Name = DefaultNamespace
+                    }
+                },
+                LoggingConfig = new LoggingConfig() { Type = "File" }
+            };
+        }
+
+        public static ConfigurationBuilder Create()
+        {
+            return new ConfigurationBuilder();
+        }
+        
+        public ConfigurationBuilder WithEventHub(string entityName, string partitionCount, IEnumerable<string> consumerGroups)
+        {
+            var namespaceConfig = _rootConfiguration.UserConfig.NamespaceConfig.FirstOrDefault(x => x.Name == DefaultNamespace);
+            if (namespaceConfig == null)
+            {
+                throw new InvalidOperationException($"Default Namespace '{DefaultNamespace}' not found.");
+            }
+
+            namespaceConfig.Entities.Add(new Entity
+            {
+                Name = entityName,
+                PartitionCount = partitionCount,
+                ConsumerGroups = consumerGroups.Select(consumerGroupName => new ConsumerGroup { Name = consumerGroupName }).ToList()
+            });
+
+            return this;
+        }
+
+        public string Build()
+        {
+            return JsonSerializer.Serialize(_rootConfiguration);
+        }
+    }
+}
--- a/src/Testcontainers.EventHubs/EventHubsBuilder.cs
+++ b/src/Testcontainers.EventHubs/EventHubsBuilder.cs
@@ -1,0 +1,102 @@
+namespace Testcontainers.EventHubs;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class EventHubsBuilder : ContainerBuilder<EventHubsBuilder, EventHubsContainer, EventHubsConfiguration>
+{
+    public const string EventHubsImage = "mcr.microsoft.com/azure-messaging/eventhubs-emulator:latest";
+
+    public const ushort EventHubsPort = 5672;
+    
+    /// <summary>
+    /// Initializes a new instance of the <see cref="EventHubsBuilder" /> class.
+    /// </summary>
+    public EventHubsBuilder()
+        : this(new EventHubsConfiguration())
+    {
+        DockerResourceConfiguration = Init().DockerResourceConfiguration;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="EventHubsBuilder" /> class.
+    /// </summary>
+    /// <param name="resourceConfiguration">The Docker resource configuration.</param>
+    private EventHubsBuilder(EventHubsConfiguration resourceConfiguration)
+        : base(resourceConfiguration)
+    {
+        DockerResourceConfiguration = resourceConfiguration;
+    }
+
+    /// <inheritdoc />
+    protected override EventHubsConfiguration DockerResourceConfiguration { get; }
+    
+    /// <summary>
+    /// Sets the event hub configuration
+    /// </summary>
+    /// <param name="configurationBuilder"></param>
+    /// <returns></returns>
+    public EventHubsBuilder WithConfigurationBuilder(ConfigurationBuilder configurationBuilder)
+    {
+        var configBytes = Encoding.UTF8.GetBytes(configurationBuilder.Build());
+        
+        return WithResourceMapping(configBytes, "Eventhubs_Emulator/ConfigFiles/Config.json");
+    }
+    
+    /// <summary>
+    /// Sets the endpoint of the azurite blob service
+    /// </summary>
+    /// <param name="azuriteBlobEndpoint"></param>
+    /// <returns></returns>
+    public EventHubsBuilder WithAzuriteBlobEndpoint(string azuriteBlobEndpoint)
+    {
+        return WithEnvironment("BLOB_SERVER", azuriteBlobEndpoint);
+    }
+    
+    /// <summary>
+    /// Sets the endpoint of the azurite table service
+    /// </summary>
+    /// <param name="azuriteTableEndpoint"></param>
+    /// <returns></returns>
+    public EventHubsBuilder WithAzuriteTableEndpoint(string azuriteTableEndpoint)
+    {
+        return WithEnvironment("METADATA_SERVER", azuriteTableEndpoint);
+    }
+    
+    /// <inheritdoc />
+    public override EventHubsContainer Build()
+    {
+        Validate();
+        
+        var waitStrategy = Wait.ForUnixContainer().UntilMessageIsLogged("Emulator Service is Successfully Up!");
+        
+        var eventHubsBuilder = DockerResourceConfiguration.WaitStrategies.Count() > 1 ? this : WithWaitStrategy(waitStrategy);
+        return new EventHubsContainer(eventHubsBuilder.DockerResourceConfiguration);
+    }
+
+    /// <inheritdoc />
+    protected override EventHubsBuilder Init()
+    {
+        return base.Init()
+            .WithImage(EventHubsImage)
+            .WithEnvironment("ACCEPT_EULA", "Y")
+            .WithPortBinding(EventHubsPort, true);
+    }
+
+    /// <inheritdoc />
+    protected override EventHubsBuilder Clone(IResourceConfiguration<CreateContainerParameters> resourceConfiguration)
+    {
+        return Merge(DockerResourceConfiguration, new EventHubsConfiguration(resourceConfiguration));
+    }
+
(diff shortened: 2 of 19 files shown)


tests/Testcontainers.EventHubs.Tests/EventHubsContainerTest.cs
```diff
@@ -0,0 +1,71 @@
...
+    private const string EventHubName = "testeventhub";
+    private const string EventHubConsumerGroupName = "testconsumergroup";
+    
+    private EventHubsContainerTest()
+    {
+        _azuriteContainer = new AzuriteBuilder()
+            .WithNetwork(_network)
+            .WithNetworkAliases(AzuriteNetworkAlias)
+            .Build();
+    }
```
eddumelendez, line 23: This is an idea still but due to EventHubs emulator will not work without Azurite then we can provide two ways to run EventHub Container implementation.

1. Embed Azurite, so, users don't worry about it.
2. Allow external Azurite in case users need a specific configuration

/cc @kiview WDYT?

eddumelendez:
Also, please add documentation for the module.

tests/Testcontainers.EventHubs.Tests/EventHubsContainerTest.cs
scrocquesel-ml150: Does a pattern exist already ?

Saglodha:
Hi @WakaToa / @eddumelendez / @kiview , 

I am from Azure Event Hubs Product team and was looking at this PR. I do realize that there were some plausible solutions offered above to get this integration setup. 

Could you please let us know if we could extend any help from our side to integrate Event Hubs emulator with testcontainers framework? We'll be happy to provide any product related assistance here.

tests/Testcontainers.EventHubs.Tests/EventHubsContainerTest.cs
eddumelendez: recently, service bus was added with the same pattern. See https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers.ServiceBus/ServiceBusBuilder.cs#L114

src/Testcontainers.EventHubs/EventHubsBuilder.cs
```diff
@@ -0,0 +1,127 @@
...
+        return Merge(DockerResourceConfiguration, new EventHubsConfiguration(azuriteTableEndpoint: azuriteTableEndpoint))
+            .WithEnvironment("METADATA_SERVER", azuriteTableEndpoint);
+    }
+    
+    /// <inheritdoc />
+    public override EventHubsContainer Build()
+    {
+        Validate();
+        
+        var waitStrategy = Wait.ForUnixContainer().UntilMessageIsLogged("Emulator Service is Successfully Up!");
```
HofmeisterAn, line 75: The wait strategy should be moved to `Init()`.

src/Testcontainers.EventHubs/Configuration/ConfigurationBuilder.cs
```diff
@@ -0,0 +1,53 @@
+﻿namespace Testcontainers.EventHubs.Configuration
```
HofmeisterAn, line 1: Please change all to file scoped namespaces.

src/Testcontainers.EventHubs/EventHubsBuilder.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// </summary>
+    /// <param name="configurationBuilder"></param>
+    /// <returns></returns>
+    public EventHubsBuilder WithConfigurationBuilder(ConfigurationBuilder configurationBuilder)
+    {
+        var configBytes = Encoding.UTF8.GetBytes(configurationBuilder.Build());
+        
+        return Merge(DockerResourceConfiguration, new EventHubsConfiguration(configurationBuilder: configurationBuilder))
+            .WithResourceMapping(configBytes, "Eventhubs_Emulator/ConfigFiles/Config.json");
+    }
```
HofmeisterAn, line 46: Does this builder cover all Event Hubs configuration? Are we responsible for it? Is there a default configuration? The `EventHubsBuilder` will fail if the user does not call `WithConfigurationBuilder(ConfigurationBuilder)` explicitly. Until now, no other module requires an additional explicit call to a builder method (expect license agreements). `_ = new FooBuilder.Build()` always returns a working configuration. This is something I favor.

src/Testcontainers.EventHubs/EventHubsBuilder.cs
```diff
@@ -0,0 +1,127 @@
...
+            .NotNull()
+            .NotEmpty();
+    }
+    
+    /// <inheritdoc />
+    protected override EventHubsBuilder Init()
+    {
+        return base.Init()
+            .WithImage(EventHubsImage)
+            .WithEnvironment("ACCEPT_EULA", "Y")
```
HofmeisterAn, line 106: Please do not hide license agreements. Right now, we use the following [pattern](https://github.com/testcontainers/testcontainers-dotnet/blob/60f50b6f608a5f2c3eebfbcd1f84a2a693f20a89/src/Testcontainers.ServiceBus/ServiceBusBuilder.cs#L43-L55).

src/Testcontainers.EventHubs/EventHubsConfiguration.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        ConfigurationBuilder = BuildConfiguration.Combine(oldValue.ConfigurationBuilder, newValue.ConfigurationBuilder);
+        AzuriteBlobEndpoint = BuildConfiguration.Combine(oldValue.AzuriteBlobEndpoint, newValue.AzuriteBlobEndpoint);
+        AzuriteTableEndpoint = BuildConfiguration.Combine(oldValue.AzuriteTableEndpoint, newValue.AzuriteTableEndpoint);
+    }
+    
+    /// <summary>
+    /// Gets the configuration builder
+    /// </summary>
+    public ConfigurationBuilder ConfigurationBuilder { get; }
```
HofmeisterAn, line 69: It looks like we do not use the `ConfigurationBuilder` sometime later. In that case, we do not need to store it in the `EventHubsConfiguration`.

src/Testcontainers.EventHubs/Testcontainers.EventHubs.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>net6.0;net8.0;netstandard2.0;netstandard2.1</TargetFrameworks>
```
HofmeisterAn, line 3:
```suggestion
        <TargetFrameworks>net8.0;net9.0;netstandard2.0;netstandard2.1</TargetFrameworks>
```

We need to merge `main` and add the test project to the [CI](https://github.com/testcontainers/testcontainers-dotnet/blob/60f50b6f608a5f2c3eebfbcd1f84a2a693f20a89/.github/workflows/cicd.yml#L34-L84) pipeline.

src/Testcontainers.EventHubs/EventHubsBuilder.cs
```diff
@@ -0,0 +1,127 @@
+using DotNet.Testcontainers;
```
HofmeisterAn, line 1: Please use the global usings file.

tests/Testcontainers.EventHubs.Tests/EventHubsContainerTest.cs
```diff
@@ -0,0 +1,71 @@
+namespace Testcontainers.EventHubs;
+
+public abstract class EventHubsContainerTest : IAsyncLifetime
+{
+    private readonly AzuriteContainer _azuriteContainer;
+
+    private EventHubsContainer _eventHubsContainer;
+
+    private readonly INetwork _network = new NetworkBuilder().WithName(NetworkName).Build();
```
HofmeisterAn, line 9: Please do not use static network names. This may clash with already existing resources and may prevent tests from running in parallel. See our [best practices](https://dotnet.testcontainers.org/api/best_practices/).

tests/Testcontainers.EventHubs.Tests/EventHubsContainerTest.cs
HofmeisterAn: I think it makes sense to align with the pattern Service Bus is using and then follow up with something that allows us to use an existing Azurite instance. Please also see this issue: https://github.com/testcontainers/testcontainers-dotnet/issues/1323. This is something we need to figure out and support in general.

tests/Testcontainers.EventHubs.Tests/Testcontainers.EventHubs.Tests.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>net8.0</TargetFrameworks>
```
HofmeisterAn, line 3:
```suggestion
        <TargetFrameworks>net9.0</TargetFrameworks>
```

tests/Testcontainers.EventHubs.Tests/Testcontainers.EventHubs.Tests.csproj
```diff
@@ -0,0 +1,22 @@
...
+        <PackageReference Include="xunit"/>
+    </ItemGroup>
+    <ItemGroup>
+        <ProjectReference Include="../../src/Testcontainers.Azurite/Testcontainers.Azurite.csproj"/>
+        <ProjectReference Include="../Testcontainers.Commons/Testcontainers.Commons.csproj"/>
+        <ProjectReference Include="..\..\src\Testcontainers.EventHubs\Testcontainers.EventHubs.csproj" />
+    </ItemGroup>
+    <ItemGroup>
+      <PackageVersion Update="Azure.Messaging.EventHubs" Version="5.11.3" />
+    </ItemGroup>
```
HofmeisterAn, line 21:
```suggestion
```

Already included above.

HofmeisterAn (changes requested):
Thanks for the PR. There are a few things we need to address before we can merge the PR.

src/Testcontainers.EventHubs/EventHubsBuilder.cs
rafek1241: With configuration builder is needed, because we don't have a default eventhub assigned to the default namespace. We can eventually introduce something, but I don't advise to do so as usually we want to have something custom. For an example, we have azure function:
```C#
public static class ProcessSalesData
{
    [FunctionName("ProcessSalesData")]
    public static void Run(
        [EventHubTrigger("SalesDataHub", Connection = "EventHubConnectionAppSetting", ConsumerGroup = "RealTimeAnalytics")] string[] events,
        ILogger log)
    {
        foreach (var eventData in events)
        {
            log.LogInformation($"C# Event Hub trigger function processed an event: {eventData}");
        }
    }
}
``` 

Our eventhub is `SalesDataHub` and usually in integration tests we won't enforce to change it to the ones that is created by the testcontainer.eventhubs. We'd rather to replicate solution that we have in our infra with custom event hub names and just automatically subscribe to this when new message came up.

rafek1241: On a second thought we can introduce `eh1` by default. Always we will be able to expand that even further, but for most of the time maybe that will be enought (for most scenarios).

rafek1241:
@HofmeisterAn

I see that @WakaToa is not so responsive so I've made my fork on a top of his changes and implement everything you suggested & aligned to the latest changes. I will create new PR for that (as I dont' know how to merge it) and we can abandon this, I think.

PR: #1342

PR #1195, opened 2024-06-12
chore: Remove PulsarContainer.CreateAuthenticationTokenAsync(TimeSpan) default arg

entvex:
## What does this PR do?

This update modifies the CreateAuthenticationTokenAsync to accept an expiry time, including an option for infinite expiration. In addition, the PulsarContainerTest and Pulsar documentation have been updated accordingly.

## Why is it important?

It’s essential for end users to have clarity about the lifespan of a token. By requiring them to set an expiry time, any uncertainty is eliminated. Additionally, we now offer support for tokens that never expire.

## How to test this PR

This is covered by the updated tests


HofmeisterAn (approved):
Thanks 💪.

PR #1196, opened 2024-06-12
feat: Add MongoDB replica set support

### LINKED ISSUE #1154: [Enhancement]: Add one node replica set support
### Problem

There is no one-node replica set support for testcontainers-dotnet but the java version seems to have something. However, by looking at the java code I have some doubts whether that's properly initializing the replica set.

### Solution

Add a `WithReplicaSet(string replicaSetName = "rs0")` or similar builder method.
Add another method to initialize the replica set.
Add another method to obtain the ReplicaSet connection string which should be something similar to `mongodb://{hostname}:{mapped_port}/?replicaSet=rs0`

### Benefit

It would allow testing mongoDb replica set and things such as change streams, only available on replica sets. Also, it would get closer to the Java version in functionality.

### Alternatives

Currently, mongoDb containers with one node replica set are possible to be created with `Ductus.FluentDocker`.
Also with testcontainers, but the random port is a challenge.

Yes

artiomchi:
## What does this PR do?

Adding an option to initialise MongoDb as a single node replica set.

To do this we need a keyfile generated with limited permissions, and add a wait strategy which will initialise the replica set on startup

## Why is it important?

Some application would expect the Mongo server to be running in a replica set to improve resiliency, and would set read/write strategies on their commands, which will fail if a replica set is not initialised.

## Related issues

There is already an open discussion about it, with some suggestions but no working solution. This PR provides a working implementation and closes #1154 

## How to test this PR

The single node replica set should behave mostly the same as a single mongo instance.

When running in a replica set, you can also call `rs` commands to get the status of the replica set.

e.g. the `rs.status()` command will display the status of the replica set.


artiomchi:
The last test run should have been green, but somehow the container has crashed/stopped? Could someone try and re-run the last test run to see if it was an issue in the GitHub runner?

artiomchi:
@HofmeisterAn I managed to re-run the tests- looks like an intermittent issue in GitHub actions. The tests are green now, and I've been using this code in my integration tests already

Let me know if anything else is needed to have this merged

HofmeisterAn:
Thank you for the pull request 🙏. I will try to test and review it in the coming days. I am aware there are a couple of module pull requests waiting to be reviewed. I have not forgotten them; I am addressing them one by one, prioritizing potential bugs and features that benefit the base library.

artiomchi:
No rush, thank you very much @HofmeisterAn!

artiomchi:
Hi, it's been almost two months since the PR was created. Is there anything we can do to help speed up the process to approve and merge this?

src/Testcontainers.MongoDb/MongoDbBuilder.cs
```diff
@@ -69,6 +76,22 @@ public override MongoDbContainer Build()
...
+                .WithCommand(DockerResourceConfiguration.Command.Concat(["--replSet", DockerResourceConfiguration.ReplicaSetName, "--keyFile", "/tmp/keyfile", "--bind_ip_all"]).ToArray())
+                .WithResourceMapping(Encoding.Default.GetBytes("""
+                    #!/bin/bash
+                    openssl rand -base64 32 > "/tmp/keyfile"
+                    chmod 600 /tmp/keyfile
+                    chown 999:999 /tmp/keyfile
+                    """.Replace("\r", "")), "/docker-entrypoint-initdb.d/01-init-keyfile.sh", UnixFileModes.OtherRead | UnixFileModes.OtherExecute)
+                .WithWaitStrategy(Wait.ForUnixContainer().UntilCommandIsCompleted(
+                    "mongosh -u $MONGO_INITDB_ROOT_USERNAME -p $MONGO_INITDB_ROOT_PASSWORD --quiet --eval " +
+                    $"\"try {{ rs.status().ok }} catch (e) {{ rs.initiate({{'_id':'{DockerResourceConfiguration.ReplicaSetName}', members: [{{'_id':1, 'host':'127.0.0.1:27017'}}]}}).ok }}\""));
```
HofmeisterAn: Can we move these lines into `WithReplicaSet(string)` (then you won't need to keep `ReplicaSetName` in the configuration class)? QQ: Can we initialize replica set with a CLI command? Then wen can move the configuration to [WithStartupCallback](https://github.com/testcontainers/testcontainers-dotnet/blob/1a647621c7b9ed01b801efdddaac37206afc6b6e/src/Testcontainers.Kafka/KafkaBuilder.cs#L68) and deceid their if we need to initialize it or not.

HofmeisterAn (commented):
Sorry for the late response. Work has been very busy over the last few days, and I just returned from parental leave. I'm trying to address the PRs, but it will take some time. The implementation looks good; I just have a question about whether it makes sense to move the initialization to `WithReplicaSet(string)`. Apologies again.

src/Testcontainers.MongoDb/MongoDbBuilder.cs
artiomchi: I didn't want to put them into `WithReplicaSet(string)`, since they're changing some other properties, like the command, to add additional parameters. If it was set in that method, and subsequent calls to change them would break the replica set setup, but it wouldn't be obvious why. Setting them here means this should correctly configure the replica set, regardless of other config values that are set.

i.e. calling the following should still work correctly
```
new MongoDbBuilder().WithReplicaSet().WithCommand("--some-flag")
```

artiomchi: Is there any benefit to moving the replica init call into the startup callback instead of a wait strategy? In either case, this should run before the container is marked as ready, but after the container has started, and the wait strategy seemed to match the exact requirements for what we do - finish the configuration of the mongodb server before we allow connections in. Happy to change it if the startup callback makes more sense

HofmeisterAn: > I didn't want to put them into `WithReplicaSet(string)`, since they're changing some other properties, like the command, to add additional parameters. [...] Setting them here means this should correctly configure the replica set, regardless of other config values that are set.

This is a general issue. Modules come pre-configured and are opinionated. Developers can always modify the builder configuration in ways that break the setup. It's a balancing act between providing enough flexibility and ensuring reliability.

I believe no other module contains this kind of logic within the `Build()` method. Until now, we've only configured and selected the appropriate wait strategy in that method IIRC.

I would like to [suggest](https://github.com/testcontainers/testcontainers-dotnet/commit/a33964aabbd39a13583338ff5334938c397b8fd7) a few changes to keep this consistent with the other modules. Honestly, I prefer moving the `else` block from `Build()` to `WithReplicaSet(string)` too, and only keep the adjustment of the wait strategy there, as we do in [other modules](https://github.com/testcontainers/testcontainers-dotnet/blob/87184e3e53b270ac352cdf59af6c2334488c7fd1/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#L85). WDYT?

The PR looks good so far. Thanks a lot for your contribution already.

artiomchi: Hey Andre, thanks for your comments and suggestions!
I've taken again at the library internals, and realised I misunderstood how `WithCommand(string[])` works, which is why I took my initial approach.

I moved most of the RS config to `WithReplicaSet(string)`, as you initially suggested.

The only thing I've left is the init wait strategy, and I left it in `Build()`, similar to the other module you linked. I also added it at the end of it, after the other strategies are added, since we need to make sure this strategy runs, even if other custom strategies are added, otherwise the replica set won't be initialised.

Hopefully this looks much better than the initial version. Let me know what you think

HofmeisterAn (approved):
Thank you for your patience, the PR, and the contribution. This is truly a great improvement 🚀 that many developers have requested — thank you!

Zephyris94:
Hi, I am using it now via
_mongoDbContainer = new MongoDbBuilder()
    .WithReplicaSet()
    .WithImage("mongo:latest")
    .WithUsername("")
    .WithPassword("")
    .WithPortBinding(_mongoPort, true)
    .WithWaitStrategy(Wait.ForUnixContainer()
       .UntilPortIsAvailable(_mongoPort))
    .Build();
    

The port is: private const int _mongoPort = 27017;

 Because I've added transactions to my project, but running it this way makes mongodb fail on startup
 the overall issue is
 Message: 
System.TimeoutException : A timeout occurred after 30000ms selecting a server using CompositeServerSelector{ Selectors = WritableServerSelector, LatencyLimitingServerSelector{ AllowedLatencyRange = 00:00:00.0150000 }, OperationsCountServerSelector }. Client view of cluster state is { ClusterId : "2", Type : "ReplicaSet", State : "Connected", Servers : [{ ServerId: "{ ClusterId : 2, EndPoint : "127.0.0.1:52699" }", EndPoint: "127.0.0.1:52699", ReasonChanged: "Heartbeat", State: "Connected", ServerVersion: 7.0.0, TopologyVersion: { "processId" : ObjectId("66d8951c02314d40f9304eac"), "counter" : NumberLong(0) }, Type: "ReplicaSetGhost", WireVersionRange: "[0, 21]", LastHeartbeatTimestamp: "2024-09-04T17:13:22.0736177Z", LastUpdateTimestamp: "2024-09-04T17:13:22.0736185Z" }] }.

Version="3.10.0"

Here is a file with full log from start to end. Sadly I am not good enough with mongodb/docker to say what's going wrong with it, I can just state that if I remove the WithReplicaSet() - tests that don't do transactions will go green, and everything works as a charm with this config I stated above
[mongodblog.txt](https://github.com/user-attachments/files/16874505/mongodblog.txt)

UPD: played around a bit, it didn't work with mongo:4.4 so eventually I ended up with 
_mongoDbContainer = new MongoDbBuilder()
    .WithImage("mongo:latest")
    .WithUsername("")
    .WithPassword("")
    .WithPortBinding(_mongoPort, true)
    .WithCommand("mongod", "--replSet", "rs0", "--bind_ip_all")
    .WithWaitStrategy(Wait.ForUnixContainer()
       .UntilPortIsAvailable(_mongoPort))
    .Build();
   and
await _mongoDbContainer.StartAsync();

await _mongoDbContainer.ExecScriptAsync("rs.initiate();");

This worked, and all tests worked (both with and without transactions inside them)

HofmeisterAn:
Modules are pre-configured. If you are using them, it is not necessary to override the default configurations, such as wait strategies, etc. Doing so can even be a disadvantage and may break the module. Instead, use:

```csharp
new MongoDbBuilder().WithReplicaSet().Build();
```

Override modules only if you are confident in what you are doing.

Zephyris94:
> it is not necessary to override the default configurations

Didn't know that. Actually this helped and worked, thank you very much, looks much cleaner

PR #1225, opened 2024-08-05
Create GarnetContainerTest.cs

Meir017:
## What does this PR do?

Add redis tests using Microsoft's Garnet docker image https://microsoft.github.io/garnet/docs/welcome/releases#docker

## Why is it important?

Demonstrating how the garnet docker-image works for redis

## Related issues

-


HofmeisterAn (commented):
Thanks for the PR. I'll close this in favor of: #1641.

The Redis module should be compatible with the container image. I'd like to reduce the number of copy-paste modules and the associated maintenance work. If Garnet introduces features or configuration options that require dedicated support in the future, we can add a separate module at that time.

PR #1213, opened 2024-07-15
fix: Add Keycloak health port 9000 starting from major version 25

paulomorgado:
## What does this PR do?

Starting with tag `25.0` Keycloak uses port 9000 instead of port 8080 for health checks.

## Why is it important?

Currently the readiness for image `25.0` and above is not correct and might consider the container ready when it's yet ready.

## Related issues

- Closes #1208

## How to test this PR

Use tag 24.0 and tag 25.0 and asset that all correctly wait for the container to be ready.

PR #1231, opened 2024-08-11
feat: Improve parsing of Docker image tag version `MatchVersion(Predicate<Version>)`

0xced:
## What does this PR do?

This pull request handles major-only version number, e.g. `2` without any dot that `Version.TryParse` doesn't parse. Matching a version was recently introduced in #1213 but did not account for major-only versions which are quite common in Docker image tags.

## Why is it important?

Verifying a version on the major part of the tag version is sometimes enough.

## Related issues

Relates #1213

## How to test this PR

The `MatchVersion_ReturnsTrue_WhenVersionMatchesPredicate` test was updated from a `Fact` to a `Theory` handling many different version number inputs.

### DIFF
--- a/src/Testcontainers/Images/DockerImage.cs
+++ b/src/Testcontainers/Images/DockerImage.cs
@@ -1,6 +1,7 @@
 namespace DotNet.Testcontainers.Images
 {
   using System;
+  using System.Globalization;
   using System.Linq;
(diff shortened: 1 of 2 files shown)


HofmeisterAn (approved):
Thanks 💪.

PR #1232, opened 2024-08-12
chore: Apply fix for github.com/apache/pulsar/issues/22811 only for affected versions

entvex:
## What does this PR do?

This commit changes the default Pulsar image from version 3.2.3 to 3.0.6. Additionally, it updates the logic to specifically handle image version v3.2.0-3 and 3.3.0 due to known issues in certain versions. As it is fixed in 3.2.4 and 3.3.1. Please see [https://pulsar.apache.org/release-notes/](https://pulsar.apache.org/release-notes/)

## Why is it important?

Right now the fix is applied to [the latest tag](https://github.com/testcontainers/testcontainers-dotnet/blob/1a647621c7b9ed01b801efdddaac37206afc6b6e/src/Testcontainers.Pulsar/PulsarContainer.cs#L66). Which can cause issues. If the user decides to use the latest tag or use an Apache Pulsar version with the issue resolved.

## How to test this PR

Ran all the tests


src/Testcontainers.Pulsar/PulsarBuilder.cs
```diff
@@ -4,7 +4,7 @@ namespace Testcontainers.Pulsar;
 [PublicAPI]
 public sealed class PulsarBuilder : ContainerBuilder<PulsarBuilder, PulsarContainer, PulsarConfiguration>
 {
-    public const string PulsarImage = "apachepulsar/pulsar:3.2.3";
+    public const string PulsarImage = "apachepulsar/pulsar:3.3.1";
```
HofmeisterAn: Typically, we avoid updating the image version of modules to prevent developers from encountering potential breaking. However, in this case, I think we can make an exception.

src/Testcontainers.Pulsar/PulsarContainer.cs
```diff
@@ -63,9 +63,9 @@ public async Task<string> CreateAuthenticationTokenAsync(TimeSpan expiryTime, Ca
         {
             int secondsToMilliseconds;
 
-            if (_configuration.Image.Tag.StartsWith("3.2") || _configuration.Image.Tag.StartsWith("latest"))
+            if (_configuration.Image.Tag.Equals("3.2.0") || _configuration.Image.Tag.Equals("3.2.1") || _configuration.Image.Tag.Equals("3.2.2") || _configuration.Image.Tag.Equals("3.2.3") || _configuration.Image.Tag.Equals("3.3.0") )
```
HofmeisterAn: FYI: We now have a new [API](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/tests/Testcontainers.Tests/Unit/Images/TestcontainersImageTest.cs#L111-L119) available to check the image version.

src/Testcontainers.Pulsar/PulsarBuilder.cs
entvex: If you would like to avoid updating the image version so often. How do you feel about setting the target version to a LTS version [release-policy](https://pulsar.apache.org/contribute/release-policy/) ?

src/Testcontainers.Pulsar/PulsarContainer.cs
entvex: Very nice! I'll adopt that!

src/Testcontainers.Pulsar/PulsarBuilder.cs
HofmeisterAn: Yes, LTS sounds good. Since there doesn't seem to be a 3.0 tag, should we go with the latest patch, 3.0.6?

entvex: Yes 👍 !

entvex:
I have made the changes and if it all looks good to you. Then it is ready to merge 😄

HofmeisterAn (changes requested):
Thanks for the PR. Could you please remove the unnecessary changes (such as whitespaces) to keep the PR changes to a bare minimum?

entvex:
> Thanks for the PR. Could you please remove the unnecessary changes (such as whitespaces) to keep the PR changes to a bare minimum?

Sure i'll try 👍

entvex:
I have removed the unnecessary white spaces.
Sorry it took so long @HofmeisterAn

HofmeisterAn (commented):
Thanks for the PR and adjustments.

PR #930, opened 2023-06-22
fix: Use the actual Docker endpoint to extract the socket path for the Resource Reaper

HofmeisterAn:
## What does this PR do?

Instead of relying on the automatically discovered Docker endpoint to retrieve the Docker socket, the pull request utilizes the actual Docker endpoint. If the auto-discovery mechanism fails to detect a Docker endpoint and the user manually configures the Docker endpoint using the builder API, Testcontainers will encounter a null reference exception during runtime.

## Why is it important?

The fix is significant because it allows the support of test environments where the auto-discovery mechanism cannot detect a Docker endpoint (although this is not recommended).

## Related issues

- Closes #929

## How to test this PR

@paule96 Would it be possible for you to test the fix and see if it resolves the issue? It is a bit challenging to set up a test in CI due to the required remote Docker endpoint, etc.

PR #1235, opened 2024-08-19
feat: Get Docker endpoint from Docker context

0xced:
## What does this PR do?

This pull request enables seamless integration with alternative Docker engines such as [Colima][1] or [OrbStack][2].

These alternative engines make use of the [Docker contexts][3] feature to set themselves as the current Docker context.

## Why is it important?

Without this change it requires a [specific configuration][4] that's far from obvious to figure out. It took me several hours fighting `NullReferenceException` because I'd only set the `DOCKER_HOST` environment variable without setting `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE`. The ryuk container would fail to start causing the `NullReferenceException`.

## Related issues

- Relates #930

## How to test this PR

A new test has been added (`DockerProcessTest.GetCurrentEndpoint`) but installing [Colima][1] or [OrbStack][2] and actually using it with Testcontainers is the one true way to test this pull request. I did it with both Colima and OrbStack and it works on my machine.

[1]: https://github.com/abiosoft/colima
[2]: https://orbstack.dev
[3]: https://docs.docker.com/engine/manage-resources/contexts/
[4]: https://github.com/testcontainers/testcontainers-java/issues/5034#issuecomment-1036433226


kiview:
Hi @0xced,
this looks to me, like this is relying on a Docker CLI to inspect the context. We normally don't integrate with Docker (or other container runtimes) through the Docker CLI. There is the possibility to read the context directly from the file system (since I assume Docker.DotNet won't support context resolution), I remember @eddumelendez looking into this in the past.

0xced:
I've address this concern in the follow-up commits by reading the configuration from the file system, like `DockerRegistryAuthenticationProvider` does.

I re-used the code that performs `docker context inspect --format {{.Endpoints.docker.Host}}` but in `DockerConfigTest.GetCurrentEndpoint()` to ensure that the implementation that reads the file system matches what the `docker inspect` command returns.

tests/Testcontainers.Tests/Unit/Builders/DockerConfigTest.cs
```diff
@@ -0,0 +1,39 @@
...
+        FileName = "docker",
+        Arguments = "context inspect --format {{.Endpoints.docker.Host}}",
+        RedirectStandardOutput = true,
+        UseShellExecute = false,
+      };
+      docker.Start();
+      docker.WaitForExit(2000);
+      var endpoint = docker.StandardOutput.ReadToEnd().Trim();
+      return new Uri(endpoint);
+    }
```
HofmeisterAn: Probably we can reuse the [DockerCLI](https://github.com/testcontainers/testcontainers-dotnet/blob/87184e3e53b270ac352cdf59af6c2334488c7fd1/tests/Testcontainers.Commons/DockerCli.cs#L44-L54) class here (gettin the host value).

0xced: Absolutely! Done in 512784ffcb43f257757039f6c31d5230899666e1.

0xced:
I think I finally got an implementation that matches the `docker context inspect --format {{.Endpoints.docker.Host}}` command invocation, that was quite some work. 😓

Fortunately the Docker CLI source code is [well commented](https://github.com/docker/cli/blob/v25.0.0/cli/command/cli.go#L364-L390)!

I don't like the many `catch { return null; }` parts but we wouldn't want to throw at that stage where we're trying to figure out the default endpoint to use. 🤷‍♀️

HofmeisterAn:
Hey, very cool pull request 😎, thanks! I've been thinking about supporting this feature for a while, so it's awesome to see it.

I added `GetDockerContext()` to the `ICustomConfiguration` interface to keep the implementation consistent when accessing configurations. I plan to add more documentation in the next few days, especially for [https://dotnet.testcontainers.org/](https://dotnet.testcontainers.org/), since it's lacking attention at the moment.

After that, we can go ahead and merge the PR. Thank you once more.

HofmeisterAn:
> I don't like the many `catch { return null; }` parts but we wouldn't want to throw at that stage where we're trying to figure out the default endpoint to use. 🤷‍♀️

I removed them. Let's see how many issues pop up. I think I've covered most error paths (except invalid JSON). I'm not a huge fan of swallowing errors because it makes it hard to help and provide support on Slack, etc. when we don't know the actual configuration and the potential underlying issue. It would be nice to have a logger available, but the auto-discovery mechanism runs too early.

0xced:
Please don't merge yet, I have found issues with the tests (the DockerConfig class is not well testable because of its static path properties). If you have a custom context some tests will fail locally but they will pass on CI as the current context will be the default one.

src/Testcontainers/Builders/DockerDesktopEndpointAuthenticationProvider.cs
```diff
@@ -15,7 +15,7 @@ internal sealed class DockerDesktopEndpointAuthenticationProvider : RootlessUnix
     /// Initializes a new instance of the <see cref="DockerDesktopEndpointAuthenticationProvider" /> class.
     /// </summary>
     public DockerDesktopEndpointAuthenticationProvider()
-      : base(GetSocketPathFromHomeDesktopDir(), GetSocketPathFromHomeRunDir())
+      : base(DockerConfig.Instance.GetCurrentEndpoint()?.AbsolutePath, GetSocketPathFromHomeDesktopDir(), GetSocketPathFromHomeRunDir())
```
HofmeisterAn, line 18: Why do we use the Docker Desktop provider here? This prevents users from overriding the `DOCKER_HOST` configuration (which is likely irrelevant) but it also prevents them from setting the `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` configuration. IIRC, this setup was necessary for Colima. On the test host, the socket path is `$COLIMA_HOME/default/docker.sock`, but in the VM that runs the containers it is a different path. Is it always `/var/run/docker.sock` (I do not remember anymore)?

HofmeisterAn:
> Please don't merge yet, I have found issues with the tests (the DockerConfig class is not well testable because of its static path properties). If you have a custom context some tests will fail locally but they will pass on CI as the current context will be the default one.

We can overload the ctor to also accept the default `.docker` and `meta` base directory paths. This would allow us to eliminate the `GetDockerConfigFile()` method too and read the path directly from `customConfigurations`.

HofmeisterAn:
Ready for final review/merge?

0xced:
I just ran the test one more time on my machine and found a way where a test could fail. I just pushed a fix (266ee03d60125ba509e0a92093c9b90d0a54048e) for the test. Hope you don't mind bringing in the `SkippableFact` dependency.

So yes, ready for final review/merge.

HofmeisterAn (approved):
💪

PR #1237, opened 2024-08-20
feat: Add Db2 module

kevin0x90:
## What does this PR do?

Adds a module to use Db2 with testcontainers.

## How to test this PR

Automated tests have been added to the project that verify startup and database connectivity.


kevin0x90:
TODO: use code includes like

```
<!--codeinclude-->
[Usage Example](../../tests/Testcontainers.Elasticsearch.Tests/ElasticsearchContainerTest.cs) inside_block:UseElasticsearchContainer
<!--/codeinclude-->
```

in documentation

src/Testcontainers.Db2/Db2Builder.cs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /// <inheritdoc />
+    protected override Db2Builder Init() => base.Init()
+        .WithImage(Db2Image)
+        .WithPortBinding(Db2Port, true)
+        .WithDatabase(DefaultDatabase)
+        .WithUsername(DefaultUsername)
+        .WithPassword(DefaultPassword)
+        .WithLicenseAgreement()
```
HofmeisterAn: Can we please use the same pattern as we do in the ServiceBus module: [1)](https://github.com/testcontainers/testcontainers-dotnet/blob/888da44936457c0d3dea0d149e4fb61f23544b33/src/Testcontainers.ServiceBus/ServiceBusBuilder.cs#L43-L55), [2)](https://github.com/testcontainers/testcontainers-dotnet/blob/888da44936457c0d3dea0d149e4fb61f23544b33/src/Testcontainers.ServiceBus/ServiceBusBuilder.cs#L64-L76)?

src/Testcontainers.Db2/Db2Builder.cs
```diff
@@ -0,0 +1,202 @@
...
+    /// <summary>
+    /// Sets the Db2 autoconfig setting.
+    /// </summary>
+    /// <param name="autoConfig">The Db2 autoconfig setting.</param>
+    /// <returns>A configured instance of <see cref="Db2Builder" />.</returns>
+    public Db2Builder WithAutoconfig(bool autoConfig)
+    {
+        return Merge(DockerResourceConfiguration, new Db2Configuration(autoConfig: autoConfig))
+            .WithEnvironment("AUTOCONFIG", autoConfig.ToString());
+    }
```
HofmeisterAn: Do we need to provide an API for these configurations? Usually, modules only include what's necessary to run the container; we do not provide an API for every single configuration. Developers can always configure the environment variables themselves (if they are optional).

src/Testcontainers.Db2/Db2Builder.cs
```diff
@@ -0,0 +1,202 @@
...
+        .WithImage(Db2Image)
+        .WithPortBinding(Db2Port, true)
+        .WithDatabase(DefaultDatabase)
+        .WithUsername(DefaultUsername)
+        .WithPassword(DefaultPassword)
+        .WithLicenseAgreement()
+        .WithArchiveLogs(false)
+        .WithAutoconfig(false)
+        .WithInMemoryDatabase()
+        .WithPrivileged(true);
```
HofmeisterAn: Does it need to run in privileged mode?

src/Testcontainers.Db2/Db2Builder.cs
```diff
@@ -0,0 +1,202 @@
...
+            .WithEnvironment("LICENSE", DefaultLicenseAgreement);
+    }
+
+    /// <summary>
+    /// Maps the database to memory.
+    /// </summary>
+    /// <returns>A configured instance of <see cref="Db2Builder" />.</returns>
+    public Db2Builder WithInMemoryDatabase()
+    {
+        return Merge(DockerResourceConfiguration, new Db2Configuration(licenseAgreement: DefaultInMemoryDatabasePath))
```
HofmeisterAn: This line passes the value to the wrong argument.

src/Testcontainers.Db2/Db2Configuration.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// <summary>
+    /// License agreement value.
+    /// </summary>
+    public string LicenseAgreement { get; }
+
+    /// <summary>
+    /// Path to the database files that should be mapped into memory (tmpfs).
+    /// </summary>
+    public string InMemoryDatabasePath { get; }
```
HofmeisterAn: We only include the properties in the configuration type that we will need later in the container type.

src/Testcontainers.Db2/Db2Builder.cs
```diff
@@ -0,0 +1,202 @@
+namespace Testcontainers.Db2;
+
+/// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+[PublicAPI]
+public sealed class Db2Builder : ContainerBuilder<Db2Builder, Db2Container, Db2Configuration>
+{
+    public const string Db2Image = "icr.io/db2_community/db2:latest";
```
HofmeisterAn: Please pin the version.

tests/Testcontainers.Db2.Tests/Testcontainers.Db2.Tests.targets
```diff
@@ -0,0 +1,22 @@
...
+        <XmlPoke
+            XmlInputPath="linux.runsettings"
+            Value="$(MSBuildProjectDirectory)/$(OutputPath)clidriver/bin"
+            Query="/RunSettings/RunConfiguration/EnvironmentVariables/PATH"
+        />
+        <XmlPoke
+            XmlInputPath="linux.runsettings"
+            Value="$(MSBuildProjectDirectory)/$(OutputPath)clidriver"
+            Query="/RunSettings/RunConfiguration/EnvironmentVariables/DB2_CLI_DRIVER_INSTALL_PATH"
+        />
```
HofmeisterAn, line 20: I moved these parts to a separate targets file. Can't we set the values directly in the `runsettings` file? (I'm not sure if the actual base path is `$(MSBuildProjectDirectory)/$(OutputPath)` already; and we can simply use the relative path).

HofmeisterAn (changes requested):
Thank you for the PR and your contribution! I apologize for the delayed review and feedback. I just hadn't found the time earlier. I have a few minor change requests (or questions), but nothing major. I'm happy to merge the PR once those are addressed 🙏.

src/Testcontainers.Db2/Db2Builder.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        // By default, the base builder waits until the container is running. However, for Db2, a more advanced waiting strategy is necessary
+        // If the user does not provide a custom waiting strategy, append the default Db2 waiting strategy.
+        var db2Builder = DockerResourceConfiguration.WaitStrategies.Count() > 1
+            ? this
+            : WithWaitStrategy(Wait.ForUnixContainer()
+                .UntilMessageIsLogged("All databases are now active")
+                .UntilMessageIsLogged("Setup has completed.")
+                .AddCustomWaitStrategy(new WaitUntil(DockerResourceConfiguration))
+            );
```
HofmeisterAn: Please move the wait strategy configuration to `Init()`. We do not need the pattern here. We only need it if we need to configure the wait strategy with developer-provided configurations.

src/Testcontainers.Db2/Db2Builder.cs
kevin0x90: Will adjust

src/Testcontainers.Db2/Db2Builder.cs
kevin0x90: I thought it makes these settings more discoverable for users.

src/Testcontainers.Db2/Db2Builder.cs
kevin0x90: Sadly yes according to the documentation
https://www.ibm.com/docs/en/db2/12.1?topic=system-linux

src/Testcontainers.Db2/Db2Builder.cs
kevin0x90: Will adjust

src/Testcontainers.Db2/Db2Builder.cs
kevin0x90: Will adjust

tests/Testcontainers.Db2.Tests/Testcontainers.Db2.Tests.targets
kevin0x90: I’m also not sure but to me it looked like the environment variables need to have an absolute path that’s why I implemented it in this way in order to use variables from MSBuild for the absolute path construction.

src/Testcontainers.Db2/Db2Builder.cs
HofmeisterAn: From our experience providing APIs (except when necessary to start the container), for configurations that can simply be set through environment variables have several disadvantages:

- Depending on the number of configurations, providing APIs for all of them requires significant effort.
- Distinguishing between different image versions and their respective configurations is difficult.
- We need to include and maintain documentation.
- Keeping the APIs and documentations up to date.

Testcontainers' modules are opinionated, pre-configured, and follow best practice configurations. A module does not need to support all possible use cases. If developers require a specific configuration, they can always use the generic builder (`ContainerBuilder()`).

If certain configurations are useful for development and testing, we should set them by default, as we do for [PostgreSQL](https://github.com/testcontainers/testcontainers-dotnet/blob/2c6c915e84f03d239a94e26b81f867f03f6a0942/src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs#L92-L95). If the APIs are not necessary, I recommend removing them.

src/Testcontainers.Db2/Db2Builder.cs
kevin0x90: I have just seen that I also have to adjust this part. Will fix it in the next few days.

Testcontainers.sln
```diff
@@ -214,13 +218,13 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Testcontainers.Xunit.Tests", "tests\Testcontainers.Xunit.Tests\Testcontainers.Xunit.Tests.csproj", "{E901DF14-6F05-4FC2-825A-3055FAD33561}"
 EndProject
 Global
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
 		Release|Any CPU = Release|Any CPU
 	EndGlobalSection
```
HofmeisterAn:
```suggestion
	GlobalSection(SolutionConfigurationPlatforms) = preSolution
		Debug|Any CPU = Debug|Any CPU
		Release|Any CPU = Release|Any CPU
	EndGlobalSection
	GlobalSection(SolutionProperties) = preSolution
		HideSolutionNode = FALSE
	EndGlobalSection
```

HofmeisterAn (commented):
Thanks for updating the PR. I've updated the remaining parts. The PR looks good! If you're okay with it, I'm happy to merge it.

I just have one final question: Does it make sense to set any of these configurations as defaults (in `Init()`) to improve startup performance (or improves development experience)?

```csharp
.WithEnvironment("TO_CREATE_SAMPLEDB", "false")
.WithEnvironment("PERSISTENT_HOME", "true")
.WithEnvironment("REPODB", "false")
.WithEnvironment("BLU", "false")
.WithEnvironment("HADR_ENABLED", "false")
.WithEnvironment("ARCHIVE_LOGS", "false")
.WithEnvironment("AUTOCONFIG", "false")
```

HofmeisterAn (approved):

kevin0x90:
> Thanks for updating the PR. I've updated the remaining parts. The PR looks good! If you're okay with it, I'm happy to merge it.
> 
> I just have one final question: Does it make sense to set any of these configurations as defaults (in `Init()`) to improve startup performance (or improves development experience)?
> 
> ```cs
> .WithEnvironment("TO_CREATE_SAMPLEDB", "false")
> .WithEnvironment("PERSISTENT_HOME", "true")
> .WithEnvironment("REPODB", "false")
> .WithEnvironment("BLU", "false")
> .WithEnvironment("HADR_ENABLED", "false")
> .WithEnvironment("ARCHIVE_LOGS", "false")
> .WithEnvironment("AUTOCONFIG", "false")
> ```

Thank you very much for finishing up the PR for the parts I had not addressed yet. Yes it would have some positive effect regarding startup performance but in general the official image takes quite some time to startup even with the optimizations so I’m not sure if it makes a big difference.

PR #1257, opened 2024-09-09
Bump Azurite to 3.31.0

boekabart:
## What does this PR do?

Bump to the latest (June 2024) release of Azurite image.

## Why is it important?

Our renovatebot is bumping the Azure.Storage.Blobs package to 12.21.2, which is incompatible with the 3.28.0 azurite. Keeping Azure client packages up-to-date is kind of important to not get far behind with the risk of losing support for your API version in Azure.

## Related issues

-

## How to test this PR

Regression test; using latest Azure (eg Storage.Blobs) client package.

## Follow-ups

Any way we can set up smth like renovatebot on this project itself, to keep packages up-to-date more automatically?

### DIFF
--- a/src/Testcontainers.Azurite/AzuriteBuilder.cs
+++ b/src/Testcontainers.Azurite/AzuriteBuilder.cs
@@ -4,7 +4,7 @@ namespace Testcontainers.Azurite;
 [PublicAPI]
 public sealed class AzuriteBuilder : ContainerBuilder<AzuriteBuilder, AzuriteContainer, AzuriteConfiguration>
 {
-    public const string AzuriteImage = "mcr.microsoft.com/azure-storage/azurite:3.28.0";
+    public const string AzuriteImage = "mcr.microsoft.com/azure-storage/azurite:3.31.0";
 
     public const ushort BlobPort = 10000;
 


HofmeisterAn:
Thanks for the pull request. We try not to update module versions unless it is absolutely necessary (due to bugs, etc.). This avoids breaking changes for others. Our [Best Practices](https://dotnet.testcontainers.org/api/best_practices/) suggest overriding and pinning the version instead. It also ensures that the chosen client dependency (NuGet) stays compatible with the selected image version. Appreciate your understanding. Let me know if you have any questions.

boekabart:
I appreciate the policy and its good intentions. But at the moment the end result is that anyone that uses Testcontainers.Azurite with the latest Azure client package (at least this particular one) will get this error. Of course, this is a result of the particular ecosystem (hosted cloud) where it's fine for MS's nuget packages to not be 'backward compatible' with Older Versions Of Azure (as there should not be any older versions of Azure out there), but it is a realistic scenario that the Testcontainers project will have to deal with.

Also, in this particular (and likely other ones like AWS) scenario, I think you serve your users best by *actively* upgrading this package, since The Real Thing will just as actively be upgraded by Microsoft. So, those 'breaking changes' you want to prevent, actually may be a disservice to your users in this case, since their Actual Deployment is also changing without their control...

(Having said that - thanks for the reminder that I can override the version, that fixes my immediate issue).

PR #1258, opened 2024-09-09
feat: Change visibility of `AccountName` and `AccountKey` to public, move to `AzuriteBuilder`

### LINKED ISSUE #1238: [Enhancement]: In AzuriteContainer, is there a reason AccountName and AccountKey are not public?
### Problem

The visiblity of `AccountName` and `AccountKey` are private in `AzuriteContainer`. This makes it difficult to construct a custom connection string if your AzuriteContainer is on a network. `MsSqlBuilder` makes all relevant info public so we can easily build connection strings.

### Solution

I would like to move `AccountName` and `AccountKey` to `AzuriteBuilder` to make it consistent with the `MsSqlBuilder`, and change the visibility to public.

Furthermore, would there be value in letting the consumer customize these values?

### Benefit

Better consistency across modules.

### Alternatives

Right now if I was to build a custom connection string, I'd have to hard code the AccountName and AccountKey. This isn't terrible, but it also shouldn't be necessary.

Yes

mikecole:
## What does this PR do?

The visiblity of AccountName and AccountKey are private in AzuriteContainer. This makes it difficult to construct a custom connection string if your AzuriteContainer is on a network. 

## Why is it important?

This helps the consumer to construct a connection string.

## Related issues

- Closes #1238 

## How to test this PR

As this widened the visibility of two variables, risk is low. The solution compiled and all tests ran.

### DIFF
--- a/src/Testcontainers.Azurite/AzuriteBuilder.cs
+++ b/src/Testcontainers.Azurite/AzuriteBuilder.cs
@@ -12,6 +12,10 @@ public sealed class AzuriteBuilder : ContainerBuilder<AzuriteBuilder, AzuriteCon
 
     public const ushort TablePort = 10002;
 
+    public const string AccountName = "devstoreaccount1";
+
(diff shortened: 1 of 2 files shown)


HofmeisterAn (approved):
Thanks 🙌

Issue #1264, opened 2024-09-19
[Bug]: MSSQL containers is not running on Ubuntu 22.04.5


JoseDevUFC:
### Testcontainers version

3.10.0

### Using the latest Testcontainers version?

Yes

### Host OS

Ubuntu 22.04      OS Version: 22.04.5 LTS     Kernel Version: 6.8.0-1014-azure     Image Version: 20240915.1.0     Systemd version: 249.11-0ubuntu3.12

### Host arch

x86

### .NET version

8.0.401

### Docker version

```shell
Docker Amazon ECR Credential Helper 0.8.0
Docker Compose v2 2.27.1
Docker-Buildx 0.17.1
Docker Client 26.1.3
Docker Server 26.1.3
```

### Docker info

```shell
docker info
  shell: /usr/bin/bash -e {0}
  env:
    DOTNET_ROOT: /usr/share/dotnet
Client: Docker Engine - Community
 Version:    26.1.3
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.17.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.27.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 15
 Server Version: 26.1.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7f7fdf5fed64eb6a7caf99b3e12efcf9d60e311c
 runc version: v1.1.14-0-g2c9f560
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-1014-azure
 Operating System: Ubuntu 22.04.5 LTS
 OSType: linux
[...]

Issue #1136, opened 2024-03-04
[README]: Contributing new modules


HofmeisterAn:
### Problem

Hi 👋, in the past few weeks, I've encountered several issues with the GitHub-hosted runners. Due to the increasing number of modules, pulling all the images occupies too much disk space. I've implemented a temporary fix to free up some disk space, but this won't be enough as we add more new modules.

Until we find a suitable fix and implement a mechanism to handle the growing number of modules and image sizes, I prefer not to merge pull requests that involve **very simple** module configurations. Which can be accomplished with just a few lines using the generic container builder API. If you believe the module is valuable for the community, please consider creating an issue first and discussing it there to avoid unnecessary work on creating a pull request. OC, I will review and try to merge the remaining and outstanding module PRs.

### Solution

\-

### Benefit

\-

### Alternatives

\-

### Would you like to help contributing this enhancement?

Yes

mrudat:
Would it be worthwhile to refactor modules so that each image or class of images (e.g. Web/Database/... Servers that might want to share a common interface) has its own repository?

That would mean that exactly one docker image should be created for each image-specific repository, presuming that the same configuration used from multiple implementations in different languages results in an identical container being tested.

We could also describe the image-specific configuration in a common configuration format and use that to generate the language-specific code.

For example, username/password/... is often configured via environment variables, and the specific environment variable names are specific to individual images. However, the code for capturing the username/password is shared between multiple images across multiple languages, which suggests we have too much nearly identical custom-maintained code and are waiting for a refactoring pass to extract the common functionality.

eddumelendez:
Hi, I think this can help to free some more space for linux workers https://github.com/jlumbroso/free-disk-space

HofmeisterAn:
Hi Eddú, we are doing this already:

https://github.com/testcontainers/testcontainers-dotnet/blob/734fb5205492fbf154588e8fd09e4dde47cd3a6e/.github/workflows/cicd.yml#L48-L64

kiview:
> Would it be worthwhile to refactor modules so that each image or class of images (e.g. Web/Database/... Servers that might want to share a common interface) has its own repository?

From past experiences with tc-java, this leads to a lot of operational overhead (e.g. releases), unless we have good automation in place.

Splitting in parallelizing the CI workflow on a per-module or module-sets level can help and it is what we essentially do in tc-java as well.

kevin0x90:
[...]

PR #1265, opened 2024-09-19
fix: Bump MSSQL image version, remove Azure SQL Edge and Papercut module

HofmeisterAn:
## What does this PR do?

With the recent VM image update (`Ubuntu-22.04`) for GitHub-hosted runners and Microsoft-hosted agents, we encounter issues where the MSSQL and Azure SQL Edge modules crash during container startup. We experienced similar issues with `Ubuntu-24.04`.

We need to update the MSSQL version to one that is compatible with the new VM image: `mcr.microsoft.com/mssql/server:2022-CU14-ubuntu-22.04`. Usually, we try to avoid updating module image versions to prevent breaking changes, and we recommend that developers pin the image version. If you encounter this issue, please use the container builder API (`WithImage(string)`) to update the image accordingly:

    WithImage("mcr.microsoft.com/mssql/server:2022-CU14-ubuntu-22.04")

Unfortunately, we also need to remove the Azure SQL Edge module and mark the NuGet dependency as deprecated. There is no new version available, and the image is [EOL](https://techcommunity.microsoft.com/t5/sql-server-blog/azure-sql-edge-update-september-2023/ba-p/3930827) anyway.

@TechLiam I decided to remove the Papercut module as well. The pipeline frequently fails when sending or receiving SMTP messages. We can add the module again if it becomes stable and someone takes responsibility for addressing the issues/module.

I am  really sorry for the inconvenience, especially since this is the second issue with the MSSQL module in such a short time. I want to stress that both issues were caused by upstream changes, and we are working to address them and provide fixes/workaround quickly.

🚩 FYI: We saw other images failing too, but these are the two we are using in Testcontainers for .NET.

## Why is it important?

Our CI pipeline keeps failing. We need to address the issue to get it running again.

## Related issues

- Relates https://github.com/actions/runner-images/issues/10646
- Relates https://github.com/actions/runner-images/issues/10649
- Relates https://github.com/testcontainers/testcontainers-dotnet/issues/1264


TechLiam:
That is fair enough on removing the Papercut module I do use it so would be happy to support and take ownership of it can you point me to the logs for the problem to take a look @HofmeisterAn 
Or is it to do with this if so then ok I understand https://github.com/testcontainers/testcontainers-dotnet/issues/1136

HofmeisterAn:
@TechLiam Initially, it failed with the following [error](https://github.com/testcontainers/testcontainers-dotnet/runs/30374716242#r34):

> System.Net.Mail.SmtpException: The operation has timed out.

I tried several changes to stabilize it, but unfortunately, nothing worked. I switched to `SendAsync`, which no longer threw the `SmtpException`, but then receiving the message timed out.

So basically, what I tried was this:

```diff
diff --git a/tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs b/tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs
index 6feba9c..5ed6add 100644
--- a/tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs
+++ b/tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs
@@ -23,17 +23,19 @@ public sealed class PapercutContainerTest : IAsyncLifetime

         Message[] messages;

+        using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
+
         using var httpClient = new HttpClient();
         httpClient.BaseAddress = new Uri(_papercutContainer.GetBaseAddress());

         using var smtpClient = new SmtpClient(_papercutContainer.Hostname, _papercutContainer.SmtpPort);

         // When
-        smtpClient.Send("from@example.com", "to@example.com", subject, "A test message");
+        smtpClient.SendAsync("from@example.com", "to@example.com", subject, "A test message", null);

         do
         {
-            var messagesJson = await httpClient.GetStringAsync("/api/messages")
+            var messagesJson = await httpClient.GetStringAsync("/api/messages", timeoutCts.Token)
                 .ConfigureAwait(true);

             var jsonDocument = JsonDocument.Parse(messagesJson);
```

The pipeline failed after 30 seconds with the following [error](https://github.com/testcontainers/testcontainers-dotnet/runs/30394857308#r34):

> System.Threading.Tasks.TaskCanceledException: The operation was canceled.

Then I reduced the HTTP client timeout to 1 second and retried getting the message, but this didn’t work either (receiving timeout). Right now, I believe there is an issue where Papercut is not receiving or processing the message correctly. Maybe the wait strategy indicates readiness too early, and we send the message too soon—I'm not sure.

TechLiam:
Issued a PR to reintroduce the module https://github.com/testcontainers/testcontainers-dotnet/pull/1268
I've added the wait strategy I think that needed for the test to not be flakey

philipnguyxn:
Encountered this issue again on Azure DevOps pipeline with a self-hosted agent image even when using the suggested `WithImage("mcr.microsoft.com/mssql/server:2022-latest")` / `WithImage("mcr.microsoft.com/mssql/server:2022-CU14-ubuntu-22.04")` workaround. Tested locally on a Mac M1 Pro didn't encounter the error. 

Image info:
```
OS: linux-x64
Dotnet SDK: ^8.0
```

kwaazaar:
This issue is not solved for me. It appears ALWAYS when I run multiple tests in parallel (run all tests from VS) on my dev-laptop. Not om my dev-desktop. On my laptop I have Docker Desktop use WSL2. On my desktop I have not. Maybe that enables to reproduce this.

PR #1266, opened 2024-09-20
feat: Update Azurite image version to 3.32.0

michalchecinski:
## What does this PR do?

This PR updates Azurite version from `3.28.0` to `3.32.0`

## Why is it important?

This change is needed to support the Azurite/Azure Storage API version 2024-11-04, which is required by `Azure.Storage.Blobs` version >= `12.20.0`

## Related issues

-


HofmeisterAn:
Hi, we don't update the image version of modules to avoid developers running into breaking changes. Instead, we [recommend](https://dotnet.testcontainers.org/api/best_practices/) pinning the image version.

PR #1268, opened 2024-09-20
feat: Reintroduce Papercut module

TechLiam:
## What does this PR do?
Added a wait strategy using HTTP request is succeeded for the health endpoint for Papercut WebUI Service.
This should stop the failing on the pipeline raised here https://github.com/testcontainers/testcontainers-dotnet/pull/1265

## Why is it important?
The wait strategy allows the container to be ready for testing and usage when starting up especially in a very busy environment

## Related issues
None

## How to test this PR
Tests pass now when running dotnet cake
Screenshot 2024-09-20 172804

## Follow-ups
Hopefully, there is enough here to stop the issues @HofmeisterAn was having when doing other modules


HofmeisterAn (commented):
Hey Liam, as I mentioned on Slack, thanks so much for the PR and all the effort you put into it, really appreciated!

I updated the Cake script to run the tests sequentially for now (I know it's slower, but I have some thoughts on that below). Could I kindly ask you to merge the `develop` branch and revert the changes related to Cake and the pipeline for now? I'd prefer to keep the PRs separate.

I've been thinking about a few things:

1. I think we should avoid pruning images after each test (assembly). I want to prevent removing **all** images on a developer’s machine when someone uses the Cake task.
    - I'd prefer using an MSBuild target (this works from a CLI and an IDE too) that developers can opt into, for example, using an environment variable. The MSBuild target could then parse log messages and prune only the images actually used by the tests.
2. I changed the pipeline to run sequentially to make reading the log messages easier, it was quite messy and difficult to follow before. My idea is to spawn multiple runners for the different modules to run the tests in parallel (in the future). We already have the implementation to merge code coverage from multiple runners, and I think this would have a few advantages: it should run much faster, be more stable, and it would eliminate the need to free up disk space, so we don’t need to worry about the size of the images.

WDYT?

TechLiam:
Hi @HofmeisterAn 
This all makes sense I have merged back develop into my branch and removed any changes that are not papercut module related hopefully, everything else is good from your point of view. Sorry, this too a while, hectic week no time for open source committing :)

HofmeisterAn (approved):
Thanks 🥳

PR #1269, opened 2024-09-22
feat: Add Neo4j Enterprise Edition support (`WithEnterpriseEdition(bool)`)

### LINKED ISSUE #1261: [Enhancement]: Support Neo4J Enterprise
### Problem

It seems that the current Neo4J Container is only built with the regular neo4j community version image in mind.

However, there is also the possibility of using a licensed enterprise edition, which offers quite a few distinct features (https://neo4j.com/pricing/) which would make a difference even in a testing context.

I know about the possibility to define a custom image with the `.WithImage(...)` method on the builder, but the enterprise version also requires an environment variable to be set, in order to work at all (`NEO4J_ACCEPT_LICENSE_AGREEMENT=yes`).

### Solution

For convenience reasons, it would be neat to have a public `.WithEnterpriseEdition(...)` method on the builder which sets both the image and the environment variable right away.

This is also the public API Java TestContainers provide (https://java.testcontainers.org/modules/databases/neo4j/#choose-your-neo4j-license)

### Benefit

Purely QoL, there are no immediate functional benefits.

### Alternatives

I assume the current alternative is to just build an extension method yourself, or to just type out the two methods on the builder

Yes

Sossenbinder:
## What does this PR do?

See https://github.com/testcontainers/testcontainers-dotnet/issues/1261

- Added small QoL method to support Neo4j enterprise
- Added validation to throw when enterprise is requested without the correct license agreement environment variable
- Added tests for enterprise & configuration changes
- Changed existing test to use IAsyncDisposable instead of sync-over-async disposal

## Why is it important?

Purely QoL, there are no immediate functional benefits.

## Related issues

- Closes #1261


HofmeisterAn (commented):
I made some minor adjustments and organized the `using` statements. Additionally, I changed the member to `WithEnterpriseEdition(bool)`. I still expect developers to agree to the license, as we cannot do that for them. I would like to add the new member to the documentation as well; then I am happy to merge it. LMKWYT. Thanks for the PR!

Sossenbinder:
> I made some minor adjustments and organized the `using` statements. Additionally, I changed the member to `WithEnterpriseEdition(bool)`. I still expect developers to agree to the license, as we cannot do that for them. I would like to add the new member to the documentation as well; then I am happy to merge it. LMKWYT. Thanks for the PR!

Sounds perfectly fine to me 👍 I'm happy to adhere to the repo standards modified from your side, so this works for me.

Thanks for applying them, this is probably something I missed while working on the very niche code section required for my changes 😄

HofmeisterAn (commented):
There are two small things I would like to mention:

- Java uses a slightly different approach to handle license agreements for images. Specifically, they use [LicenseAcceptance](https://github.com/testcontainers/testcontainers-java/blob/b1f8ada00b27e51d3f629a0beb9978e2ecd0817a/core/src/main/java/org/testcontainers/utility/LicenseAcceptance.java), which reads accepted license agreements for images from a file. Perhaps we can consider adopting a similar mechanism in the future to support the same approach across all modules.
- The current implementation has a minor flaw. If someone wants to use a different version of Neo4j (pin a version), they need to use the builder methods in the correct order. This will fall back to the default image version (not something we want):

    ```csharp
    _ = new Neo4jBuilder().WithImage("neo4j:5.23-enterprise").WithEnterpriseEdition(true).Build();
    ```
    I haven’t looked into the Java implementation, so I’m not sure if they encounter a similar issue. My initial thought is that we could change the image version only if it’s not already an enterprise version. Additionally, it might make sense to just append the `-enterprise` tag?

Sossenbinder:
> There are two small things I would like to mention:
> 
> - Java uses a slightly different approach to handle license agreements for images. Specifically, they use [LicenseAcceptance](https://github.com/testcontainers/testcontainers-java/blob/b1f8ada00b27e51d3f629a0beb9978e2ecd0817a/core/src/main/java/org/testcontainers/utility/LicenseAcceptance.java), which reads accepted license agreements for images from a file. Perhaps we can consider adopting a similar mechanism in the future to support the same approach across all modules.
> - The current implementation has a minor flaw. If someone wants to use a different version of Neo4j (pin a version), they need to use the builder methods in the correct order. This will fall back to the default image version (not something we want):
> 
>     ```csharp
>     _ = new Neo4jBuilder().WithImage("neo4j:5.23-enterprise").WithEnterpriseEdition(true).Build();
>     ```
>     I haven’t looked into the Java implementation, so I’m not sure if they encounter a similar issue. My initial thought is that we could change the image version only if it’s not already an enterprise version. Additionally, it might make sense to just append the `-enterprise` tag?

That's a fair point. I saw the Java implementation as well, but kind of skipped the file based agreement for now. It feels like that's more of a Java ecosystem style, at least it felt pretty unfamiliar from other nuget packages I used in the dotnet space, so I would agree we might want to add this for the sake of consistency. But probably not an immediate requirement.

Ah, I agree on the second point. I'm on vacation right now, so I can take this over once I'm back next week 👍

HofmeisterAn:
> Ah, I agree on the second point. I'm on vacation right now, so I can take this over once I'm back next week 👍

No rush. I might find some time to add it in the meantime. Have a nice vacation 🌞.

Sossenbinder:
Nice work! That seems pretty extensive now, I think this should do it for all cases.

HofmeisterAn:
> Nice work! That seems pretty extensive now, I think this should do it for all cases.

I had some time this afternoon to think about it, but I bet there are still a few edge cases that aren’t covered. However, it should definitely work for most cases, though. If you're okay with it, I’m happy to go ahead and merge it. Thanks for the PR!

Sossenbinder:
Sure, go ahead, that works for me 👍Thanks

HofmeisterAn (approved):

PR #1278, opened 2024-10-14
feat: Add getter for Azurite blob, queue and table endpoint

pregress:
## What does this PR do?

Exposes extra methods on azurite to get the individual endpoints for blob, queue & table.
This is useful if managed identity is used in the actual implementation.

## Why is it important?

The storage connection string parser is internal in the azure sdk. 

## Related issues
n/a

### DIFF
--- a/src/Testcontainers.Azurite/AzuriteContainer.cs
+++ b/src/Testcontainers.Azurite/AzuriteContainer.cs
@@ -23,9 +23,39 @@ public sealed class AzuriteContainer : DockerContainer
         properties.Add("DefaultEndpointsProtocol", Uri.UriSchemeHttp);
         properties.Add("AccountName", AzuriteBuilder.AccountName);
         properties.Add("AccountKey", AzuriteBuilder.AccountKey);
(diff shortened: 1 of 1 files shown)


HofmeisterAn (approved):
Thanks

joelbourbon:
> ## What does this PR do?
> [...] This is useful if managed identity is used in the actual implementation.

@pregress Curious to know how this helps with Managed Identity specifically.

pregress:
> > ## What does this PR do?
> > [...] This is useful if managed identity is used in the actual implementation.
> 
> @pregress Curious to know how this helps with Managed Identity specifically.

If you use this ctor: https://learn.microsoft.com/en-us/dotnet/api/azure.data.tables.tableserviceclient.-ctor?view=azure-dotnet#azure-data-tables-tableserviceclient-ctor(system-uri-azure-core-tokencredential-azure-data-tables-tableclientoptions)

And your environment under test uses the URI in the environment variables. You need the URI and not the connection string.

PR #1305, opened 2024-11-23
chore: Add test to ensure that all test projects are configured for CI

0xced:
## What does this PR do?

This pull request introduces a new test to ensure that all test projects are configured for Continuous Integration & Delivery.

## Why is it important?

Because it's too easy to forget to add a line in the cicd.yml after creating a new module. This new test will catch this common error.

## Related issues

This is a safety measure after the introduction of running each test project on a separate runner in #1295.

## How to test this PR

Of course if the `{ name: "Testcontainers", runs-on: "ubuntu-22.04" }` line is deleted from the `cicd.yaml` file then this new test won't run. 🙃


HofmeisterAn (approved):
Thanks

PR #1308, opened 2024-11-26
feat: Throw `DockerUnavailableException` when Docker is not available

0xced:
## What does this PR do?

This pull request introduces a new `DockerUnavailableException` which is thrown instead of `ArgumentException` which was not the appropriate exception to throw since the root of the problem is not a bad argument but the state of the system which is wrong.

The exception message has also been slightly improved.

## Why is it important?

Unlike `ArgumentException`, `DockerUnavailableException` can be caught and handled.

## Related issues

Once both this pull request and #1263 are merged, the error messages will be improved, giving a better hint to the user at what went wrong. Here are two sample messages from the new `DockerUnavailableException`.

Running Testcontainers after quitting Docker Desktop:

> Docker is either not running or misconfigured. Please ensure that Docker is available at unix:///var/run/docker.sock or unix:///Users/0xced/.docker/run/docker.sock
You can customize your configuration using either the environment variables or the ~/.testcontainers.properties file. For more information, visit:
https://dotnet.testcontainers.org/custom_configuration/

Running Testcontainers after quitting [OrbStack](https://orbstack.dev):

> Docker is either not running or misconfigured. Please ensure that Docker is available at unix:///var/run/docker.sock or unix:///Users/0xced/.orbstack/run/docker.sock
You can customize your configuration using either the environment variables or the ~/.testcontainers.properties file. For more information, visit:
https://dotnet.testcontainers.org/custom_configuration/

## How to test this PR

Quit Docker Desktop or configure a bad endpoint then try to run a container. Observe the improved exception and error message.


HofmeisterAn (commented):
Maybe we should discuss this in an issue. We can merge the PR, but I think these features would be better in Docker.DotNet. Things like detecting the container runtime, getting private registry credentials, etc. should be part of Docker.DotNet where more developers benefit from. This is one reason why we forked Docker.DotNet and use our own implementation. The Java's client has way more features and is much more developer-friendly than our .NET counterpart.

0xced:
> but I think these features would be better in Docker.DotNet

Absolutely. Could be challenging, though!

HofmeisterAn:
IIRC, we did not throw an exception on purpose because either one of the other providers may find a container runtime, or the user can set one explicitly via `WithDockerEndpoint(string)`. If we want to have a more meaningful exception, why not simply throw it in `Validate()` and be done with it?

**Edit**: This comment was intended for: https://github.com/testcontainers/testcontainers-dotnet/pull/1263. However, I now see that you are catching the exception here.

0xced:
I've read your comment several times now, both in this context and in the context of #1263 but I still can't make sense of it. 😕

I mentioned #1263 in the original description of this pull request because by merging only this pull request, the error message would read like this:
> Docker is either not running or misconfigured. Please ensure that Docker is available at unix:///var/run/docker.sock
You can customize […]

If merging both this pull request and #1263, the error message would read like this, hinting that running OrbStack is likely to solve the issue.
> Docker is either not running or misconfigured. Please ensure that Docker is available at unix:///var/run/docker.sock **or unix:///Users/0xced/.orbstack/run/docker.sock**
You can customize […]

HofmeisterAn:
> I've read your comment several times now, both in this context and in the context of #1263 but I still can't make sense of it. 😕
> 
> I mentioned #1263 in the original description of this pull request because by merging only this pull request, the error message would read like this:
> 
> > Docker is either not running or misconfigured. Please ensure that Docker is available at unix:///var/run/docker.sock
> > You can customize […]
> 
> If merging both this pull request and #1263, the error message would read like this, hinting that running OrbStack is likely to solve the issue.
> 
> > Docker is either not running or misconfigured. Please ensure that Docker is available at unix:///var/run/docker.sock **or unix:///Users/0xced/.orbstack/run/docker.sock**
> > You can customize […]

Maybe I'm mixing something up or there's a misunderstanding, sorry about that. I'm referring to this comment: https://github.com/testcontainers/testcontainers-dotnet/pull/1235#issuecomment-2299693993. IIRC, throwing an exception during auto-discovery might break the library and prevent users from specifying an endpoint later via the container builder API. It's perfectly fine if the auto-discovery mechanism doesn't find an endpoint at all. I'll take a look at it when I'm back home.

0xced:
> Throwing an exception during auto-discovery might break the library and prevent users from specifying an endpoint later via the container builder API. It's perfectly fine if the auto-discovery mechanism doesn't find an endpoint at all.

No worries, this new `DockerUnavailableException` is **not** thrown during auto-discovery. It's thrown if auto-discovery did not discover anything and `.WithDockerEndpoint()` was not explicitly called.

Enjoy your days off!

HofmeisterAn (approved):
Thanks 🙇.

PR #1315, opened 2024-12-14
feat: Add pausing and unpausing container

stefannikolei:
## What does this PR do?

Added Methods to Pause and Unpause an container

## Why is it important?

Wanted to change some tests in CoreWCF to Testcontainers. Pause and Unpause is used there. Tho I need those methods to continue the implementation

### DIFF
--- a/src/Testcontainers/Containers/DockerContainer.cs
+++ b/src/Testcontainers/Containers/DockerContainer.cs
@@ -606,11 +606,6 @@ namespace DotNet.Testcontainers.Containers
         return;
       }
 
-      if (!TestcontainersStates.Paused.Equals(State))
-      {
-        return;
-      }
-
       Unpausing?.Invoke(this, EventArgs.Empty);
 
       await _client.UnpauseAsync(_container.ID, ct)
--- a/src/Testcontainers/Containers/DockerContainer.cs
+++ b/src/Testcontainers/Containers/DockerContainer.cs
@@ -606,7 +606,7 @@ namespace DotNet.Testcontainers.Containers
     {
       ThrowIfLockNotAcquired();
 
-      if (!Exists())
+      if (State != TestcontainersStates.Paused)
       {
         return;
       }
--- a/src/Testcontainers/Containers/DockerContainer.cs
+++ b/src/Testcontainers/Containers/DockerContainer.cs
@@ -19,7 +19,7 @@ namespace DotNet.Testcontainers.Containers
   [PublicAPI]
   public class DockerContainer : Resource, IContainer
   {
-    private const TestcontainersStates ContainerHasBeenCreatedStates = TestcontainersStates.Created | TestcontainersStates.Running | TestcontainersStates.Exited;
+    private const TestcontainersStates ContainerHasBeenCreatedStates = TestcontainersStates.Created | TestcontainersStates.Running | TestcontainersStates.Paused | TestcontainersStates.Exited;
 
     private const TestcontainersHealthStatus ContainerHasHealthCheck = TestcontainersHealthStatus.Starting | TestcontainersHealthStatus.Healthy | TestcontainersHealthStatus.Unhealthy;
 
@@ -312,6 +312,7 @@ namespace DotNet.Testcontainers.Containers
         .ConfigureAwait(false);
     }
 
+    /// <inheritdoc />
     public async Task PauseAsync(CancellationToken ct = default)
     {
       using var disposable = await AcquireLockAsync(ct)
@@ -321,6 +322,7 @@ namespace DotNet.Testcontainers.Containers
         .ConfigureAwait(false);
     }
 
+    /// <inheritdoc />
     public async Task UnpauseAsync(CancellationToken ct = default)
     {
       using var disposable = await AcquireLockAsync(ct)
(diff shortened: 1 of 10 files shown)


HofmeisterAn (commented):
Thanks for the PR. Before I review it, could you please add a [test](https://github.com/testcontainers/testcontainers-dotnet/tree/develop/tests/Testcontainers.Platform.Linux.Tests)? I'm not sure how we can properly test it, but maybe it's enough to run a simple `alpine` container, pause it, and check its state?

stefannikolei:
> Thanks for the PR. Before I review it, could you please add a [test](https://github.com/testcontainers/testcontainers-dotnet/tree/develop/tests/Testcontainers.Platform.Linux.Tests)? I'm not sure how we can properly test it, but maybe it's enough to run a simple `alpine` container, pause it, and check its state?

Yeah, I was not sure either. Gonna look into a basic test. Where would you want me to put it? testContainers.Tests Unit/Containers/Unix

HofmeisterAn:
> > Thanks for the PR. Before I review it, could you please add a [test](https://github.com/testcontainers/testcontainers-dotnet/tree/develop/tests/Testcontainers.Platform.Linux.Tests)? I'm not sure how we can properly test it, but maybe it's enough to run a simple `alpine` container, pause it, and check its state?
> 
> Yeah, I was not sure either. Gonna look into a basic test. Where would you want me to put it? testContainers.Tests Unit/Containers/Unix

Just put it into the `Testcontainers.Platform.Linux.Tests` test project, thanks.

stefannikolei:
> > > Thanks for the PR. Before I review it, could you please add a [test](https://github.com/testcontainers/testcontainers-dotnet/tree/develop/tests/Testcontainers.Platform.Linux.Tests)? I'm not sure how we can properly test it, but maybe it's enough to run a simple `alpine` container, pause it, and check its state?
> 
> > 
> 
> > Yeah, I was not sure either. Gonna look into a basic test. Where would you want me to put it? testContainers.Tests Unit/Containers/Unix
> 
> 
> 
> Just put it into the `Testcontainers.Platform.Linux.Tests` test project, thanks.

Done

src/Testcontainers/Containers/DockerContainer.cs
```diff
@@ -522,6 +560,69 @@ await _client.StopAsync(_container.ID, ct)
...
+
+      if (!Exists())
+      {
+        return;
+      }
+
+      if (!TestcontainersStates.Paused.Equals(State))
+      {
+        return;
+      }
```
HofmeisterAn: **question**: Here, I am not sure if we should check if the resource exist at all, like we do in all the other methods (in the end, it probably does not matter). More relevant is the second if. Should we throw an exception or log a warning if the container was not paused before?

src/Testcontainers/Containers/DockerContainer.cs
```diff
@@ -522,6 +560,69 @@ await _client.StopAsync(_container.ID, ct)
...
+
+      Unpausing?.Invoke(this, EventArgs.Empty);
+
+      await _client.UnpauseAsync(_container.ID, ct)
+        .ConfigureAwait(false);
+
+      _container = await _client.Container.ByIdAsync(_container.ID, ct)
+        .ConfigureAwait(false);
```
HofmeisterAn, line 615: **note**: I removed the try-catch. We expect that the resource exists here; if not, an exception is thrown. I think it was copied from `UnsafeStopAsync(CancellationToken)`, which is special because the resource may no longer exist if auto-remove was configured.

HofmeisterAn (commented):
Thanks for the PR! I made some minor changes (mostly related to the documentation). If you’re OK with the changes, we can go ahead and merge the PR. Thanks again.

src/Testcontainers/Containers/DockerContainer.cs
stefannikolei: Yeah it was copied.

src/Testcontainers/Containers/DockerContainer.cs
stefannikolei: I think a log would be enough. I would expect that it does not throw when you hit play and it is already playing.

HofmeisterAn: I did not test it with the implementation, but the CLI shows an error message.

stefannikolei: Ok then it would probably best to stick to the cli behavior and throw an exception

stefannikolei: I can change the behavior. Which exception do you want me to throw?

HofmeisterAn: Let's check how the Docker Engine API behaves first. Probably, we can simply remove the if statement and forward the exception raised by Docker.DotNet?

stefannikolei: Docker.DotNet.DockerApiException
Docker API responded with status code=InternalServerError, response={"message":"Container 66c14bd6bb475d377c81fea724db4ee04ea674bf2dccd2ce2c46685ee4d92485 is not paused"}

HofmeisterAn: Then we can remove the `if` and use the Docker.DotNet exception, right? WDYT?

stefannikolei: It's ok for me. Going to change this and add a test case this evening. So we should be ready by tomorrow

HofmeisterAn (approved):
Thanks 👏

PR #1321, opened 2024-12-20
feat(Oracle): Add support for WithDatabase(string) for Oracle 18 and onwards

### LINKED ISSUE #1233: [Enhancement]: Allow Oracle container builder to set the database name
### Problem

The currently used Oracle image uses a database named `XEPDB1`. In the more recent versions of the image it is named `FREEPDB1`. The current API allows us to change the image to use with Oracle but doesn't permit to change the database name:
- `WithDatabase(string)` exists but is private
- `Merge(configuration)` exists but is protected and the class sealed

As a consequence there is no way to pick a newer version of this image, potentially with better security.

### Solution

Make `WithDatabase` public

### Benefit

Not having to request the image to be changed in TestContainers.

### Alternatives

N/A

Yes

0xced:
## What does this PR do?

This pull request selects the correct database default name depending on the Docker image version and enable setting a custom database name for Oracle 18 and onwards.

## Why is it important?

This enables using the new Oracle Docker images (e.g. [gvenzl/oracle-free](https://hub.docker.com/r/gvenzl/oracle-free)) which doesn't work with the current implementation because of the hardcoded `XEPDB1` default database.

> [!NOTE]
> The `gvenzl/oracle-free` image is the first one with arm64 support which is especially interesting for macOS users.

## Related issues

- Closes #1233
- Relates #1329

## How to test this PR

The `OracleContainerTest` class has been augmented to test many variations of the Oracle Docker images.

* gvenzl/oracle-xe:11
* gvenzl/oracle-xe:18
* gvenzl/oracle-xe:21
* gvenzl/oracle-free:23


0xced:
11 Oracle containers from 4 images and the GitHub runner fails with _No space left on device_ 😖

HofmeisterAn:
> 11 Oracle containers from 4 images and the GitHub runner fails with _No space left on device_ 😖

It looks like we are right at the edge. Although I expected ~20GB to be enough, I thought the root filesystem had something like ~30GB left. Maybe we can save some MBs by restoring the NuGet cache on the temp disk, but that likely won't be enough too.

```
gvenzl/oracle-xe:11-slim-faststart 1.64GB
gvenzl/oracle-xe:18-slim-faststart 4.44GB
gvenzl/oracle-xe:21-slim-faststart 4.66GB
gvenzl/oracle-xe:21.3.0-slim-faststart 4.66GB
gvenzl/oracle-free:23-slim-faststart 4.62GB
```

What do you think about removing `18` and `21` (as we are most likely testing the same configurations)? Maybe that is enough already.

0xced:
I tried something else that seems to work pretty well: splitting the Oracle tests into multiple projects so that we don't have the disk space issue.

Let's hope this can land into version 4.3.0. Mac users will be happy to be able to use the `gvenzl/oracle-free:23` images!

I also just realised that the *-slim images are actually slim but that the *-slim-faststart images are not slim at all, what a misleading name!

| image                                | size    |
|--------------------------------------|---------|
| gvenzl/oracle-free:23-slim           | 1.41 GB |
| gvenzl/oracle-free:23                | 2.19 GB |
| gvenzl/oracle-free:23-slim-faststart | 4.37 GB |
| gvenzl/oracle-free:23-faststart      | 5.34 GB |

So if you prefer to keep all the Oracle tests in the same project we might try the *-slim images instead of the *-slim-faststart images and we'll save quite some GBs!

HofmeisterAn:
Cool idea utilizing the compiler constants 😎 I will try to review it tomorrow.

jwedel:
> I also just realised that the *-slim images are actually slim but that the *-slim-faststart images are not slim at all, what a misleading name!
> 
> image	size
> gvenzl/oracle-free:23-slim	1.41 GB
> gvenzl/oracle-free:23	2.19 GB
> gvenzl/oracle-free:23-slim-faststart	4.37 GB
> gvenzl/oracle-free:23-faststart	5.34 GB
> So if you prefer to keep all the Oracle tests in the same project we might try the *-slim images instead of the *-slim-faststart images and we'll save quite some GBs!

The faststart images have performed the installation process already which probably will involves some unpacking, hence the bigger image.

However, they start significantly faster which is key in integration tests.

Would this non fast start only be used in your own tests or also as default in the shipped lib?

if the latter, I would suggest to keep the slim fast start.

HofmeisterAn:
> > I also just realised that the *-slim images are actually slim but that the *-slim-faststart images are not slim at all, what a misleading name!
> > image	size
> > gvenzl/oracle-free:23-slim	1.41 GB
> > gvenzl/oracle-free:23	2.19 GB
> > gvenzl/oracle-free:23-slim-faststart	4.37 GB
> > gvenzl/oracle-free:23-faststart	5.34 GB
> > So if you prefer to keep all the Oracle tests in the same project we might try the *-slim images instead of the *-slim-faststart images and we'll save quite some GBs!
> 
> The faststart images have performed the installation process already which probably will involves some unpacking, hence the bigger image.
> 
> However, they start significantly faster which is key in integration tests.
> 
> Would this non fast start only be used in your own tests or also as default in the shipped lib?
> 
> if the latter, I would suggest to keep the slim fast start.

The default image tag is `21.3.0-slim-faststart`. I believe @0xced suggested the `*-slim` tags in case we want to keep all Oracle tests within the same project and run them on a single runner. Anyway, we still use `*-slim-faststart` and recommend that developers override the default image and pin their preferred version.

src/Testcontainers.Oracle/OracleBuilder.cs
```diff
@@ -63,6 +77,18 @@ public OracleBuilder WithPassword(string password)
...
+        var defaultServiceName = GetDefaultServiceName();
+        if (DockerResourceConfiguration.Database == null)
+        {
+            return new OracleContainer(WithDatabase(defaultServiceName).DockerResourceConfiguration);
+        }
+
+        if (DockerResourceConfiguration.Database != defaultServiceName)
+        {
+            return new OracleContainer(WithEnvironment("ORACLE_DATABASE", DockerResourceConfiguration.Database).DockerResourceConfiguration);
+        }
```
HofmeisterAn, line 90: Doesn't it make sense to set the environment variable in `WithDatabase(string)`? Images that don't support it will ignore it, and we can remove the check here, right?

src/Testcontainers.Oracle/OracleBuilder.cs
```diff
@@ -83,6 +108,18 @@ protected override void Validate()
     {
         base.Validate();
 
+        const string message = "The image '{0}' does not support configuring the database. It is only supported on Oracle 18 and onwards.";
+
+        Predicate<OracleConfiguration> databaseConfigurationNotSupported = value =>
+            value.Database != null && value.Image.MatchVersion(v => v.Major < 18);
```
HofmeisterAn, line 114: Shouldn't the predicate check if images < 18 use the default service (database) name? It should only throw an exception if the service name does not match the default name, right?

HofmeisterAn (dismissed):
Thank you for the PR again. It looks good. I just have two minor questions. Let me know what you think and if we need to address them.

src/Testcontainers.Oracle/OracleBuilder.cs
0xced: Yes, good point! Addressed in e17bd73ce6a9969e64aef44223e5426222a06546.

HofmeisterAn: 😕 looks like the container does not properly start if the env var corresponds to the default value… let's revert it, and keep it as it was - sorry.

0xced: Indeed, reverted in 9af66e27732650c6a8f5f7f5991f2b16f1c9d071. I think that was the exact reason why I wrote it that way and also why I wrote a test that uses the default database name explicitly! 

But this raised another interesting issue: the container failing to start did not end up failing the wait strategy (at least on my machine, using [OrbStack](https://orbstack.dev)). I'll investigate if it's an Oracle specific issue or not.

src/Testcontainers.Oracle/OracleBuilder.cs
0xced: Technically, yes. But the documentation of the `WithDatabase` method states this:
> The database can only be set for Oracle 18 and onwards.

So an exception is to be expected (even if it would probably work with the default database name).

src/Testcontainers.Oracle/OracleBuilder.cs
0xced: > I'll investigate if it's an Oracle specific issue or not.

It's definitely not an Oracle specific issue. If for some reason a container exits without completing the readiness check then it will wait for the wait strategy timeout (which defaults to 1 hour). I will file a new issue about this.

HofmeisterAn: This likely depends on the wait strategy. For example, getting the logs from a stopped container works just fine. That's probably why we try it again and again.

src/Testcontainers.Oracle/OracleBuilder.cs
HofmeisterAn: All right! 👍 We don't need to make it more complex than it already is. Until now, I have tried to avoid providing multiple builders, but here it may make sense for the future.

HofmeisterAn (approved):
Thanks ✌️.

PR #1335, opened 2025-01-11
feat(ServiceBus): Add support to use existing MSSQL container instances

### LINKED ISSUE #1323: [Enhancement]: Allow Azure Service Bus to use an existent instance of MsSqlContainer
### Problem

The current implementation of `Testcontainers.ServiceBus` automatically sets up `MsSqlContainer`, but if running integration tests using the Azure Service Bus and SQL Server, we end up having 2 SQL Server containers + 1 Azure Service Bus emulator. This can consume significant resources in the integration tests' environment. When using `WebApplicationFactory<T>` these numbers can easily scale.

### Solution

A possible solution is to create a public overload of [ServiceBusBuilder.WithMsSqlContainer](https://github.com/testcontainers/testcontainers-dotnet/blob/develop/src/Testcontainers.ServiceBus/ServiceBusBuilder.cs#L114) that receives a `MsSqlContainer`, the password and the network alias as *optional* parameters and use `DependsOn` if `MsSqlContainer` and is not null.

Another possible solution would be to expose an API on `ServiceBusBuilder` to allow consumers to set the `MSSQL_SA_PASSWORD` and `SQL_SERVER` environment variables like the `ServiceBusBuilder.WithAcceptLicenseAgreement` is exposed. This way we can benefit from methods like `DependsOn` and wait strategies to synchronize the order in which containers should start.

The third option would be to expose the already created SQL Server from `Testcontainers.ServiceBus` and let consumers use it to test against it, but allowing them to override the SQL Server image if desired.

### Benefit

If testing against SQL Server and Azure service bus, we can have one container less on the test environment, saving resources and speeding up test execution. Consumers may use different SQL Server images, meaning one image less to pull before test execution in pipelines.

### Alternatives

Alternatively, it's possible to implement an Azure Service Bus emulator on the test environment, using `ContainerBuilder` and `IContainer` and manually bind the Azure Service Bus emulator container with an existing SQL Server container.

Yes

lgcmotta:
## What does this PR do?

This PR exposes an overload of `WithMsSqlContainer`, enabling users to attach an existing `MsSqlContainer` instance to the Azure Service Bus emulator. Previously, ServiceBusContainer automatically spun up its own SQL Server container, leading to unnecessary duplication when users were already testing with Azure Service Bus and SQL Server (see #1323).

## Why is it important?

This enhancement addresses the resource usage issues described in #1323. By allowing a single `MsSqlContainer` to be shared, the overhead of launching a separate SQL Server container to use the Azure Service Bus emulator is eliminated. This can be especially beneficial in CI/CD scenarios or other environments where minimizing resource consumption and reducing test execution time is a priority.

## Related issues

- Closes #1323 

## How to test this PR

- A new test has been added that demonstrates attaching a custom `MsSqlContainer` to `ServiceBusContainer`.
- Ensure both containers (`ServiceBusContainer` and `MsSqlContainer`) are on the same network.
- Use a non-default SQL Server image, for example: `mcr.microsoft.com/azure-sql-edge:latest`.
- Verify that only the containers you intend to run are launched (Azure SQL Edge for SQL Server, and the Azure Service Bus emulator), and that `mcr.microsoft.com/mssql/server:2022-CU14-ubuntu-22.04` (the default in `MsSqlBuilder`) is not spun up.

## Follow-ups

None at this time. Inline summary documentation is included in the new overload.


tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs
```diff
@@ -37,9 +37,18 @@ public static TheoryData<Type> GetContainerImplementations(bool expectDataProvid
...
             // TODO: If a module contains multiple container implementations, it would require all container implementations to implement the interface.
             foreach (var containerType in testAssembly.Value.Where(type => type.IsAssignableTo(typeof(IContainer))))
             {
+                var typeAssemblyName = containerType.Assembly.GetName().Name;
+
+                if (!string.IsNullOrWhiteSpace(testAssemblyName) && !string.IsNullOrWhiteSpace(typeAssemblyName) && !testAssemblyName.Contains(typeAssemblyName))
+                {
+                    continue;
+                }
+
```
lgcmotta: When adding a new test that creates an MS SQL instance, the `MsSqlContainer` type was exported under `Testcontainers.ServiceBus.Tests.dll`. This caused the test to detect `IContainer` implementations that aren’t part of the Service Bus assembly. Ideally, verifying that `MsSqlContainer` implements `IDatabaseContainer` should happen in its theory case (`Testcontainers.MsSql.Tests`), rather than in the `Testcontainers.ServiceBus.Tests.dll` theory case.

I’d appreciate any suggestions on how to handle these cross-assembly references without causing false positives.

HofmeisterAn: I'll take a look at that, kown issue: https://github.com/testcontainers/testcontainers-dotnet/blob/f1b2e0b84b4d72cfea51ee4aceb3280c5d30e31f/tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs#L40

HofmeisterAn: I believe you are encountering this issue. The following change should resolve it. I can create a PR and add it sometime in the next few days.

```diff
diff --git a/tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs b/tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs
index 0c637fe..3fa694f 100644
--- a/tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs
+++ b/tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs
@@ -40,6 +40,19 @@ public sealed class DatabaseContainersTest
             // TODO: If a module contains multiple container implementations, it would require all container implementations to implement the interface.
             foreach (var containerType in testAssembly.Value.Where(type => type.IsAssignableTo(typeof(IContainer))))
             {
+                var testAssemblyName = testAssembly.Key.GetName().Name!;
+
+                var containerTypeAssemblyName = containerType.Assembly.GetName().Name!;
+
+                // If a module utilizes another one of our modules, do not include the container type
+                // if it does not belong to the actual module. For example, the ServiceBus module
+                // utilizes the MsSql module. We do not want to include the MsSqlContainer type
+                // twice or place it in the wrong test.
+                if (!testAssemblyName.Contains(containerTypeAssemblyName, StringComparison.OrdinalIgnoreCase))
+                {
+                    continue;
+                }
+
                 var hasDataProvider = testAssembly.Value.Exists(type => type.IsSubclassOf(typeof(DbProviderFactory)));
 
                 if (expectDataProvider && hasDataProvider)

```

lgcmotta: @HofmeisterAn, thanks, I'll be waiting to sync my branch!

src/Testcontainers.ServiceBus/ServiceBusBuilder.cs
```diff
@@ -54,6 +54,28 @@ public ServiceBusBuilder WithAcceptLicenseAgreement(bool acceptLicenseAgreement)
...
+    /// <param name="msSqlContainer">An existing instance of <see cref="MsSqlContainer"/> to use as the database backend.</param>
+    /// <param name="databaseNetworkAlias">The network alias that will be used to connect to the SQL Server container.</param>
+    /// <param name="saPassword">The SA password for the SQL Server container. Defaults to <see cref="MsSqlBuilder.DefaultPassword"/>.</param>
+    /// <returns>A configured instance of <see cref="ServiceBusBuilder"/>.</returns>
+    public ServiceBusBuilder WithMsSqlContainer(
+        MsSqlContainer msSqlContainer,
+        string databaseNetworkAlias,
+        string saPassword = MsSqlBuilder.DefaultPassword)
+    {
+        return Merge(DockerResourceConfiguration, new ServiceBusConfiguration(databaseContainer: msSqlContainer))
```
rafek1241: nitpick: you have repetition with `private ServiceBusBuilder WithMsSqlContainer()` method. Introduce common private method when depends on the context you will use `.DependsOn()`.

src/Testcontainers.ServiceBus/ServiceBusContainer.cs
```diff
@@ -33,8 +33,11 @@ public string GetConnectionString()
     /// <inheritdoc />
     protected override async Task UnsafeCreateAsync(CancellationToken ct = default)
     {
-        await _configuration.Networks.Single().CreateAsync(ct)
-            .ConfigureAwait(false);
+        foreach (var network in _configuration.Networks.ToArray())
```
rafek1241: What is it for? Why you expect to have more than a single network in service bus?

tests/Testcontainers.ServiceBus.Tests/ServiceBusContainerWithCustomMsSqlTest.cs
```diff
@@ -0,0 +1,60 @@
+﻿namespace Testcontainers.ServiceBus;
+
+public class ServiceBusContainerWithCustomMsSqlTest : IAsyncLifetime
```
rafek1241: This can be made with even more efficient way unless you're making something different in the test case `public async Task ReceiveMessageReturnsSentMessage()`.

You can do that in the same way as in azurite tests, or take a look at my PR: https://github.com/testcontainers/testcontainers-dotnet/pull/1342/files#diff-25b47ff76fdbb15075d0f4194d50310896defe9e2c50d2e0a2d5d872ec5a19c1

define abstract class with test case and in inherited classes you can provide different setup for containers.

src/Testcontainers.ServiceBus/ServiceBusContainer.cs
lgcmotta: By default, if users do not pass an instance of MSSQL, the `ServiceBusContainer` will create a network and provide an instance of a `MsSqlContainer`.  But if they do provide a custom instance, they will have to manually assign a network (`INetwork`) to both `ServiceBusContainer` and the provided `MsSqlContainer`. But, since the `Init` method already assigned a network to the `DockerResourceConfiguration` at this point the container will have at least two networks. By only invoking `Single` we're are not ensuring that the user's network between MSSQL and ASB Emulator is created

src/Testcontainers.ServiceBus/ServiceBusBuilder.cs
lgcmotta: I think it can be achieved by using the `private` version of `WithMsSqlContainer` to build the MSSQL container and invoke the `public` version. `DependsOn` won't harm the `private` version

src/Testcontainers.ServiceBus/ServiceBusContainer.cs
lgcmotta: hey, @rafek1241 can we resolve this one?

src/Testcontainers.ServiceBus/ServiceBusBuilder.cs
lgcmotta: @rafek1241 done: https://github.com/testcontainers/testcontainers-dotnet/pull/1335/commits/8a5f3a0b613446a55e7eae1cc7998e439d536a9a

tests/Testcontainers.ServiceBus.Tests/ServiceBusContainerWithCustomMsSqlTest.cs
lgcmotta: done! https://github.com/testcontainers/testcontainers-dotnet/pull/1335/commits/89c3040bf4d59eb722946f3f57e87cf694eb9e33

src/Testcontainers.ServiceBus/ServiceBusBuilder.cs
lgcmotta: @rafek1241, I'm going to change it back... when invoking this we ended up having 2 MSSQL containers:

```csharp
private ServiceBusBuilder WithMsSqlContainer()
{
    var msSqlContainer = new MsSqlBuilder()
        .WithNetwork(DockerResourceConfiguration.Networks.Single())
        .WithNetworkAliases(DatabaseNetworkAlias)
        .Build();

    return WithMsSqlContainer(msSqlContainer, DatabaseNetworkAlias);
}
```
which results in having the default MSSQL and the custom on when running the tests:
image

while he approach from before works as expected:
```csharp
private ServiceBusBuilder WithMsSqlContainer()
{
    var msSqlContainer = new MsSqlBuilder()
        .WithNetwork(DockerResourceConfiguration.Networks.Single())
        .WithNetworkAliases(DatabaseNetworkAlias)
        .Build();

    return Merge(DockerResourceConfiguration, new ServiceBusConfiguration(databaseContainer: msSqlContainer))
        .WithEnvironment("MSSQL_SA_PASSWORD", MsSqlBuilder.DefaultPassword)
        .WithEnvironment("SQL_SERVER", DatabaseNetworkAlias);
}
```
only the custom MSSQL is running with ASB emulator:
image

src/Testcontainers.ServiceBus/ServiceBusContainer.cs
HofmeisterAn: I think I mentioned this in Slack or GitHub. We should change this approach: `Init()` should not assign an MSSQL instance immediately. Instead, we assign it in `Build()` if none was provided. The same applies to the network. We are already using a similar approach with the [wait strategies](https://github.com/testcontainers/testcontainers-dotnet/blob/4aa61f22186b9a686a0b59334bcbdee13cf7562e/src/Testcontainers.MongoDb/MongoDbBuilder.cs#L92-L101) (if we rely on user configurations).

rafek1241: > hey, @rafek1241 can we resolve this one?

I did additional PR review, but I don't want to approve/reject without @HofmeisterAn input as he knows this repository the best.

lgcmotta: @HofmeisterAn and @rafek1241, I think it makes sense not to assign the network or the MSSQL instance on `Intit()`. Since `ContainerConfiguration.Networks` is an `IEnumerable<INetwork>`, I assumed that assigning multiple networks to a given container instance was possible... Looking at the `ContainerBuilder.WithNetwork(INetwork)` overload, I realized that the code always creates a new collection and assigns it to the container's configuration. So in this piece of code, it's safe to use `Single()`. Users will still take care of assigning the same network to MSSQL and ASB emulator if a custom instance is desired instead of the default one...

lgcmotta: The `foreach` loop when creating the networks was reverted here: https://github.com/testcontainers/testcontainers-dotnet/pull/1335/commits/c54334cca2ef0d66a16bc6d5342760ace21eebe3

I changed the default behavior setup from `Init()` to `Build()` here: https://github.com/testcontainers/testcontainers-dotnet/pull/1335/commits/a4ef713b8223fa07444c12c7a3b5287b969cbaff.
  
By doing this, I was able to achieve success in applying @rafek1241's suggestion on https://github.com/testcontainers/testcontainers-dotnet/pull/1335#discussion_r1921143833.

Tests results:
1. The `ServiceBusBuilder` with the default `MsSqlContainer` (Ubuntu-based image) test run:
image
image

2. The `ServiceBusBuilder` with a custom `MsSqlContainer` (Azure edge image) test run:
image
image

3. Containers when both tests are running simultaneously: 
image
image

tests/Testcontainers.Databases.Tests/DatabasesContainerTest.cs
HofmeisterAn: I've pushed the changes.

lgcmotta: @HofmeisterAn got it, this branch is now updated

HofmeisterAn (commented):
Thanks for the PR. I've slightly rearranged the `Build()` method and added `INetwork` as an argument to `WithMsSqlContainer(...)`. I believe it feels more natural for developers to pass the network along with the container instance, as this way they don't have to remember to add the Service Bus Container to the same network. It's more likely to be configured correctly this way, I think. Let me know what you think. I'm happy to merge the PR if you're okay with the changes. Thanks ✌️.

lgcmotta:
> Thanks for the PR. I've slightly rearranged the `Build()` method and added `INetwork` as an argument to `WithMsSqlContainer(...)`. I believe it feels more natural for developers to pass the network along with the container instance, as this way they don't have to remember to add the Service Bus Container to the same network. It's more likely to be configured correctly this way, I think. Let me know what you think. I'm happy to merge the PR if you're okay with the changes. Thanks ✌️.

Hey @HofmeisterAn, I agree with you, it makes sense to pass the `INetwork` with the `MsSqlContainer`. You can merge these changes!! Thank you for the feedback and contribution to the PR!

HofmeisterAn (approved):

PR #1294, opened 2024-11-09
feat: Update docker image to Apache Pulsar 4.0 LTS

keshramjith:
## What does this PR do?

- Upgraded the default Apache Pulsar docker image to 4.0 LTS.
- Changed user to root, to create secret.key for auth.
- Fixed start up to use correct secret.key file.
- Remove deprecated usage of `apply-config-from-env-with-prefix.py` causing 'standalone' cluster to disappear.

## Why is it important?

To use the latest version of Apache Pulsar 4.0 LTS to allow developers to take advantage of the new features by having a working testing environment.

## Related issues

- Closes #1291

PR #1355, opened 2025-01-29
fix(Pulsar): Support Apache Pulsar 4

### LINKED ISSUE #1291: [Enhancement]: Upgrade Testcontainers.Pulsar to use Apache Pulsar v4.0.0
### Problem

Upgrade docker image of Testcontainers.Pulsar to Apache Pulsar v4.0.0

### Solution

Changing docker image version and any other necessary changes required for Apache Pulsar v4.0.0

### Benefit

Using the latest [LTS](https://pulsar.apache.org/blog/2024/10/24/announcing-apache-pulsar-4-0/) version of Apache Pulsar

### Alternatives

None

Yes

entvex:
## What does this PR do?
- Replaced deprecated script arguments with the updated ones to ensure proper configuration handling.
- Upgraded the Pulsar image to version **3.0.8**.
- Updated default file paths to align with version **4.0.X** for better compatibility and usability.

## Why is it important?
The new Apache Pulsar **4.0.X** container image enforces stricter security by reducing user privileges. As a result, the `secret.key` file needs to be created in a location that can be written to in both **3.0.X** and **4.0.X** versions.
This change ensures **backward compatibility**, which is a core principle of Testcontainers, while also taking advantage of the latest improvements in Pulsar **4.0.X**.

## How to test this PR
- Ran the test suite using Apache Pulsar broker images for both versions **3.0.8** and **4.0.2** to successfully validate compatibility across both versions.
- Tests passed in both scenarios.

### Open Question
@HofmeisterAn, would you prefer that I configure the tests to explicitly validate against both image versions (e.g., **3.0.X** and **4.0.X**), or is it sufficient to test exclusively with the latest default version (`3.0.X`)? Let me know how you'd like to proceed.

- Closes #1291
- Closes #1294

### DIFF
--- a/src/Testcontainers.Pulsar/PulsarBuilder.cs
+++ b/src/Testcontainers.Pulsar/PulsarBuilder.cs
@@ -4,7 +4,7 @@ namespace Testcontainers.Pulsar;
 [PublicAPI]
 public sealed class PulsarBuilder : ContainerBuilder<PulsarBuilder, PulsarContainer, PulsarConfiguration>
 {
-    public const string PulsarImage = "apachepulsar/pulsar:3.0.8";
+    public const string PulsarImage = "apachepulsar/pulsar:3.0.9";
 
     public const ushort PulsarBrokerDataPort = 6650;
 
--- a/src/Testcontainers.Pulsar/PulsarBuilder.cs
+++ b/src/Testcontainers.Pulsar/PulsarBuilder.cs
@@ -4,7 +4,7 @@ namespace Testcontainers.Pulsar;
 [PublicAPI]
 public sealed class PulsarBuilder : ContainerBuilder<PulsarBuilder, PulsarContainer, PulsarConfiguration>
 {
-    public const string PulsarImage = "apachepulsar/pulsar:3.0.6";
+    public const string PulsarImage = "apachepulsar/pulsar:3.0.8";
 
     public const ushort PulsarBrokerDataPort = 6650;
 
@@ -12,7 +12,7 @@ public sealed class PulsarBuilder : ContainerBuilder<PulsarBuilder, PulsarContai
 
     public const string StartupScriptFilePath = "/testcontainers.sh";
 
-    public const string SecretKeyFilePath = "/pulsar/secret.key";
+    public const string SecretKeyFilePath = "/tmp/secret.key";
 
     public const string Username = "test-user";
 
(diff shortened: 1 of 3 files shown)


src/Testcontainers.Pulsar/PulsarBuilder.cs
```diff
@@ -4,15 +4,15 @@ namespace Testcontainers.Pulsar;
 [PublicAPI]
 public sealed class PulsarBuilder : ContainerBuilder<PulsarBuilder, PulsarContainer, PulsarConfiguration>
 {
-    public const string PulsarImage = "apachepulsar/pulsar:3.0.6";
+    public const string PulsarImage = "apachepulsar/pulsar:3.0.8";
```
HofmeisterAn: Do the changes require bumping the patch version?

HofmeisterAn (commented):
Thanks for the PR! It looks like this PR relates to https://github.com/testcontainers/testcontainers-dotnet/pull/1294 but with fewer necessary changes 👍.

> @HofmeisterAn, would you prefer that I configure the tests to explicitly validate against both image versions (e.g., **3.0.X** and **4.0.X**), or is it sufficient to test exclusively with the latest default version (`3.0.X`)? Let me know how you'd like to proceed.

Usually, we test both versions if they are not compatible. We follow this pattern, which should be easy to apply to `PulsarContainerTest`. I believe we just need to inherit from `PulsarContainerTest`:

https://github.com/testcontainers/testcontainers-dotnet/blob/888da44936457c0d3dea0d149e4fb61f23544b33/tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs#L101-L108

Extending the tests would be great.

entvex:
> Thanks for the PR! It looks like this PR relates to #1294 but with fewer necessary changes 👍.
> 
> > @HofmeisterAn, would you prefer that I configure the tests to explicitly validate against both image versions (e.g., **3.0.X** and **4.0.X**), or is it sufficient to test exclusively with the latest default version (`3.0.X`)? Let me know how you'd like to proceed.
> 
> Usually, we test both versions if they are not compatible. We follow this pattern, which should be easy to apply to `PulsarContainerTest`. I believe we just need to inherit from `PulsarContainerTest`:
> 
> https://github.com/testcontainers/testcontainers-dotnet/blob/888da44936457c0d3dea0d149e4fb61f23544b33/tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs#L101-L108
> 
> Extending the tests would be great.

Sure, I'll do that. So we know that the two LTS versions work.

src/Testcontainers.Pulsar/PulsarBuilder.cs
entvex: > Do the changes require bumping the patch version?

It's not strictly required. But I would advise we do bump the patch version. It's also [backwards compatible](https://pulsar.apache.org/contribute/release-policy/#compatibility-between-releases) so it won't create any issues.

HofmeisterAn (approved):
Thanks 🙌.

PR #1362, opened 2025-02-06
feat: Add SFTP module

wim07101993:
## What does this PR do?

Adds a test container for an SFTP server. I implemented the atmoz/sftp server.

## Why is it important?

There is no container for SFTP yet.

## Related issues

none

## How to test this PR

Create a test container and write a file to it.


tests/Testcontainers.Sftp.Tests/SftpContainerTest.cs
```diff
@@ -0,0 +1,35 @@
...
+    [Fact]
+    [Trait(nameof(DockerCli.DockerPlatform), nameof(DockerCli.DockerPlatform.Linux))]
+    public async Task IsConnectedReturnsTrue()
+    {
+        // Given
+        var host = _sftpContainer.Hostname;
+
+        var port = _sftpContainer.GetMappedPublicPort(SftpBuilder.SftpPort);
+
+        using var sftpClient = new SftpClient(host, port, SftpBuilder.DefaultUsername, SftpBuilder.DefaultPassword);
```
HofmeisterAn, line 26: Adding the following line:

```csharp
using var textFile = sftpClient.CreateText("/home/sftp/upload/foo");
textFile.WriteLine("Foo");
```

fails with:

    Renci.SshNet.Common.SftpPathNotFoundException: No such file

HofmeisterAn (commented):
Thanks for the PR. I made some minor changes to align with the repository standards. I also noticed that uploading a file fails. I think we should address this issue before merging the PR.

tests/Testcontainers.Sftp.Tests/SftpContainerTest.cs
HofmeisterAn: Ah, I messed up the path. The correct path is `upload/foo`.

HofmeisterAn (approved):
Thanks 🙏

Issue #1085, opened 2024-01-05
[Enhancement]: Workaround for moby bug where random ports in a high range causes intermittent race conditions


martimors:
### Problem

It has previously been brought up multiple times across the testcontainers SDKs (eg. https://github.com/testcontainers/testcontainers-dotnet/issues/386) that the host port for a container can suddenly change right after the container has started. This causes tests to sometimes pass and sometimes fail, depending on what happened before; port changing or connection to the container from the host.

The underlying issue here seems to be https://github.com/moby/moby/issues/42442 , which has been open since 2021 without being fixed. It thus only happens on some systems and not all.

### Solution

Since the many projects already decided to work around this issue by using testcontainers in a way that specifically selects an open port instead of auto-assigning a high random port (https://github.com/Kong/kubernetes-ingress-controller/pull/5005 for example), I suggest making a permanent workaround as a feature in the .NET SDK. It can be automatic behind the scenes, or opt-in.

### Benefit

The project won't randomly fail tests on ipv6 enabled clients.

### Alternatives

Probably, I'm not too familiar with this project yet.

### Would you like to help contributing this enhancement?

No

HofmeisterAn:
Yes, this is a known issue. I'm curious where and how you encountered it. This problem should already be "resolved", considering this comment: https://github.com/docker/for-mac/issues/5588#issuecomment-934600089, incl. binding ports solely to the [IPv4](https://github.com/testcontainers/testcontainers-dotnet/issues/825) host IP (respectively resolving the right port to the right address IPv4/IPv6). Since .NET considers the empty string approach mentioned in the comment, port binding works reliable (AFAIK).

martimors:
We are only seeing this on MacOS clients. Here is the setup:

```
Client:
 Version:           24.0.7
 API version:       1.43
 Go version:        go1.21.3
 Git commit:        afdd53b4e3
 Built:             Thu Oct 26 07:06:42 2023
 OS/Arch:           darwin/arm64
 Context:           default

Server:
 Engine:
  Version:          24.0.7
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.10
  Git commit:       311b9ff
  Built:            Thu Oct 26 09:08:29 2023
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.25
  GitCommit:        d8f198a4ed8892c764191ef7b3b06d8a2eeb5c7f
 runc:
[...]

PR #1363, opened 2025-02-07
chore: Remove binding of ports to IPv4 only

### LINKED ISSUE #825: [Enhancement]: Resolve port bindings according to IPv4 and IPv6
### Problem

In some cases, Docker does not assign the same port to the IPv4 and IPv6 socket. Additionally, the operating system can resolve `localhost` to either an IPv4 or IPv6 address. To access the application or service running inside the container, it is essential to use the appropriate port binding according to the resolved IP. Otherwise, Testcontainers will be unable to connect to the app or service.

To avoid running into the issue mentioned above, Testcontainers for .NET only binds IPv4 bindings at present. However, this approach has a disadvantage, as it does not work in IPv6-only environments.

### Solution

Something like:

```csharp
// Contains two items:
// IsIPv6 ::1
// IsIPv4 127.0.0.1
var ips = Dns.GetHostAddresses("localhost");

var httpPort = 80;
var httpPortKey = httpPort + "/tcp";
var ipv4Binding = new PortBinding { HostIP = "0.0.0.0", HostPort = string.Format(CultureInfo.CurrentCulture, "{0}", ++httpPort) };
var ipv6Binding = new PortBinding { HostIP = "::", HostPort = string.Format(CultureInfo.CurrentCulture, "{0}", ++httpPort) };

IDictionary<string, IList<PortBinding>> mappedPortBindings = new Dictionary<string, IList<PortBinding>>();
mappedPortBindings.Add(new KeyValuePair<string, IList<PortBinding>>(httpPortKey, new List<PortBinding> { ipv4Binding, ipv6Binding }));
mappedPortBindings.TryGetValue(httpPortKey, out var portBindings);

// Lets imagine Dns.GetHostAddresses(string) returns the preferred order.
var addressFamilies = ips
  .Select(ip => ip.AddressFamily)
  .ToList();

// We can order the port bindings due to their address family, and use the first item.
var portBinding = portBindings
  .Select(portBinding => new IPEndPoint(IPAddress.Parse(portBinding.HostIP), ushort.Parse(portBinding.HostPort, NumberStyles.None, CultureInfo.InvariantCulture)))
  .OrderBy(portBinding => addressFamilies.IndexOf(portBinding.AddressFamily))
  .First();
```

### Benefit

Support either IPv4, IPv6 or both environments.

### Alternatives

- https://github.com/testcontainers/testcontainers-node/pull/485

Yes

### LINKED ISSUE #1360: [Bug]: ryuk container are not shutdown
### Testcontainers version

4.1.0

Yes

### Host OS

Windows

### Host arch

x86

### .NET version

9.0.102

### Docker version

```shell
Client:
 Version:           27.5.0-rd
 API version:       1.45 (downgraded from 1.47)
 Go version:        go1.22.10
 Git commit:        7a37716
 Built:             Tue Jan 14 17:50:50 2025
 OS/Arch:           linux/amd64
 Context:           default

Server:
 Engine:
  Version:          26.1.5
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.22.5
  Git commit:       411e817ddf710ff8e08fa193da80cb78af708191
  Built:            Fri Jul 26 17:51:06 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.17
  GitCommit:        3a4de459a68952ffb703bbe7f2290861a75b6b67
 runc:
  Version:          1.1.14
  GitCommit:        2c9f5602f0ba3d9da1c2596322dfc4e156844890
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### Docker info

```shell
Client:
 Version:    27.5.0-rd
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.19.3
    Path:     /mnt/c/Program Files/Rancher Desktop/resources/resources/linux/docker-cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.32.4
    Path:     /mnt/c/Program Files/Rancher Desktop/resources/resources/linux/docker-cli-plugins/docker-compose

Server:
 Containers: 18
  Running: 1
  Paused: 0
  Stopped: 17
 Images: 31
 Server Version: 26.1.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3a4de459a68952ffb703bbe7f2290861a75b6b67
 runc version: 2c9f5602f0ba3d9da1c2596322dfc4e156844890
 init version:
 Security Options:
  seccomp
   Profile: builtin
 Kernel Version: 5.15.167.4-microsoft-standard-WSL2
 Operating System: Rancher Desktop WSL Distribution
 OSType: linux
 Architecture: x86_64
 CPUs: 24
 Total Memory: 31.2GiB
 Name: CLFPF47JGAG
 ID: cd163be4-8ae6-4913-ba67-44fc96297a11
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Default Address Pools:
   Base: 172.27.0.0/16, Size: 20
```

### What happened?

Image

ryuk containers pile up and are not shutdown when the test execution is finished.
I would have expected that ryuk shuts itself down and does not leave containers running after the test execution is done.

### Relevant log output

```shell

```

### Additional information

_No response_

daviian:
## What does this PR do?

Removes the HostIP from the Port Binding so that it does not restrict itself to IPv4.

## Why is it important?

Not needed anymore.

## Related issues

- Closes #825
- Closes #1360

### DIFF
--- a/src/Testcontainers/Clients/ContainerConfigurationConverter.cs
+++ b/src/Testcontainers/Clients/ContainerConfigurationConverter.cs
@@ -148,9 +148,8 @@ namespace DotNet.Testcontainers.Clients
 
       public override IEnumerable<KeyValuePair<string, IList<PortBinding>>> Convert([CanBeNull] IEnumerable<KeyValuePair<string, string>> source)
       {
-        // https://github.com/moby/moby/pull/41805#issuecomment-893349240.
         return source?.Select(portBinding => new KeyValuePair<string, IList<PortBinding>>(
-          GetQualifiedPort(portBinding.Key), new[] { new PortBinding { HostIP = IPAddress.Any.ToString(), HostPort = portBinding.Value } }));
+          GetQualifiedPort(portBinding.Key), new[] { new PortBinding { HostPort = portBinding.Value } }));
       }
     }
   }


HofmeisterAn (approved):
Thanks

BrunoJuchli:
Could this possibly have caused the problem described [here](https://github.com/testcontainers/testcontainers-dotnet/issues/1085) to re-occur?
Since updating to TestContainers 4.2.0 we experience randomly failing tests.
From the error symptoms it looks like we're connecting to the wrong port of the container.
To give some context:

We run a fake SMTP Server in a docker container. The system-under-test connects to the SMTP port to send emails.
The test connects to the HTTP Port to retrieve all emails that have been sent.

The ports are configured using `ContainerBuilder.WithPortBinding(int port, assignRandomHostPort: true)`

We've not modified this code in the last few months; it used to work reliably.
Now we get SMTP responses on our HTTP Requests.

Of course, this being a timing/concurrency issue, it's entirely likely that the root cause has existed in the beginning, and only the execution timing has shifted recently.

HofmeisterAn:
@BrunoJuchli Which container runtime do you use?

BrunoJuchli:
@HofmeisterAn 
Thank you for the super quick reply 🥇 
Unfortunately I'm not too knowledgeable in this area, is it by chance what I get from the logs here?

> [testcontainers.org 00:00:00.17] Connected to Docker:
>   Host: unix:///var/run/docker.sock
>   Server Version: 26.1.3
>   Kernel Version: 6.8.0-1021-azure
>   API Version: 1.45
>   Operating System: Ubuntu 24.04.1 LTS

We're running this on GitHub Action - Ubuntu-latest. If the info above doesn't make it obvious to you on first sight what container runtime is used, I'm happy to dig deeper.
In that case I guess I will have to look into the runner-image and find out whether containerd, wasmtime, gvisor, kata containers,... is used, and which version thereof, correct?

HofmeisterAn:
> @HofmeisterAn Thank you for the super quick reply 🥇 Unfortunately I'm not too knowledgeable in this area, is it by chance what I get from the logs here?
> 
> > [testcontainers.org 00:00:00.17] Connected to Docker:
> > Host: unix:///var/run/docker.sock
> > Server Version: 26.1.3
> > Kernel Version: 6.8.0-1021-azure
> > API Version: 1.45
> > Operating System: Ubuntu 24.04.1 LTS
> 
> We're running this on GitHub Action - Ubuntu-latest. If the info above doesn't make it obvious to you on first sight what container runtime is used, I'm happy to dig deeper. In that case I guess I will have to look into the runner-image and find out whether containerd, wasmtime, gvisor, kata containers,... is used, and which version thereof, correct?

I assume you're encountering a different issue. I think the issue you're referring to is related to environments running Docker Desktop. I haven't noticed any flakiness in our CI pipeline (we're using GH-hosted runners too, though we use Ubuntu 22.04).

Have you been using 24.04 for a while? I believe there were some issues with the runners in general, if I'm not mistaken.

BrunoJuchli:
> I assume you're encountering a different issue. I think the issue you're referring to is related to environments running Docker Desktop. I haven't noticed any flakiness in our CI pipeline (we're using GH-hosted runners too, though we use Ubuntu 22.04).
> 
> Have you been using 24.04 for a while? I believe there were some issues with the runners in general, if I'm not mistaken.

@HofmeisterAn 
Thanks for the tip! We've not been using Ubuntu 24 for long.

I've executed some test runs with Ubuntu 24 and Ubuntu 22. Since our tests are quite long-running, the sample size isn't large - and the conclusion not a 100% reliable, but:
- Ubuntu 24: 2 out of 5 tests runs fail
- Ubuntu 22: 0 out of 6 test runs fail.

From previous runs I see that close to half of the test runs on Ubuntu 24 in recent days failed. So having 6 test runs on Ubuntu 22 pass looks like a reliable enough indicator to me.
= we're switching to Ubuntu 22, for now.

HofmeisterAn:
For reference, we have encountered some issues with the VM images in the past (anyone that depends on Docker) due to incompatible kernel versions. I do not know the exact details, but you can find more information about those issues here:

- https://github.com/actions/runner-images/issues/10649
- https://github.com/testcontainers/testcontainers-dotnet/pull/1265

It might be a good idea to start running our tests against version 24.04 as well, to see if there are any issues we can address or information we need to provide to the upstream: [GitHub Actions Runner Images](https://github.com/actions/runner-images).

PR #1367, opened 2025-02-08
feat: Add Apache Cassandra module

### LINKED ISSUE #1365: [Enhancement]: Add support for Apache Cassandra
### Problem

.NET Testcontainers do not support Apache Cassandra.

### Solution

Add a new builder for Apache Cassandra.

### Benefit

.NET has excellent and mature support for connecting to Cassandra. Adding .NET support for Cassandra in Testcontainers would enable .NET developers to write end-to-end acceptance and integration tests using the familiar Testcontainers paradigm, just like their Java counterparts do.

### Alternatives

Starting the Cassandra container outside of the solution manually.

Yes

gavindekock:
* Added support for Cassandra 5.0.3 TestConainer

## What does this PR do?

This PR adds an Apache Cassandra test container for .NET. The implementation follows a similar approach to the Apache Cassandra test container for Java.

## Why is it important?

.NET also has excellent mature and stable support for connecting to Cassandra. Therefore, this addition enables developers who use Cassandra with .NET to write end-to-end acceptance and integration tests using the familiar Testcontainers paradigm, just as their Java counterparts do.

## Related issues

- Closes #1365


src/Testcontainers.Cassandra/CassandraBuilder.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    /// <inheritdoc cref="ContainerBuilder{TBuilderEntity, TContainerEntity, TConfigurationEntity}" />
+    [PublicAPI]
+    public sealed class CassandraBuilder : ContainerBuilder<CassandraBuilder, CassandraContainer, CassandraConfiguration>
+    {
+        public const string CassandraImage = "library/cassandra:5.0.3";
+
+        public const ushort CqlPort = 9042;
+
+        public const string DefaultLocalDataCentre = "datacentre1";
```
aholeton: I believe 'datacenter1' should be the default with the American English spelling.

`public const string DefaultLocalDataCenter = "datacenter1";`

src/Testcontainers.Cassandra/CassandraContainer.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /// <summary>
+    /// Gets the Cassandra connection string.
+    /// </summary>
+    /// <returns>The Cassandra connection string.</returns>
+    public string GetConnectionString()
+    {
+        throw new NotImplementedException();
+    }
```
HofmeisterAn: Does Cassandra support the ADO.NET provider? If so, we need to implement this interface accordingly and update the tests. I assume we can use `CqlConnection` in that case.

HofmeisterAn (changes requested):
Thanks for the PR! I made a few minor changes to align with the repository standards, such as file-scoped namespaces, indentation, etc. Overall, the PR looks good. I have one question before merging it.

src/Testcontainers.Cassandra/CassandraContainer.cs
gavindekock: There appears to be some support for ADO.NET but it's not the recommended approach. Please see https://github.com/gavindekock/csharp-driver/tree/master/doc/features/components/adonet.

There is support for using a ConnectionString in the client driver, while I've not used that in the past in production I can't (at the moment) see why not.

HofmeisterAn (approved):
Thanks 🙏.

PR #1368, opened 2025-02-11
chore: Set PackageLicenseExpression property to MIT

ax-meyer:
## What does this PR do?

Switches from using the license file to indicate the license in the nuget package to using PackageLicenseExpression to explicitly define the license for the nugets.

## Why is it important?

This allows
- Allows tools like nuget-license (https://github.com/sensslen/nuget-license) to automatically detect the license of a package - currently we need to configure manual overrides and update the version in the override each time testconatainers gets updated
- nuget.org to show the license in the sidebar directly like this (see here: https://www.nuget.org/packages/dotnet-ef)
grafik

## Related issues

- no releated issues

Manual test:
- Run `nuget pack` for any project in the repo
- Start manual upload process for a nuget here: https://www.nuget.org/packages/manage/upload
- The page will show the package metadata with license information for verification before finishing the upload


HofmeisterAn (approved):
Thanks 💪.

PR #1370, opened 2025-02-14
chore: Add WithAcceptLicenseAgreement(bool) to container builder

HofmeisterAn:
## What does this PR do?

This PR introduces a new container builder method `WithAcceptLicenseAgreement(bool)`, which provides a base implementation and makes it easier for developers to implement modules that require license agreements.

Basically, modules need to override the following and call `ValidateLicenseAgreement()` in their `Build()` method. I expect modules to override `WithAcceptLicenseAgreement(bool)` and include proper documentation that references the relevant license information.

```csharp
protected override string AcceptLicenseAgreementEnvVar { get; } = "ACCEPT_EULA";

protected override string AcceptLicenseAgreement { get; } = "Y";

protected override string DeclineLicenseAgreement { get; } = "N";

public override ServiceBusBuilder WithAcceptLicenseAgreement(bool acceptLicenseAgreement)
{
    var licenseAgreement = acceptLicenseAgreement ? AcceptLicenseAgreement : DeclineLicenseAgreement;
    return WithEnvironment(AcceptLicenseAgreementEnvVar, licenseAgreement);
}

public override ServiceBusContainer Build()
{
    ValidateLicenseAgreement();
}
```

## Why is it important?

Recent module PRs include more and more modules that require license agreements. Previously, the code to check license acceptance was always copied manually. This PR simplifies the implementation of license acceptance, reduces code duplication, and ensures a consistent pattern across all modules.

## Related issues

\-


0xced:
Should this be implemented in the MsSql module too?

Currently, `ACCEPT_EULA=Y` is hardcoded so it would be a breaking change.

https://github.com/testcontainers/testcontainers-dotnet/blob/fa658551a0b22ba4ac8b36dba260b82c3844a9fc/src/Testcontainers.MsSql/MsSqlBuilder.cs#L64

HofmeisterAn:
@0xced Yes, I know the MSSQL module is still left. I did not change it because I am uncertain of the best way to transition to the new API without causing a breaking change (interfering with developers; ensuring that developers are aware of the change in advance).

TheConstructor:
@HofmeisterAn how about toggeling an internal flag with the license-acceptance method and writing warnings on build, if the flag suggests, that the license was not explicitly accepted or declined. Of course thr plan to enforce explicit license acceptance should go into the changelog, too.

I don't see, that there is a 100% option, but this should reach most

Issue #1153, opened 2024-04-10
[Feature]: Module for MailHog container for integration testing SMTP clients


Spelchure:
### Problem

Most of web applications need somewhat sending emails. MailHog is good choice SMTP for integration testing for sending emails. ( It also has API for checking whether email is sent or not) But it has drawback that it does not support SMTP over TLS under the hood.(Users of MailHog use proxy (e.g. stunnel) for eliminating this issue AFAIK) 

### Solution

We can create module for MailHog and encapsulate SSL logic inside this module. Users may decide whether they are want to use SSL or not through configuration. I’m planning to contribute this module what do you think about it?

### Benefit

Users easily use this module for integration testing for SMTP mail sending

### Alternatives

Users can use other docker images rather than MailHog or they may want to configure MailHog container themselves without using module 

### Would you like to help contributing this enhancement?

Yes

HofmeisterAn:
Hi 👋 did you try the existing Papercut ([1](https://github.com/jijiechen/Papercut), [2](https://github.com/testcontainers/testcontainers-dotnet/blob/aaafffce452bcd1d8ecedac452e21aca11d55ba7/tests/Testcontainers.Papercut.Tests/PapercutContainerTest.cs#L21-L46)) [module](https://www.nuget.org/packages/Testcontainers.Papercut)? I think both are very similar, although I do not know if they share/support the same features.

Spelchure:
Sorry my bad I didn't see Papercut module but it does not support SSL also if I am correct but I think adding extra SMTP server module while Papercut exists seems redundant. Thank you for your response

HofmeisterAn:
A few days ago, I came across [stalwartlabs/mail-server](https://github.com/stalwartlabs/mail-server) or Mailpit, another mail servers that supports various features, which could be alternatives. I think we can offer multiple modules for the same/similar services, as we do with databases. However, I would prefer to consolidate and prioritize support for established providers rather than incorporating all kinds of providers. For simpler modules, developers can always fall back to the generic builder API (MailHog seems quite established, though).

feslima:
@HofmeisterAn whether or not the repo of the SMTP server is active is important for the priorization process? It seems there is no active development for the mailhog repo ([source](https://github.com/mailhog/MailHog/issues/442#issuecomment-1493415258)).

HofmeisterAn:
[...]

PR #1377, opened 2025-02-18
feat: added smtp4dev container

wim07101993:
## What does this PR do?

Adds a test container for an SMTP server. I implemented the rnwood/smtp4dev server (https://github.com/rnwood/smtp4dev).

## Why is it important?

The current SMPT test container seems unmaintained. Without having looked too deeply into it I have noticed the last commit was more than 8 years ago. Next to that the jijiechen/Papercut container is a fork of ChangemakerStudios/Papercut-SMTP of which it is 450 commits behind. The original however does not have a web interface or api as far as I can tell.

I use smtp4dev regularly and think it has more or less the same functionality.

## Related issues

- [[Feature]: Module for MailHog container for integration testing SMTP clients #1153](https://github.com/testcontainers/testcontainers-dotnet/issues/1153)

## How to test this PR

Send a mail using the SMTP-server in the test container. Smtp4Dev has a web-ui in which you can see the inbox.


tests/Testcontainers.Smtp4Dev.Tests/SftpContainerTest.cs
rgylesbedford, line 1: Rename file to `Smtp4DevContainerTest.cs` ;-)

wim07101993: lol don't know how that slipped through :D

Issue #1383, opened 2025-02-26
[Bug]: The `UntilMessageIsLogged` wait strategy is incompatible with the reuse feature


0xced:
### Testcontainers version

4.3.0

### Using the latest Testcontainers version?

Yes

### Host OS

Any

### Host arch

Any

### .NET version

Any (9.0.2)

### Docker version

```shell
not relevant
```

### Docker info

```shell
not relevant
```

### What happened?

Running this app twice in a row succeeds the first time but fails the second and subsequent times.

The Oracle container was chosen because it's using the _until message is logged_ wait strategy.

**reuse.csproj**
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net9.0</TargetFramework>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Oracle.ManagedDataAccess.Core" Version="23.7.0" />
    <PackageReference Include="Testcontainers.Oracle" Version="4.3.0" />
  </ItemGroup>

</Project>
```

**Program.cs**
```csharp
using System;
using Oracle.ManagedDataAccess.Client;
using Testcontainers.Oracle;

await using var container = new OracleBuilder()
    .WithImage("gvenzl/oracle-free:23-slim-faststart")
    .WithName("Oracle.Reuse")
    .WithReuse(true)
    .Build();
await container.StartAsync();

var connectionString = new OracleConnectionStringBuilder(container.GetConnectionString())
{
    UserID = "SYS",
    DBAPrivilege = "SYSDBA",
};
await using var dataSource = OracleClientFactory.Instance.CreateDataSource(connectionString.ConnectionString);
await using var command = dataSource.CreateCommand("select version from v$instance");

try
{
    Console.WriteLine($"✅ {await command.ExecuteScalarAsync()}");
}
catch (Exception exception)
{
    Console.WriteLine($"❌ {exception}");
}
```

### Relevant log output

```shell
=== First run ===

[testcontainers.org 00:00:00.09] Connected to Docker:
  Host: unix:///Users/0xced/.orbstack/run/docker.sock
  Server Version: 27.5.1
  Kernel Version: 6.12.15-orbstack-00304-gd0ddcf70447d
  API Version: 1.47
  Operating System: OrbStack
  Total Memory: 15.66 GB
[...]

PR #1392, opened 2025-03-07
fix: Add support for postgres 9.2 and lower

KijongHan:
## What does this PR do?

This PR adds support for running Postgres containers below version 9.3. Adds to the `PostgreSqlBuilder.Build` method to check for Postgres versions below 9.3 (which doesn't have support for `pg_isready` to check for database readiness) using `Image.MatchVersion` - for these versions check for message logged in container using `UntilMessageIsLogged `helper method for the message "PostgreSQL init process complete; ready for start up."

Other strategies tried
- using psql to connect to container db and running `select 1` or `select version()` - these were not reliable and the database was not in a state ready to accept connections
- using "database system is ready to accept connections" - but couldn't reliably connect on the first instance of the message

related github discussion that helped determine above strategy: https://github.com/docker-library/postgres/issues/146

## Why is it important?

some older codebases are built on older versions of postgres, and being able to build tests using those older versions before migrating the version gives peace of mind for validating successful migration

### DIFF
--- a/src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs
+++ b/src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs
@@ -74,9 +74,16 @@ public sealed class PostgreSqlBuilder : ContainerBuilder<PostgreSqlBuilder, Post
     {
         Validate();
 
+        // The PostgreSql image does not contain pg_isready before version 9.3. Wait for logging message that says server is ready - "PostgreSQL init process complete; ready for start up."
+        // this was used instead of using psql select version() because the database is still not ready to accept external connection https://github.com/docker-library/postgres/issues/146
+        var isLegacyPostgreSql = DockerResourceConfiguration.Image.MatchVersion(v => v.Major <= 9 && v.Minor < 3);
         // By default, the base builder waits until the container is running. However, for PostgreSql, a more advanced waiting strategy is necessary that requires access to the configured database and username.
         // If the user does not provide a custom waiting strategy, append the default PostgreSql waiting strategy.
-        var postgreSqlBuilder = DockerResourceConfiguration.WaitStrategies.Count() > 1 ? this : WithWaitStrategy(Wait.ForUnixContainer().AddCustomWaitStrategy(new WaitUntil(DockerResourceConfiguration)));
+        var postgreSqlBuilder = DockerResourceConfiguration.WaitStrategies.Count() > 1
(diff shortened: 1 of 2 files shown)


HofmeisterAn (commented):
Thanks for the PR! The changes look good. However, TBH I'm not sure if I like adding support for a version that is EOL. Developers can simply override the default wait strategy or fall back to the generic container builder. But I agree that it is very difficult for developers to understand which versions are supported and which are not. This is something I've been thinking about a lot, but I haven't come up with a proper solution yet.

I don't know how to proceed from here. It's probably best to merge it so that other developers don't run into the same issue 🤔.

Note: I believe it was necessary to wait until the log messages were logged twice (for a reliable indication). I need to double-check the other language implementations.

0xced:
I have a [UntilDatabaseIsAvailable](https://github.com/0xced/testcontainers-dotnet/tree/UntilDatabaseIsAvailable) branch on my fork which would solve this issue. The `UntilDatabaseIsAvailable` wait strategy works with any ADO.NET provider. It could even have been used to workaround #1383! I actually copy/pasted this wait strategy in two projects as a workaround while waiting for Testcontainers 4.4.0 to be released.

I think it would be a perfect fit for this use case too (supporting PostgreSQL containers below version 9.3) and for any other potentially flaky wait strategy for database containers.

**Edit**: I just opened #1401. Then it could be used like this:
```csharp
new PostgreSqlBuilder()
    .WithImage("postgres:9.2")
    .WithWaitStrategy(Wait.ForUnixContainer().UntilDatabaseIsAvailable(NpgsqlFactory.Instance));
```

HofmeisterAn (commented):
Thanks for the PR. I had a quick discussion in our Slack workspace and created this discussion this morning: https://github.com/testcontainers/testcontainers-dotnet/discussions/1407, LMKWYT. I am heading out for a few days off now, I will review your (@0xced) PR when I get back. After that, we can decide how to proceed with this PR. Thanks again 🙏.

KijongHan:
> Thanks for the PR. I had a quick discussion in our Slack workspace and created this discussion this morning: #1407, LMKWYT. I am heading out for a few days off now, I will review your (@0xced) PR when I get back. After that, we can decide how to proceed with this PR. Thanks again 🙏.

Yeah I think your proposal there is quite reasonable, given this image is outside the LTS Im happy to close off this PR until the need arises. Appreciate your efforts!

PR #1395, opened 2025-03-11
feat: Add OpenSearch module

### LINKED ISSUE #1137: [Feature]: Add OpenSearch Container
### Problem

[OpenSearch](https://opensearch.org)
Codebases that rely on OpenSearch heavily to ingest logging & tracing.

 As of now OpenSearch provides Docker images, but there is not yet a test container available for this image, however they do already exist in [Go & Java ](https://golang.testcontainers.org/modules/opensearch/).

### Solution

Providing an OpenSearch container PR, so people can benefit from this image inside the testcontainers

### Benefit

Obviously having a opensearch container availble

### Alternatives

I am not aware of any viable alternative apart from ElasticSearch containers

Yes

digital88:
## What does this PR do?

Adds OpenSearch testcontainers implementation.

## Why is it important?

Currently there is no OpenSearch testcontainers implementation.

## Related issues

Hopefully closes https://github.com/testcontainers/testcontainers-dotnet/issues/1137

## How to test this PR

tests/Testcontainers.OpenSearch.Tests


digital88:
@HofmeisterAn  Hello, please take a look.

digital88:
I think this is ready for review

HofmeisterAn:
Thanks for updating. I'll review it in the next days.

src/Testcontainers.OpenSearch/OpenSearchBuilder.cs
```diff
@@ -0,0 +1,181 @@
...
+
+    public const int OpenSearchTransportPort = 9300;
+
+    public const int OpenSearchPerformanceAnalyzerPort = 9600;
+
+    public const string DefaultUsername = "admin";
+
+    public const string DefaultPassword = "yourStrong(!)P@ssw0rd";
+
+    public const string DefaultOldInsecurePassword = "admin";
```
HofmeisterAn: Do we need this (do we need to support the old version OOB)? If this is specific to a major version, we can set a different password as well, similar to what we're doing in Keycloak: https://github.com/testcontainers/testcontainers-dotnet/blob/6306178dde1a7d7bd75217d566bc383041ca6715/src/Testcontainers.Keycloak/KeycloakBuilder.cs#L68

Or throw an error in `Validate()`.

HofmeisterAn (commented):
Thanks for the PR! I finally had a time to review it. Overall, it looks good. I made a few adjustments for alignment and changed how the wait strategy is configured.

If we configure it using `WithDisabledSecurity(bool)`, the order of method calls starts to matter, which we generally try to avoid. For example, if a user calls `WithPassword(string)` afterward, the readiness check will fail.

In cases where builder APIs have dependencies like this, we either use a dedicated method to handle it or apply the final configuration in `Build()`.

I still need to review the tests tomorrow. If you're okay with the changes I made, we can go ahead and merge it then.

digital88:
> Thanks for the PR! I finally had a time to review it. Overall, it looks good. I made a few adjustments for alignment and changed how the wait strategy is configured.
> 
> If we configure it using `WithDisabledSecurity(bool)`, the order of method calls starts to matter, which we generally try to avoid. For example, if a user calls `WithPassword(string)` afterward, the readiness check will fail.
> 
> In cases where builder APIs have dependencies like this, we either use a dedicated method to handle it or apply the final configuration in `Build()`.
> 
> I still need to review the tests tomorrow. If you're okay with the changes I made, we can go ahead and merge it then.

Hello. Thanks for review. Overall all changes look good, I appreciate you taking time to do it.

src/Testcontainers.OpenSearch/OpenSearchBuilder.cs
digital88: I think it's reasonable to support old image versions too, given this is a brand new module. The old 'admin' password is specific (hardcoded) to all images before 2.12.0. Using same approach as in Keycloak module would be nice. I will make a check in Build().
Also the whole `DifferentImagesTest` file would not be needed. And the small section in docs too.

digital88: I pushed this change, please take a look.

HofmeisterAn (approved):
Thanks for the contribution 🙇.

PR #1398, opened 2025-03-12
feat(EventHubs, ServiceBus): Replace wait strategy with HTTP health check

scrocquesel-ml150:
## What does this PR do?

Reduce startup time of the service bus testcontainer by up to 10 seconds using the new available features.

## Why is it important?

Reduce total test run time.

## Related issues

- https://github.com/Azure/azure-service-bus-emulator-installer/issues/15
- https://github.com/Azure/azure-service-bus-emulator-installer/issues/35

### DIFF
--- a/src/Testcontainers.ServiceBus/ServiceBusBuilder.cs
+++ b/src/Testcontainers.ServiceBus/ServiceBusBuilder.cs
@@ -120,7 +120,7 @@ public sealed class ServiceBusBuilder : ContainerBuilder<ServiceBusBuilder, Serv
             .WithPortBinding(ServiceBusHttpPort, true)
             .WithEnvironment("SQL_WAIT_INTERVAL", "0")
             .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request =>
-                request.ForPath("/health").ForPort(ServiceBusHttpPort)));
+                request.ForPort(ServiceBusHttpPort).ForPath("/health")));
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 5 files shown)


src/Testcontainers.ServiceBus/ServiceBusBuilder.cs
```diff
@@ -115,9 +117,10 @@ protected override ServiceBusBuilder Init()
         return base.Init()
             .WithImage(ServiceBusImage)
             .WithPortBinding(ServiceBusPort, true)
-            .WithWaitStrategy(Wait.ForUnixContainer()
-                .UntilMessageIsLogged("Emulator Service is Successfully Up!")
-                .AddCustomWaitStrategy(new WaitTwoSeconds()));
+            .WithPortBinding(ServiceBusHttpPort, true)
+            .WithEnvironment("SQL_WAIT_INTERVAL", "0")
+            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request =>
+                request.ForPath("/health").ForPort(ServiceBusHttpPort)));
```
HofmeisterAn: Shouldn't we check the body (similar to what we do [here](https://github.com/testcontainers/testcontainers-dotnet/blob/99995505d8161e58d19b111fef0b4a614f8cb979/src/Testcontainers.Couchbase/CouchbaseBuilder.cs#L284-L305))? Does Event Bus contain this health check too?

krishankumar95: @HofmeisterAn ; Yep EventHub Emulator also gets the same Health Check API with the latest release. The Req/Response contracts are same for both. 

Container URI for HC API :
```
GET : {host}:5300/health
```
Respose Body:
```
{"status":"healthy","message":"Service is healthy."}
```

scrocquesel-ml150: In case the service is not available, the endpoint returns a "503 Service Unavailable" with this body

```
{
  "status": "unhealthy",
  "message": "Service not available. Please try after some time."
}
```

This is not really documented so I keep it simple and similar to what is done in k8s.

For Event Bus, I just test and it seems it is available. Would you like a unique PR for both ?

HofmeisterAn: Ah, if it returns an HTTP status other than 200, I don't think we need to check the body. I just want to make sure we won't get something like this:

```
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 96

{
    "status": "unhealthy",
    "message": "Service not available. Please try after some time."
}
```

> For Event Bus, I just test and it seems it is available. Would you like a unique PR for both ?

Whatever is easier for you. I am OK with adding both in this PR. Thanks 🙏.

HofmeisterAn (dismissed):
Thanks!

HofmeisterAn (approved):

PR #1378, opened 2025-02-20
chore: Remove `SharedContainerInstance` and use `ContainerFixture` instead

0xced:
Now that `Testcontainers.Xunit` is available we can get rid of `SharedContainerInstance` in the `Testcontainers.Commons` project since it serves the same purpose.

PR #1401, opened 2025-03-15
feat: Add a wait strategy that waits until the ADO.NET database is available

0xced:
## What does this PR do?

This pull request introduces a new wait strategy for containers implementing the `IDatabaseContainer` interface. It tries to open a connection with the container connection string until it works or until it times out.

## Why is it important?

This new wait strategy is quite powerful as it can be used for all database containers. It could have been used to workaround #1383 and it could be used as an alternative to #1392.

## Related issues

Mentioned above.

## How to test this PR

New tests have been added to use this new wait strategy for database containers in addition to the existing wait strategies.

## Follow-ups

Explicitly supporting PostgreSQL < 9.3 should be done in a follow-up pull request by using this new wait strategy.


src/Testcontainers/Configurations/WaitStrategies/IWaitForContainerOS.cs
```diff
@@ -112,6 +114,25 @@ public interface IWaitForContainerOS
...
+    /// <param name="timeout">
+    /// The maximum duration to retry. Once reached, a <see cref="TimeoutException" /> is thrown.
+    /// Defaults to 5 minutes if not specified.
+    /// </param>
+    /// <returns>A configured instance of <see cref="IWaitForContainerOS" />.</returns>
+    /// <remarks>
+    /// This wait strategy must only be applied to containers implementing the <see cref="IDatabaseContainer"/> interface.
+    /// </remarks>
+    [PublicAPI]
+    IWaitForContainerOS UntilDatabaseIsAvailable(DbProviderFactory dbProviderFactory, TimeSpan frequency = default, TimeSpan timeout = default);
```
HofmeisterAn: Instead of using `frequency` and `timeout`, we should use the `Action<IWaitStrategy> waitStrategyModifier = null` argument, just like the other wait strategies do. That already gives you the ability to configure the [behaviors](https://github.com/testcontainers/testcontainers-dotnet/blob/2dc9f574a0e05b629bc4dcf9ca6d627aaa233aa7/src/Testcontainers/Configurations/WaitStrategies/IWaitStrategy.cs#L10-L31).

HofmeisterAn (changes requested):

HofmeisterAn (commented):
We need to update the Db2 tests as well. If you're busy, I can take care of both tasks.

tests/Testcontainers.FirebirdSql.Tests/FirebirdSqlContainerTest.cs
```diff
@@ -93,4 +93,13 @@ public FirebirdSqlSysdba()
...
     }
+
+    [UsedImplicitly]
+    public sealed class FirebirdSqlWaitForDatabase : FirebirdSqlContainerTest
+    {
+        public FirebirdSqlWaitForDatabase()
+            : base(new FirebirdSqlBuilder().WithWaitStrategy(Wait.ForUnixContainer().UntilDatabaseIsAvailable(FirebirdClientFactory.Instance)).Build())
+        {
+        }
+    }
```
HofmeisterAn: Is there any reason why you didn’t apply the container fixture to all the tests? I think we can add the xUnit module to the common test project and use the container fixture everywhere (for `IDatabaseContainer`). That should also reduce overhead and improve performance.

src/Testcontainers/Configurations/WaitStrategies/IWaitForContainerOS.cs
0xced: I addressed this in f579265e5180a960183027f0db59663a11a17a2b. Indeed it doesn't make sense to reinvent a wait strategy mechanism when there's already one!

I'm just a little sad that by doing so, the timeout exception is degraded.

😀 Before (includes the connection string and the timeout)
```
System.TimeoutException: The database was not available at "Host=127.0.0.1;Port=32897;Database=postgres;Username=postgres;Password=postgres" after waiting for 2 seconds.
```

🙁 After (no useful information at all)
```
System.TimeoutException: The operation has timed out.
```

tests/Testcontainers.FirebirdSql.Tests/FirebirdSqlContainerTest.cs
0xced: > Is there any reason why you didn’t apply the container fixture to all the tests?

I just missed some I guess. I will fix it soon for Firebird and Db2. And it will be a huge win for Db2, which is taking over 7 minutes to start on my machine!

> I think we can add the xUnit module to the common test project and use the container fixture everywhere (for IDatabaseContainer).

That's not a good idea. Adding a project reference to `Testcontainers.Xunit.csproj` in the `Testcontainers.Commons` project would prevent testing `Testcontainers.XunitV3.csproj` in other test projects because of xunit incompatibilities between v2 and v3. I purposefully removed the `PackageReference` to `xunit` in #1378.

0xced:
Now using a `DbContainerFixture` in `Db2ContainerTest` (done in a15cfc5340c1237da0db2866fa8182a909e19a94)

0xced:
Now using a `DbContainerFixture` in `FirebirdSqlContainerTest` too (done in ca8a1784e907039cbf09da0cb1fafbbe24bd0d20)

And here something extremely interesting happened! The test executes in 2 seconds when using the `UntilDatabaseIsAvailable` vs 30 seconds for the default wait strategy!

FirebirdSqlDefault vs FirebirdSqlWaitForDatabase

src/Testcontainers/Configurations/WaitStrategies/IWaitForContainerOS.cs
HofmeisterAn: That's something I noticed as well. Maybe we can improve it in a separate PR.

HofmeisterAn (commented):
Thanks for the adjustments.

Unfortunately, I'm running into an issue locally when I run the Cassandra tests multiple times.

The `ConnectionStateReturnsOpen()` test fails for `CassandraContainerDefaultConfiguration` when I run all the Cassandra tests together (including `CassandraFixtureWaitForDatabase`). But if I run just the `CassandraContainerDefaultConfiguration` tests on their own, everything passes.

> All hosts tried for query failed (tried 127.0.0.1:52448: SocketException 'A request to send or receive data was disallowed because the socket is not connected and (when sending on a datagram socket using a sendto call) no address was supplied.')

I believe this is a connection pooling issue. I think I've seen something similar before.

HofmeisterAn (commented):
I've updated the remaining test projects as well. They now all use the same structure and test at least the default and overridden wait strategy configurations.

I'll update the module documentation later today. At least, we need to review it, as it might no longer be accurate.

IMO, the PR looks good. We just need to figure out why Cassandra is failing and check whether other modules are experiencing similar issues.

**Edit**:

On CI Cassandra failed with a similar issue:

>   Cassandra.NoHostAvailableException : All hosts tried for query failed (tried 127.0.0.1:32769: SocketException 'Transport endpoint is not connected')
Raw output

0xced:
> On CI Cassandra failed with a similar issue

Of course, _it works on my machine_. 🤷🏻‍♂️

I tried something in 3d22bf56b437b670a6ac13bcf0523b83a7d88de2 for FirebirdSql. This could be applied to all other db modules too.

The default wait strategy automatically switches to the new `UntilDatabaseIsAvailable` strategy if the `DbProviderFactory` is available at runtime (probed through reflection in order not to take a hard dependency on the ADO.NET provider). This requires two tests projects, one that does not reference the ADO.NET provider and one that does.

What do you think? Good idea or not?

Note that applying the same mechanism to the PostgreSQL module would automatically add support for older images where `pg_isready` is not available, fixing #1392 without relying on specific log messages.

HofmeisterAn:
> Of course, _it works on my machine_. 🤷🏻‍♂️

I've tried it on two different machines and it's failing in CI. I'll take a look at it after the public holidays.

> What do you think? Good idea or not?

I'd prefer to discuss this in a separate PR. The PR looks good overall, except for the necessary Cassandra fix.

I'm not a huge reflection fan (I guess no one is 😆), but it's an interesting idea. It could help support more configurations and might even be faster in some cases. But we definitely need a way to opt out or fallback.

0xced:
> I'd prefer to discuss this in a separate PR. The PR looks good overall, except for the necessary Cassandra fix.

OK, I removed the commits that use reflection to get the `DbProviderFactory` instances and force-pushed. I kept the work in a new `UntilDatabaseIsAvailable+Reflection` branch to discuss later.

kiview:
For Postgres a specific hint, relying on a single SQL query (or even on `pg_isready` for that matter), is usually a tight race condition on the vanilla PostgreSQL image, since the engine will restart after bootstrapping (and strategy might succeed before the restart).

HofmeisterAn:
> For Postgres a specific hint, relying on a single SQL query (or even on `pg_isready` for that matter), is usually a tight race condition on the vanilla PostgreSQL image, since the engine will restart after bootstrapping (and strategy might succeed before the restart).

For `pg_isready`, the race condition won't happen if you specify the host argument AFAIK. That's what we've been using for a long time (we did have issues without the host argument, indeed).

HofmeisterAn:
I had some time to look into the Cassandra issue. It looks like a connection pooling issue. The test that [opens](https://github.com/testcontainers/testcontainers-dotnet/blob/3f263578ee0e64c85293c2dc9e25729d91522580/tests/Testcontainers.Cassandra.Tests/CassandraContainerTest.cs#L14) the DB connection fails after a few runs. The connection string in the exception message doesn't match the one that's actually configured and used by/for the container:

- Exception message: _All hosts tried for query failed (tried 127.0.0.1:51732: SocketException '[...]')_
- Connection string: `contact points=127.0.0.1;port=51742`

The only way I found to configure pooling is through `Cluster.Builder()`, not through ADO.NET (nor can we clear the pool).

Not sure how to move forward from here. Maybe we can ask for guidance in the upstream repo (I believe this is a general issue with the client implementation).

WDYT?

0xced:
I was finally able to reproduce the problem by changing the `CassandraContainerTest` constructor to include an artificial delay for the default configuration!

```csharp
public CassandraContainerTest(CassandraDefaultFixture fixture)
{
    this.fixture = fixture;
    if (GetType().Name.Contains("CassandraDefaultConfiguration"))
    {
        Thread.Sleep(TimeSpan.FromSeconds(2));
    }
}
```

I see two possible solutions to this problem that occurs because the `CqlConnection` class uses a **static** concurrent dictionary of clusters keyed by cluster name, i.e. shared across all connections.

1. We update the test to explicitly set different cluster names.

```patch
diff --git a/tests/Testcontainers.Cassandra.Tests/CassandraContainerTest.cs b/tests/Testcontainers.Cassandra.Tests/CassandraContainerTest.cs
index d351eab..62ebcb6 100644
--- a/tests/Testcontainers.Cassandra.Tests/CassandraContainerTest.cs
+++ b/tests/Testcontainers.Cassandra.Tests/CassandraContainerTest.cs
@@ -8,7 +8,8 @@ public abstract class CassandraContainerTest(CassandraContainerTest.CassandraDef
     public void ConnectionStateReturnsOpen()
     {
         // Given
-        using DbConnection connection = fixture.CreateConnection();
+        var connectionString = new CassandraConnectionStringBuilder(fixture.ConnectionString) { ClusterName = GetType().Name }.ConnectionString;
+        using DbConnection connection = new CqlConnection(connectionString);
 
         // When
         connection.Open();
```

2. We update the `CassandraContainer` to include a unique cluster name per host/port in the connection string.

```patch
diff --git a/src/Testcontainers.Cassandra/CassandraContainer.cs b/src/Testcontainers.Cassandra/CassandraContainer.cs
index 3e44d61..2709655 100644
--- a/src/Testcontainers.Cassandra/CassandraContainer.cs
+++ b/src/Testcontainers.Cassandra/CassandraContainer.cs
@@ -18,7 +18,9 @@ public sealed class CassandraContainer : DockerContainer, IDatabaseContainer
     {
         var properties = new Dictionary<string, string>();
         properties.Add("Contact Points", Hostname);
-        properties.Add("Port", GetMappedPublicPort(CassandraBuilder.CqlPort).ToString());
+        var port = GetMappedPublicPort(CassandraBuilder.CqlPort).ToString();
+        properties.Add("Port", port);
+        properties.Add("Cluster Name", $"{Hostname}:{port}");
         return string.Join(";", properties.Select(property => string.Join("=", property.Key, property.Value)));
     }
```

Both strategies work. I think the second one is a bit better but I'm not a Cassandra user so I'm not 100% sure.

0xced:
I just pushed solution ＃2 in ac743cef10f28b56ea6c3070c12104d1974c2094.

✅ The tests are now all passing!

HofmeisterAn:
Thanks for the adjustments.

I prefer setting the cluster name property too. Leaving it unset prevents Testcontainers users from running multiple containers. If any issues come up afterward, we can deal with them as they arise.

I also prepared an issue I'd like to share with DataStax. Unfortunately, I can't create a GitHub or Jira issue at the moment, I'll send the following to their mailing list later:

We're running into a race-condition with the DataStax C# driver for Apache Cassandra, and I believe this is a general issue with the client implementation.

We're running multiple Apache Cassandra containers on the same host. It looks like the client [caches](https://github.com/datastax/csharp-driver/blob/aa4f36e5c2e01042de82bab42a004bbd9a1c594b/src/Cassandra/Data/CqlConnection.cs#L166-L172) connections based on the cluster name. If the connection string doesn't specify a cluster name, it uses the [default](https://github.com/datastax/csharp-driver/blob/aa4f36e5c2e01042de82bab42a004bbd9a1c594b/src/Cassandra/CassandraConnectionStringBuilder.cs#L26) cluster name.

Here's what happens:

1. Create an Apache Cassandra instance listening on `localhost:51732` (**container_1**).
1. Create an Apache Cassandra instance listening on `localhost:51742` (**container_2**).
1. Create a `CqlConnection` instance for **container_1**.
    - This creates a connection using the default cluster name: `Cassandra Cluster`.
1. Run tests against **container_1**, and tear it down afterwards.
1. Create a `CqlConnection` instance for **container_2**.
    - This is where the race-condition happens. It looks like the client implementation still tries to connect to **container_1**, even though the connection string is correctly set for **container_2**.
    - This could even lead to false assumptions. For example, if we didn't tear down **container_1** beforehand, the client might end up connecting to the wrong instance.
    - From a quick look at the code, it seems like the cluster name is never removed from the internal dictionary during `Close()` or `Dispose()`.

HofmeisterAn (approved):
Thanks 🙇

PR #1410, opened 2025-03-26
chore: Use Docker Compose to host documentation locally

digital88:
## What does this PR do?

Added `compose.yml` file to been able to start local mkdocs server without installing python3.8 on host machine. Took inspiration from tc for node project.

## Why is it important?

Do not need to install python on host to be able to work with docs.

## How to test this PR

Run `docker compose up` and visit `http://localhost:8000`

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -1,0 +1,9 @@
+services:
+  docs:
+    image: python:3.8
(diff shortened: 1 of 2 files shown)


HofmeisterAn (approved):
Thanks 👍 good idea!

PR #1416, opened 2025-04-02
feat: Add Socat container implementation

eddumelendez:
## What does this PR do?

This PR adds an utility `SocatContainer` that can serve as a proxy to other containers. 

## Why is it important?

* [connect to the Docker Model Runner](https://github.com/eddumelendez/llm-runner-java-example/blob/main/src/test/java/org/example/LlmRunnerTest.java#L25-L29)
* Testcontainers for Java also use it at compose implementation, making service ports available at random ports instead of the ones defined at compose definition.

## Related issues

N/A


HofmeisterAn (approved):
Hey Eddú, thanks for the PR. While reviewing, I noticed two issues. One of them seems like a bug we hadn't caught before.

- Usually, the builders don't need to handle merging any values. That's automatically done by the underlying implementation. It looks like a constraint prevented that for the `SocatConfiguration` class. See my changes in `BuildConfiguration`.
- I think we can remove exposing the port `8080` for the _hello world_ container. I think all traffic should go through Socat and the Docker network.

If you're good with these changes, I'm happy to merge the PR.

HofmeisterAn:
It looks like there is an issue pulling images from icr.io, probably a temporary issue.

eddumelendez:
Hi @HofmeisterAn, thank you so much for all the improvements! learned so much by submitting this PR. 🚀

PR #1421, opened 2025-04-14
docs(ClickHouse): Add example

digital88:
## What does this PR do?

Add ClickHouse example to modules section in docs.

## Why is it important?

ClickHouse module docs are missing.

## How to test this PR

Simply run `docker compose up` and access the docs at: `http://localhost:8000`.

### DIFF
--- a/docs/modules/activemq.md
+++ b/docs/modules/activemq.md
@@ -30,9 +30,9 @@ You can start an Apache ActiveMQ Artemis container instance from any .NET applic
 
 Connect to the container and produce a message:
 
-=== "EstablishesConnection"
+=== "Establish connection"
     ```csharp
-    --8<-- "tests/Testcontainers.ActiveMq.Tests/ArtemisContainerTest.cs:ArtemisContainerEstablishesConnection"
+    --8<-- "tests/Testcontainers.ActiveMq.Tests/ArtemisContainerTest.cs:EstablishConnection"
     ```
 
 The test example uses the following NuGet dependencies:
--- a/docs/modules/clickhouse.md
+++ b/docs/modules/clickhouse.md
@@ -12,20 +12,22 @@ You can start a ClickHouse container instance from any .NET application. This ex
 
 === "Test class"
     ```csharp
-    --8<-- "tests/Testcontainers.ClickHouse.Tests/ClickHouseContainerTest.docs.cs:Class"
+    --8<-- "tests/Testcontainers.ClickHouse.Tests/ClickHouseContainerTest.docs.cs:UseClickHouseContainer"
     }
     ```
 
-Connecting to container example:
-=== "Connecting to ClickHouse"
+Connect to the container:
+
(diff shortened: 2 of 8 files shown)


HofmeisterAn (commented):
Thanks for the PR. I haven't merged it yet because there was another one open that changed the ClickHouse tests (it adds the ADO.NET wait strategy: https://github.com/testcontainers/testcontainers-dotnet/pull/1401). We'll need to update the documentation for this, similar to what we did for the other modules, I think.

Honestly, I don't think our current setup is ideal. The tests are optimized for testing, but not really for documentation purposes. Maybe it would make sense to use separate classes just for the documentation in the future. The original idea was to reference real source code in the docs to make sure everything compiles. But that's something we can discuss and address in a separate issue/PR.

Thanks again.

digital88:
> there was another one open that changed the ClickHouse tests

Yes, it completely changed everything 😄  I just copied the old tests' code for docs, because new tests are more..massive? (at least for a quick docs example). They also cross reference another project, I don't see good way to include all this in docs examples...This boilerplate:

> This example uses xUnit.net's IAsyncLifetime interface...

is now included in `DbContainerFixture` class in `Testcontainers.Xunit` project. I tried to use "scissors" --8<-- to reference the new tests in docs, but it's not that easy now. You either send reader straight to source code or include all base classes and base fixture classes in examples 😄

HofmeisterAn (commented):
Sorry for the back and forth. What do you think about creating a separate file (something like `ClickHouseContainerTest.docs.cs`) alongside the test class? This file would contain the sources used for documentation.

We could make the test class `partial`, and then use a nested class just for documentation purposes. That way, we keep things clean and separate.

```csharp
public abstract partial class ClickHouseContainerTest
{
    public sealed class Documentation
    {
        private readonly ClickHouseContainer _clickHouseContainer = new ClickHouseBuilder().Build();
    }
}
```

LMKWYT.

digital88:
> Sorry for the back and forth. What do you think about creating a separate file (something like `ClickHouseContainerTest.docs.cs`) alongside the test class? This file would contain the sources used for documentation.
> 
> We could make the test class `partial`, and then use a nested class just for documentation purposes. That way, we keep things clean and separate.
> 
> ```cs
> public abstract partial class ClickHouseContainerTest
> {
>     public sealed class Documentation
>     {
>         private readonly ClickHouseContainer _clickHouseContainer = new ClickHouseBuilder().Build();
>     }
> }
> ```
> 
> LMKWYT.

Yes, this looks like a solution. Not sure why we need partial class?

HofmeisterAn:
> Yes, this looks like a solution. Not sure why we need partial class?

Oh, we don't need it. I just thought it might be nice to use a separate file to keep things organized, without mixing too much into one file (while keeping the same class name).

HofmeisterAn (approved):
Thanks! I am just waiting until `icr.io` is back and working again.

PR #1424, opened 2025-04-16
feat(ServiceBus): Add builder API to upload config file

### LINKED ISSUE #1423: [Enhancement]: Azure Service Bus Custom Config.json
### Problem

It appears as though the current implementation of ServiceBusBuilder doesn't provide a way to customize the service bus configuration. For example, the Java version of TestContainers has an option like:
`.withConfig(MountableFile.forClasspathResource("/service-bus-config.json"))`

The default configuration provides a single queue, a single topic with 3 subscriptions. The problem I'm having is that those 3 subscriptions have filters on them which prevent the messages I am sending to being picked up by my consumer. 

### Solution

A solution has already been determined by the Java testcontainers. Add an method called WithConfig to pass in the .json config file.

`.withConfig(MountableFile.forClasspathResource("/service-bus-config.json"))`

### Benefit

Being able to configure the service bus to better match the production service bus would provide tests that better reflect the environment they will run in.

### Alternatives

An alternative to providing a file would be to make methods to define the queues/topics. For example:
``` C#
var serviceBusContainer = new ServiceBusBuilder()
  .WithImage("mcr.microsoft.com/azure-messaging/servicebus-emulator:latest")
  .WithQueue("my-queue", options => {
          // allow cusomtization of the queue
   })
  .WithTopic("my-topic", options => {
     options.WithSubscription("sub1", subOptions => {...})
     options.WithSubscription("sub2", subOptoins => {...})
  })
  .Build();
```

Yes

chasewallis:
## What does this PR do?

Add a simple way for users to use a custom Config.json for configuration Azure Service Bus.

## Why is it important?

Without this change, to use a custom Config.json, user must manually map the resource file to a specific location. This change reduces that headache in knowing what exactly needs to be done.

## Related issues

- Closes #1423 

## How to test this PR
Create a custom Config.json file with a unique queue name. Run the testcontainer and attempt to send/receive from that queue name.


HofmeisterAn (approved):
Thank you for taking care of it 🙇.

PR #1431, opened 2025-04-20
feat: Allow canceling container start in xUnit.net v3 fixtures

TheConstructor:
## What does this PR do?

This passes `TestContext.Current.CancellationToken` to `Container.StartAsync`, so that cancelling a test-session may cancel starting the requested Testcontainer

## Why is it important?

If the test-session is already set to cancelled, the fixture is likely disposed, just after being started, without being used by any test. Waiting for the start waists time and resources.

## How to test this PR

Start any test utilizing `Testcontainers.XunitV3`, and cancel test-execution during container creation/startup. It should cancel quickly, reporting a TaskCancelledException to the test starting up


HofmeisterAn (approved):
Thank you 🙇.

PR #1432, opened 2025-04-22
fix(EventHubs): Support default consumer group name

### LINKED ISSUE #1429: [Bug]: EventHub configuration produce errors
### Testcontainers version

4.4.0

Yes

### Host OS

Linux

### Host arch

x86

### .NET version

8

### Docker version

```shell
N/A
```

### Docker info

```shell
N/A
```

### What happened?

```csharp
var configuration = EventHubsServiceConfiguration
            .Create()
            .WithEntity("test", 2, [Azure.Messaging.EventHubs.Consumer.EventHubConsumerClient.DefaultConsumerGroupName]);

        var builder = new EventHubsBuilder()
            .WithAcceptLicenseAgreement(true)
            .WithConfigurationBuilder(configuration);

        _eventHubsContainer = builder.Build();

        await _eventHubsContainer.StartAsync();
```

1. The default namespace name produce a "Recoverable validation failed" logs but this does not fail. We should change the default name to the valid one "emulatorns1".
2. The emulator create a "$default" group and do not allow to specify it in the config. But the configuration validate that we must pass at least one consumer group which is not necessary if we use the default one.
3. When the container stop due to a a bad configuration, testcontainer still wait for it to be ready.

### Relevant log output

```shell
warn: HostHelper[0]
      Recoverable validation failed on user config:Expected string to be "emulatorns1" with a length of 11 because NamespaceName is non-modifiable.Only supported emulatorns1, but "ns-1" has a length of 4, differs near "ns-" (index 0).
Unhandled exception. FluentAssertions.Execution.AssertionFailedException: Expected string not to be equivalent to "$default" because Consumer group $default is internally created., but they are.
```

### Additional information

I can push a PR for the first two points.
Some guidance on how should the wait strategy be modified so it detect the container has stopped is welcome.

scrocquesel-ml150:
## What does this PR do?

This PR improve the eventHub service configuration builder.

1. It uses the correct default namespace name, thus removing the warning 

    ```
    Recoverable validation failed on user config:Expected string to be "emulatorns1" with a length of 11 because NamespaceName is non-modifiable.Only supported emulatorns1, but "ns-1" has a length of 4, differs near "ns-" (index 0).
    ```
2. It allows to pass an empty consumerGroup array and an array containing `$default`. This consumer group is always created by the emulator.

3. It limits the number of hubs that can be created as per the emulator quotas.

4. In case the configuration is still invalid, it doesn't wait indefinitely on a stopped container.

## Why is it important?

In my test cases, I use the constant provided by the SDK `EventHubConsumerClient.DefaultConsumerGroupName` (= `$default`) and it was not possible to configure the container with only this consumer group.

## Related issues

- Closes #1429


src/Testcontainers.EventHubs/EventHubsBuilder.cs
```diff
@@ -139,7 +139,10 @@ protected override EventHubsBuilder Init()
             .WithPortBinding(EventHubsPort, true)
             .WithPortBinding(EventHubsHttpPort, true)
             .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request =>
-                request.ForPort(EventHubsHttpPort).ForPath("/health")));
+                request.ForPort(EventHubsHttpPort).ForPath("/health"),
+                // If the configuration is invalid, the container will stop immediately.
+                // In this case, we need no to wait for the container to be ready indefinitely.
+                waitStrategyModifier => waitStrategyModifier.WithTimeout(TimeSpan.FromSeconds(30))));
```
HofmeisterAn: Thanks for your PR. Could we please revert this change? As mentioned in the issue, we need to address this differently. We do not set a default timeout on purpose because we do not know anything about the user's environment, hardware, etc.

Instead, we should throw an exception if the container stops during the wait strategy check, but this should be handled in a separate PR.

scrocquesel-ml150: done

HofmeisterAn (approved):

PR #1439, opened 2025-04-30
Upgrade Azurite: 3.28.0 → 3.34.0 (latest)

aviflax:
## What does this PR do?

Upgrades the Azurite image from 3.28.0 to 3.34.0 (the latest).

## Why is it important?

The older version of Azurite is not compatible with the latest Azure Storage client packages, so without this change, users (like me) are blocked from upgrading those packages.

## Related issues

N/A


HofmeisterAn (commented):
Thanks for the PR.

> users (like me) are blocked from upgrading those packages.

The container builder API lets you override and pin the image version using `WithImage(string)`, which is the approach we [recommend](https://dotnet.testcontainers.org/api/best_practices/). We only update image versions when it's absolutely necessary and unavoidable. In general, we avoid bumping the image version to prevent introducing breaking changes for developers.

aviflax:
@HofmeisterAn thank you! TIL!

PR #1441, opened 2025-04-30
chore: Update test-framework to xUnit.net v3

TheConstructor:
## What does this PR do?

Currently all test-projects except `Testcontainers.XunitV3.Tests` use Xunit.Net v2. After this PR all test-projects except `Testcontainers.Xunit.Tests` use Xunit.Net v3

## Why is it important?

I would expect that some day in the future Xunit.Net v2 will be discontinued. Also switching to Xunit.Net v3 means, that we can use all the new features like further Asserts and better test-cancellation. See also https://xunit.net/docs/getting-started/v3/whats-new for details

While I do think, that this should be done one day, it does not necessarily need to be today. On the other hand, I did not have problems using v3 so far (in other projects), and it will probably never be easier to switch anyway.

### DIFF
--- a/tests/Testcontainers.ArangoDb.Tests/ArangoDbContainerTest.cs
+++ b/tests/Testcontainers.ArangoDb.Tests/ArangoDbContainerTest.cs
@@ -6,7 +6,8 @@ public sealed class ArangoDbContainerTest : IAsyncLifetime
 
     public async ValueTask InitializeAsync()
     {
-        await _arangoDbContainer.StartAsync();
+        await _arangoDbContainer.StartAsync()
+            .ConfigureAwait(false);
     }
 
     public ValueTask DisposeAsync()
--- a/tests/Testcontainers.ArangoDb.Tests/ArangoDbContainerTest.cs
+++ b/tests/Testcontainers.ArangoDb.Tests/ArangoDbContainerTest.cs
@@ -4,14 +4,14 @@ public sealed class ArangoDbContainerTest : IAsyncLifetime
 {
     private readonly ArangoDbContainer _arangoDbContainer = new ArangoDbBuilder().Build();
 
-    public Task InitializeAsync()
+    public async ValueTask InitializeAsync()
     {
-        return _arangoDbContainer.StartAsync();
+        await _arangoDbContainer.StartAsync();
     }
 
-    public Task DisposeAsync()
+    public ValueTask DisposeAsync()
     {
-        return _arangoDbContainer.DisposeAsync().AsTask();
+        return _arangoDbContainer.DisposeAsync();
     }
 
     [Fact]
@@ -26,7 +26,7 @@ public sealed class ArangoDbContainerTest : IAsyncLifetime
         using var client = new ArangoDBClient(transport);
 
         // When
-        var response = await client.Database.GetDatabasesAsync()
+        var response = await client.Database.GetDatabasesAsync(TestContext.Current.CancellationToken)
             .ConfigureAwait(true);
 
         // Then
--- a/tests/Testcontainers.Tests/Unit/Configurations/WaitUntilMessageIsLoggedTest.cs
+++ b/tests/Testcontainers.Tests/Unit/Configurations/WaitUntilMessageIsLoggedTest.cs
@@ -26,7 +26,8 @@ namespace DotNet.Testcontainers.Tests.Unit
 
     public async ValueTask InitializeAsync()
     {
-      await _container.StartAsync(_cts.Token);
+      await _container.StartAsync(_cts.Token)
+        .ConfigureAwait(false);
     }
 
     public ValueTask DisposeAsync()
--- a/tests/Testcontainers.Tests/Unit/Configurations/WaitUntilMessageIsLoggedTest.cs
+++ b/tests/Testcontainers.Tests/Unit/Configurations/WaitUntilMessageIsLoggedTest.cs
@@ -24,14 +24,14 @@ namespace DotNet.Testcontainers.Tests.Unit
         .Build();
     }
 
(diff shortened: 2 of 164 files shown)


tests/Testcontainers.Tests/Unit/Configurations/WaitUntilMessageIsLoggedTest.cs
```diff
@@ -24,14 +24,15 @@ public WaitUntilMessageIsLoggedTest()
...
-      return _container.StartAsync(_cts.Token);
+      await _container.StartAsync(_cts.Token);
     }
 
-    public Task DisposeAsync()
+    public async ValueTask DisposeAsync()
     {
-      return _container.StartAsync();
+      // TODO: Should this really be StartAsync?
+      await _container.StartAsync();
```
TheConstructor: This really feels like it should call `StopAsync` or `DisposeAsync`

tests/Testcontainers.ArangoDb.Tests/ArangoDbContainerTest.cs
```diff
@@ -26,7 +26,7 @@ public async Task RetrievesDatabases()
         using var client = new ArangoDBClient(transport);
 
         // When
-        var response = await client.Database.GetDatabasesAsync()
+        var response = await client.Database.GetDatabasesAsync(TestContext.Current.CancellationToken)
```
HofmeisterAn, line 30: Does it make sense to pass the cancellation token to every call if we're never actually canceling the tests? Is there any benefit doing that (I haven't looked into the v3 features and changes yet)?

tests/Testcontainers.Tests/Unit/Configurations/WaitUntilMessageIsLoggedTest.cs
HofmeisterAn: 👍 Yes, this is a mistake.

HofmeisterAn (commented):
Indeed, if there's no major blocker, it's a good idea to switch to v3. Thanks.

tests/Testcontainers.ArangoDb.Tests/ArangoDbContainerTest.cs
TheConstructor: While it fixes https://xunit.net/xunit.analyzers/rules/xUnit1051 it probably is only of use, when you actually cancel a run or set a timeout for a test. As you cancel all running jobs after the first test-failure, it should be beneficial in the pipeline, and probably even more so locally, when you realise that you hit "run tests" just that moment to early. If you object, I can remove the token and suppress the analyser.

tests/Testcontainers.Tests/Unit/Configurations/WaitUntilMessageIsLoggedTest.cs
TheConstructor: Changed it to `DisposeAsync`

HofmeisterAn: Thanks

tests/Testcontainers.ArangoDb.Tests/ArangoDbContainerTest.cs
HofmeisterAn: Thanks for sharing. I'll try to finish the review by the end of this week.

HofmeisterAn (approved):

HofmeisterAn:
Thanks 🙇.

0xced:
This pull request is a nightmare for Rider users, see [RIDER-121579 - Duplication of tests](https://youtrack.jetbrains.com/issue/RIDER-121579). 😕

I'll try to submit a follow-up pull request to address this.

TheConstructor:
> This pull request is a nightmare for Rider users, see [RIDER-121579 - Duplication of tests](https://youtrack.jetbrains.com/issue/RIDER-121579). 😕
> 
> I'll try to submit a follow-up pull request to address this.

Interesting. I use Rider and did not experience this while authoring the PR or in other projects. Not sure if this means it is only related to v3 > 2.0.1, but then again I use the same XUnit version at work with Rider

0xced:
> I use Rider and did not experience this while authoring the PR or in other projects.

That's interesting! Have you enabled _Testing Platform support_? This issue arises with xUnit.net v3 when both _Testing Platform_ and _VSTest adapters_ support are enabled and the test project has package references to both `Microsoft.NET.Test.Sdk` and `xunit.runner.visualstudio`. At least this is my understanding of the current situation as of JetBrains Rider 2025.1.2 and xUnit v3 2.0.2.

image

TheConstructor:
> > I use Rider and did not experience this while authoring the PR or in other projects.
> 
> Have you enabled _Testing Platform support_? This issue arises with xUnit.net v3 when both _Testing Platform_ and _VSTest adapters_ support are enabled and the test project has package references to both `Microsoft.NET.Test.Sdk` and `xunit.runner.visualstudio`.

At least my work-Rider has "Enable Testing Platform support" disabled - maybe that's the reason it's not happening to me. I don't think I disabled it intentionally, though

HofmeisterAn:
I don't have that problem either (tested on two different machines). "Enable Testing Platform support" is disabled too.

HofmeisterAn:
I just refactored a test and added `IXunitSerializable`, and I noticed another issue. The Rider test explorer isn't showing all the test cases, one of them is missing. In my example, the test explorer shows 3 tests, but the CLI (`dotnet test`) runs 4 (which is correct). If I intentionally break the missing test, Rider marks the overall test red, but the specific test case still doesn't appear in the test explorer 😐.

Issue #1262, opened 2024-09-18
[Enhancement]: Single database per test.


PureKrome:
### Problem

First linked/referenced in #1165 
Shoutout to @0xced 

Context: MS-SQL containers + xUnit / testing

One of the frustrating things with running DB tests and test containers is that all the DB tests are synchronous. This is via the Collection which we run all the tests under.

This raises some issues
- slowness: would be nicer to have a multiple DB tests running at the same time
- data isolation: each test should have it's own data so it doesn't conflict/fight with other tests.
- data changed: test 1 might do something to the db data, while test 2 then expects the data to be in a fresh state, but it's modified.

a quick fix to all of this is a single container _per test method_ but that is resource expensive 😢 
Resetting the Db back after each test still means we're stuck with synchronous tests.

### Solution

Would be really lovely would be to have a mix of both!
- 🐳 single container 
- 💾 single db per test method

(I believe this is what 🐦‍⬛ RavenDb does?)

In the context of MS-SQL this could be achieved via the connection string `Initial Catalog` key/value.

So maybe this means
- we don't use the `[Collection()]` attribute (which forces all the tests to run in synchronous mode)
- when the test _run_ first starts (is this called the test app?), we create the single container
- when each test method runs, then we can set the initial catalogue to be unique.

For example - here's two classes to try and set this up. I'm not sure how close this is to #1165 PR code:

```
// Simple fixture which is ran once at start when the test run first runs..

public class SqlServerFixture : IAsyncLifetime
{
    private readonly MsSqlContainer _msSqlContainer;

    public SqlServerFixture() =>
        _msSqlContainer = new MsSqlBuilder()
            .WithImage("mcr.microsoft.com/mssql/server:2022-CU13-ubuntu-22.04")
            //.WithReuse(true)
            .Build();

    public string ConnectionString => _msSqlContainer.GetConnectionString();

    public async Task InitializeAsync()
    {
        await _msSqlContainer.StartAsync();
    }

    public async Task DisposeAsync()
    {
        if (_msSqlContainer != null)
        {
            await _msSqlContainer.StopAsync();
        }
    }
}

// Sample 'DB' base class for tests

public abstract class BaseSqlServerTest(
    SqlServerFixture _sqlServerFixture,
    ITestOutputHelper _testOutputHelper) : IClassFixture<SqlServerFixture>, IAsyncLifetime
{
    protected string ConnectionString { get; private set; }

    public async Task InitializeAsync()
    {
[...]

PR #1442, opened 2025-04-30
docs: Mention the xUnit.net v3 module (package)

TheConstructor:
## What does this PR do?

This PR adds a reference to `Testcontainers.XunitV3` to the xUnit.net-documentation page so starting out to use Testcontainers with xUnit.net v3 becomes easier.

## Why is it important?

While not hidden, the xUnit.net v3 package is not mentioned, and the documentation currently sound like `Testcontainers.Xunit` could be the right choice no matter the xUnit.net-version.

### DIFF
--- a/docs/test_frameworks/xunit_net.md
+++ b/docs/test_frameworks/xunit_net.md
@@ -1,8 +1,14 @@
 # Testing with xUnit.net
 
-The [Testcontainers.Xunit](https://www.nuget.org/packages/Testcontainers.Xunit) package simplifies writing tests with containers in [xUnit.net](https://xunit.net). By leveraging xUnit.net's [shared context](https://xunit.net/docs/shared-context), this package automates the setup and teardown of test resources, creating and disposing of containers as needed. This reduces repetitive code and avoids common patterns that developers would otherwise need to implement repeatedly.
+The [Testcontainers.Xunit](https://www.nuget.org/packages/Testcontainers.Xunit) and [Testcontainers.XunitV3](https://www.nuget.org/packages/Testcontainers.XunitV3) packages simplify writing tests with containers in [xUnit.net](https://xunit.net). By leveraging xUnit.net's [shared context](https://xunit.net/docs/shared-context), this package automates the setup and teardown of test resources, creating and disposing of containers as needed. This reduces repetitive code and avoids common patterns that developers would otherwise need to implement repeatedly.
 
-To get started, add the following dependency to your project file:
+To get started, add the following dependency to your project file for xUnit.net v3:
+
+```shell title="NuGet"
+dotnet add package Testcontainers.XunitV3
+```
+
+Or, if you're using xUnit.net v2:
 
 ```shell title="NuGet"
 dotnet add package Testcontainers.Xunit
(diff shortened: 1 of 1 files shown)


docs/test_frameworks/xunit_net.md
```diff
@@ -43,7 +49,7 @@ be115f3df138   redis:7.0                   "docker-entrypoint.s…"   3 seconds
 
 ## Creating a shared test context
 
-Sometimes, creating and disposing of a test resource can be an expensive operation that you do not want to repeat for every test. By inheriting from the `ContainerFixture<TBuilderEntity, TContainerEntity>` class, you can share the test resource instance across all tests within the same test class.
+Sometimes, creating and disposing of a test resource can be an expensive operation that you do not want to repeat for every test. By inheriting from the `ContainerFixture<TBuilderEntity, TContainerEntity>` class, you can share the test resource instance across all tests within the same test class, or even within the assembly, if you use xUnit.net V3.
```
HofmeisterAn:
```suggestion
Sometimes, creating and disposing of a test resource can be an expensive operation that you do not want to repeat for every test. By inheriting from the `ContainerFixture<TBuilderEntity, TContainerEntity>` class, you can share the test resource instance across all tests within the same test class, or even within the assembly.
```

Doesn't it work with v2 as well (using the [assembly annotation](https://github.com/testcontainers/testcontainers-dotnet/issues/1262#issuecomment-2817328066))?

docs/test_frameworks/xunit_net.md
```diff
@@ -1,8 +1,14 @@
...
+The [Testcontainers.Xunit](https://www.nuget.org/packages/Testcontainers.Xunit) and [Testcontainers.XunitV3](https://www.nuget.org/packages/Testcontainers.XunitV3) packages simplify writing tests with containers in [xUnit.net](https://xunit.net). By leveraging xUnit.net's [shared context](https://xunit.net/docs/shared-context), this package automates the setup and teardown of test resources, creating and disposing of containers as needed. This reduces repetitive code and avoids common patterns that developers would otherwise need to implement repeatedly.
 
-To get started, add the following dependency to your project file:
+To get started with xUnit.net v3, add the following dependency to your project file:
+
+```shell title="NuGet"
+dotnet add package Testcontainers.XunitV3
+```
+
+If you use xUnit.net v2, add the following dependency to your project file:
```
HofmeisterAn:
~~~suggestion
To get started, add the following dependency to your project file for xUnit.net v3:

```shell title="NuGet"
dotnet add package Testcontainers.XunitV3
```

Or, if you're using xUnit.net v2:
~~~

docs/test_frameworks/xunit_net.md
```diff
@@ -75,6 +81,8 @@ d29a393816ce   redis:7.0                   "docker-entrypoint.s…"   3 seconds
 e878f0b8f4bc   testcontainers/ryuk:0.9.0   "/bin/ryuk"              3 seconds ago
 ```
 
+To use a single container-fixture throughout all tests of a xUnit v3 test-assembly, you can use `[assembly: AssemblyFixture(typeof(YourContainerFixture))]`. The instance can then be used as a constructor-parameter like above, or be accessed through `TestContext.Current.GetFixture` in all tests of said test-assembly.
```
HofmeisterAn:
```suggestion
To use a single container fixture across all tests in xUnit.net, add the `[assembly: AssemblyFixture(typeof(TContainerFixtureEntity))]` annotation to your fixture class. The fixture can then be injected into test class constructors (as shown above), or accessed using the `TestContext.Current.GetFixture<TContainerFixtureEntity>()` method (v3 only) in any test.
```

HofmeisterAn (changes requested):
Thanks for the PR. Good idea! What do you think about the suggestions?

docs/test_frameworks/xunit_net.md
TheConstructor: The second paragraph of https://xunit.net/docs/shared-context#assembly-fixture explicitly states it is new in v3, but there is https://github.com/xunit/samples.xunit/tree/main/v2/AssemblyFixtureExample which shows how to build a custom runner to support it in v2. The experience is IMHO not as good as it is in v3 though.

Are you still in favour of removing `if you use xUnit.net V3`? We could also go `if you use xUnit.net V3 or an appropriate extension`

docs/test_frameworks/xunit_net.md
TheConstructor: `all tests in xUnit.net` sounds wrong to me, as it is your assembly building on top of xUnit.net. This also does feel related to how we decide in https://github.com/testcontainers/testcontainers-dotnet/pull/1442/files#r2070470394

`annotation to your fixture class` is ok with me, but not a 100% accurate. It is an assembly-annoataion refering to the fixture.

TheConstructor:
@HofmeisterAn I pushed a new version. I copied your suggestion for the add-package-section and variations of the other two

docs/test_frameworks/xunit_net.md
HofmeisterAn: Ah, I think I mixed it up with collection fixtures. Thanks.

HofmeisterAn (approved):
Thanks

PR #1446, opened 2025-05-11
feat: Add Typesense module

brainded:
## What does this PR do?

Adds Typesense testcontainers implementation.

## Why is it important?

No Typesense Module currently available for Dotnet.

## Related issues

No existing issues found related to Typesense.

## How to test this PR

tests/Testcontainers.Typesense.Tests contains two tests:
- Call Healthy endpoint to verify container up and running
- Call Collection endpoint which will receive an empty json array


brainded:
Not sure why Pulsar failed, I don't believe I impacted that?

HofmeisterAn:
> Not sure why Pulsar failed, I don't believe I impacted that?

No, the PR isn't related. It's Pulsar... It seems like the consumer sometimes isn't in a working state. I'll try to review it in the next days.

src/Testcontainers.Typesense/TypesenseBuilder.cs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /// <summary>
+    /// Enables Cors.
+    /// </summary>
+    /// <returns>A configured instance of <see cref="TypesenseBuilder" />.</returns>
+    public TypesenseBuilder EnableCors(bool enableCors)
+    {
+        return Merge(DockerResourceConfiguration, new TypesenseConfiguration(enableCors: enableCors))
+          .WithCommand("--enable-cors");
```
TheConstructor: Shouldn't this check if `enableCors == true`? Can CORS be disabled with this definition?

src/Testcontainers.Typesense/TypesenseConfiguration.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        // Passes the configuration upwards to the base implementations to create an updated immutable copy.
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TypesenseConfiguration" /> class.
+    /// </summary>
+    /// <param name="oldValue">The old Docker resource configuration.</param>
+    /// <param name="newValue">The new Docker resource configuration.</param>
+    public TypesenseConfiguration(TypesenseConfiguration oldValue, TypesenseConfiguration newValue)
```
TheConstructor, line 49: Doesn't this constructor need to take care of `EnableCors` and `Volume` as well?

TheConstructor (commented):
I just had a very brief look at your code and added two questions

src/Testcontainers.Typesense/TypesenseBuilder.cs
brainded: Great catch, initially I wasn't going to give someone the option, need to finish this.

src/Testcontainers.Typesense/TypesenseBuilder.cs
```diff
@@ -0,0 +1,122 @@
...
+    protected override TypesenseBuilder Init()
+    {
+        return base.Init()
+            .WithImage(TypesenseImage)
+            .WithPortBinding(DefaultPort, true)
+            .WithApiKey(DefaultApiKey)
+            .EnableCors(DefaultCors)
+            .WithTmpfsMount(DefaultVolume)
+            .WithDataDirectory(DefaultVolume)
+            .WithWaitStrategy(Wait.ForUnixContainer().UntilMessageIsLogged("Peer refresh succeeded"));
```
eddumelendez: Typesense offers a health endpoint. You can use the http strategy instead. See [java implementation](https://github.com/testcontainers/testcontainers-java/blob/main/modules/typesense/src/main/java/org/testcontainers/typesense/TypesenseContainer.java#L33-L37)

HofmeisterAn (commented):
I've updated the PR to follow the repository standards.

We only create container builder APIs for settings that are absolutely necessary to start the container. If a setting can be configured with methods like `WithCommand(string)`, we avoid adding separate APIs to keep it simple.

I removed the `--cors` flag. When it's always set, developers can't disable it later.

The container configuration type only includes properties needed to modify the setup in `Build()` or needed in the container type.

LMKWYT. Happy to merge the PR if you're okay with the changes.

brainded:
@HofmeisterAn Yeah, thank you for doing that. Been quite busy lately and just getting back to it now. The changes look good to me. Thanks again, I really appreciate it!

HofmeisterAn (approved):
> @HofmeisterAn Yeah, thank you for doing that. Been quite busy lately and just getting back to it now. The changes look good to me. Thanks again, I really appreciate it!

Oh, no worries. Thanks for the contribution!

PR #1454, opened 2025-05-25
feat: Add Toxiproxy module

iltertaha:
## What does this PR do?

Adds a new Testcontainers.Toxiproxy module that enables integration testing with [Shopify's Toxiproxy](https://github.com/Shopify/toxiproxy).
It provides a ToxiproxyContainer and a ToxiproxyBuilder to simulate various network conditions (latency, timeout, bandwidth limit, etc.) within containerized test environments.

This module uses the [Toxiproxy.Net](https://github.com/mdevilliers/Toxiproxy.Net) client library to interact with the Toxiproxy HTTP API.

Included tests cover basic proxy creation, toxic addition, and runtime behavior of network faults like latency and timeouts.

## Why is it important?

There is currently no native Toxiproxy module in testcontainers-dotnet.
Toxiproxy is a powerful tool for simulating real-world network issues such as slow connections, disconnections, or dropped packets.

## How to test this PR. 
 
`dotnet test tests/Testcontainers.Toxiproxy.Tests`


HofmeisterAn (approved):
Finally, I had time to look into this PR, something I've been looking forward to for a while. One thing I noticed is that the `Toxiproxy` module was implemented with a dependency on `Toxiproxy.Net`. We generally don't include client dependencies in our modules, so I removed it to align the module with our standards and updated the tests and docs accordingly.

I also added an example that matches the [Testcontainers for Java](https://github.com/testcontainers/testcontainers-java/blob/73726f436fb469469986866ab55f67057d0beba4/modules/toxiproxy/src/test/java/org/testcontainers/toxiproxy/ToxiproxyContainerTest.java) implementation. It would be great to add the other tests as examples too. If you think anything is missing, please feel free to create an issue so we can discuss the changes or missing parts.

Thanks for the initial work on this! BTW awesome that you're in F1 😎. I'm a big fan, just the wrong team 😬.

PR #1459, opened 2025-06-10
Update NatsBuilder.cs

Onno-H:
## What does this PR do?

Update to the latest version of NATS due to the following CVE: https://nvd.nist.gov/vuln/detail/CVE-2025-30215

## Why is it important?

CVE bad.

## Related issues
N/A
-

### DIFF
--- a/src/Testcontainers.Nats/NatsBuilder.cs
+++ b/src/Testcontainers.Nats/NatsBuilder.cs
@@ -4,7 +4,7 @@ namespace Testcontainers.Nats;
 [PublicAPI]
 public sealed class NatsBuilder : ContainerBuilder<NatsBuilder, NatsContainer, NatsConfiguration>
 {
-    public const string NatsImage = "nats:2.9";
+    public const string NatsImage = "nats:2.11";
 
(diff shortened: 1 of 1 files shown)


HofmeisterAn (commented):
I'm not sure if we should merge this or not. Thanks a lot for the contribution though. Usually, we don't update the default images to avoid breaking things. We only update them if something stops working and even then we try to keep it compatible with older versions. We [recommend](https://dotnet.testcontainers.org/api/best_practices/) developers to change and pin the image version themselves.

Maybe we should do what Java does and not have a default version at all, and make it required to set the image version (I will create a discussion for this: https://github.com/testcontainers/testcontainers-dotnet/discussions/1470).

HofmeisterAn:
Thanks again for the PR. As mentioned earlier, I'm not sure how to proceed. I'm favoring dropping the default image version in general and using an approach that aligns better with other TC language implementations: https://github.com/testcontainers/testcontainers-dotnet/discussions/1470. I'll create an issue.

**Edit**: If you have other opinions, let me know. Feedback is much appreciated.

Issue #1486, opened 2025-07-17
[Bug]: Copy a file WithResourceMapping leads to an (empty) directory in the container


cimnine:
### Testcontainers version

4.6.0

### Using the latest Testcontainers version?

Yes

### Host OS

Mac

### Host arch

ARM

### .NET version

9.0.302

### Docker version

```shell
➜ podman version
Client:        Podman Engine
Version:       5.5.2
API Version:   5.5.2
Go Version:    go1.24.4
Built:         Tue Jun 24 16:41:45 2025
Build Origin:  brew
OS/Arch:       darwin/arm64

Server:       Podman Engine
Version:      5.2.3
API Version:  5.2.3
Go Version:   go1.22.7
Built:        Tue Sep 24 02:00:00 2024
OS/Arch:      linux/arm64
```

### Docker info

```shell
➜ podman info
Client:
  APIVersion: 5.5.2
  BuildOrigin: brew
  Built: 1750776105
  BuiltTime: Tue Jun 24 16:41:45 2025
  GitCommit: ""
  GoVersion: go1.24.4
  Os: darwin
  OsArch: darwin/arm64
  Version: 5.5.2
host:
  arch: arm64
  buildahVersion: 1.37.3
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.12-2.fc40.aarch64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.12, commit: '
  cpuUtilization:
    idlePercent: 98.24
    systemPercent: 0.73
    userPercent: 1.03
  cpus: 6
  databaseBackend: sqlite
  distribution:
    distribution: fedora
    variant: coreos
    version: "40"
  eventLogger: journald
  freeLocks: 2031
  hostname: localhost.localdomain
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 6.10.10-200.fc40.aarch64
  linkmode: dynamic
  logDriver: journald
  memFree: 842330112
  memTotal: 7716700160
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: aardvark-dns-1.12.1-1.20240925084542154505.main.19.g56d105f.fc40.aarch64
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.13.0-dev
[...]

PR #1487, opened 2025-07-17
docs: Extend example copying files to a container

cimnine:
## What does this PR do?

This change should make it more explicit that, when providing a string, the target is always a directory and never a file path. To specify a file path, a instance of `FileInfo` must be passed.

## Why is it important?

When coming from `docker-compose.yml`, it is unexpected that the second argument of `WithResourceMapping` is always a path to a directory, never a file.

Given `.WithResourceMapping(new FileInfo("file.txt"), "/file.txt");`, my expectation was a file at `/file.txt`, but it ends up at `/file.txt/file.txt` in the container.

## Related issues

Relates to #1486

### DIFF
--- a/docs/api/create_docker_container.md
+++ b/docs/api/create_docker_container.md
@@ -45,7 +45,11 @@ _ = new ContainerBuilder()
 
 ```csharp title="Copying a file"
 _ = new ContainerBuilder()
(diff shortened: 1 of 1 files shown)


HofmeisterAn (approved):
Thanks

PR #1508, opened 2025-08-08
feat: Add Valkey module support

thomhurst:
Add Testcontainers module for Valkey, a Redis-compatible in-memory data store. Includes ValkeyContainer implementation with connection string support and tests.


HofmeisterAn (commented):
Thanks for the PR. I'll close this in favor of: #1641.

The Redis module should be compatible with the container image. I'd like to reduce the number of copy-paste modules and the associated maintenance work. If Valkey introduces features or configuration options that require dedicated support in the future, we can add a separate module at that time.

PR #1513, opened 2025-08-13
Add csharpier formatter

### LINKED ISSUE #1501: [Enhancement]: csharpier for formatting
### Problem

When I was working on the code, my IDE formatted the code in a way that didn't match the repository's style.

### Solution

Format the code with [csharpier](https://github.com/belav/csharpier) and automatically check the formatting via a GitHub Action.

### Benefit

- It would no longer be a manual task to check if the formatting is correct.
- It would level opinions about the formatting (_It's the way the formatter does it and that's the way it is_.)

### Alternatives

Another formatter, e.g. `dotnet format`. But that one isn't as rigorous, and it's also slow.

Yes

cimnine:
## What does this PR do?

Adds csharpier.

## Why is it important?

* It will no longer be a manual task to check if the formatting is correct.
* It will level opinions about the formatting (it's the way the formatter does it and that's the way it is.)

## Related issues

- Closes #1501 

## How to test this PR

There are two commits in this MR:
- First commit adds csharpier and infrastructure (small change)
- Second commit applies `dotnet csharpier format .` for the first time (big change)


HofmeisterAn (commented):
Thanks for the PR! Could we split this into smaller steps? Maybe start by adding the CSharpier config/tooling (disabled at first), then enable it gradually across the repo. WDYT?

Also, can we hook it into Cake so we can run a task to check formatting? Since our CI is built on Cake, that would fit perfectly.

cimnine:
> Could we split this into smaller steps? Maybe start by adding the CSharpier config/tooling (disabled at first), then enable it gradually across the repo. WDYT?

What are the advantages for choosing this route?

> Also, can we hook it into Cake so we can run a task to check formatting? Since our CI is built on Cake, that would fit perfectly.

What are the advantages of the formatter running for every project independently? A typical check across the whole project takes below 1s (after the first run). It's not significantly faster if run for each project independently.

Also, if it's integrated into the cake process, you won't get feedback in the PR. The GitHub Action can do that.

HofmeisterAn:
> What are the advantages for choosing this route?

It's easier to review and to understand/see the changes. I can also take a look at the two commits and see how far I get. I mean, the first one is trivial.

> What are the advantages of the formatter running for every project independently?

Oh, I didn't mean it should be done independently. I was just thinking of having a Cake task so we can run it the same way locally and in CI. That's already been helpful with other tooling. I just don't know much about CSharpier and how it works (its tooling around).

PR #1515, opened 2025-08-16
fix: Dispose StringWriter

HofmeisterAn:
## What does this PR do?

While reviewing the recent PR that adds KRaft support to the Kafka module, I noticed that we weren't disposing of the `StringWriter`. The PR fixes this and disposes of it now.

## Why is it important?

\-

## Related issues

\-


HofmeisterAn:
According to the docs, disposing it seams not necessary.

PR #1543, opened 2025-10-03
chore: Collect test projects at CI runtime

HofmeisterAn:
## What does this PR do?

This PR adds support for collecting test projects during CI runtime and creating the strategy on the fly. It automatically detects new test projects and adjusts the strategy as needed. For example, if only certain modules change, we don't have to build or test unrelated modules. It also lets us filter out unnecessary tests, reducing CI time, saving resources, and speeding up feedback.

This PR doesn't include the logic to filter out unrelated test projects yet. We'll add that in a separate PR. For now, projects just pass through the filter script /cc @digital88 WDYT?.

## Why is it important?

\-

## Related issues

- Relates #1136
- Relates #1417


digital88:
> WDYT?

Yes, makes sense. GH actions is a nightmare to debug )

HofmeisterAn:
> > WDYT?
> 
> Yes, makes sense. GH actions is a nightmare to debug )

Luckily, we can test this pretty easily locally too.

PR #1549, opened 2025-10-11
fix(Papercut): Pin version 7.0 and set new HTTP and SMTP ports

HofmeisterAn:
## What does this PR do?

Earlier versions of Papercut didn't use tags, so we had to rely on the `latest` tag. With version 7, Papercut introduced a breaking change and updated the ports for the HTTP and SMTP services. This PR pins Papercut to version 7 and updates the ports accordingly.

## Why is it important?

The module needs this change to start properly. Without it, the wait strategy never finishes because it cannot connect to the services.

## Related issues

\-

### DIFF
--- a/src/Testcontainers.Papercut/PapercutBuilder.cs
+++ b/src/Testcontainers.Papercut/PapercutBuilder.cs
@@ -4,11 +4,11 @@ namespace Testcontainers.Papercut;
 [PublicAPI]
 public sealed class PapercutBuilder : ContainerBuilder<PapercutBuilder, PapercutContainer, PapercutConfiguration>
 {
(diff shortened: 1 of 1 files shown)


bwets:
@HofmeisterAn Hi, would you be able to tell us when a release will be done with this fix, with nuget packages? Thanks a lot, this is stalling all our build pipelines...

HofmeisterAn:
> @HofmeisterAn Hi, would you be able to tell us when a release will be done with this fix, with nuget packages? Thanks a lot, this is stalling all our build pipelines...

No ETA. You can override the default module configuration to workaround the issue.

PR #1451, opened 2025-05-20
fix: Throw an exception if the container has exited while checking for readiness

0xced:
## What does this PR do?

This pull request ensures that if a container has exited, readiness check are aborted with the newly introduced `ContainerException`. The exception message includes both stdout and stderr of the container to help diagnosing the issue.

## Why is it important?

As described in #1445, this change significantly improves the developer experience. Developers get immediate feedback when something goes wrong.

## Related issues

Fixes #1445

## How to test this PR

A simple way to test this PR is to run the Oracle21 tests on an Apple Silicon Mac. You'll get this error.

```
DotNet.Testcontainers.Configurations.ContainerException
The gvenzl/oracle-xe:21-slim-faststart container has exited.

=== stdout of sad_noyce (63dc2f99ee704a934636f47d2372d0283697ec6731af1f9ffad564e9e34a1f0a) ===
2025-05-20T08:32:31.693685128Z CONTAINER: starting up...
2025-05-20T08:32:31.769256852Z CONTAINER: first database startup, initializing...
2025-05-20T08:32:31.904095582Z CONTAINER: starting up Oracle Database...
2025-05-20T08:32:31.945699749Z 
2025-05-20T08:32:31.945714499Z LSNRCTL for Linux: Version 21.0.0.0.0 - Production on 20-MAY-2025 08:32:31
2025-05-20T08:32:31.945716291Z 
2025-05-20T08:32:31.945717499Z Copyright (c) 1991, 2021, Oracle.  All rights reserved.
2025-05-20T08:32:31.945718624Z 
2025-05-20T08:32:31.959288638Z Starting /opt/oracle/product/21c/dbhomeXE/bin/tnslsnr: please wait...
2025-05-20T08:32:31.959295388Z 
2025-05-20T08:32:32.411257859Z TNSLSNR for Linux: Version 21.0.0.0.0 - Production
2025-05-20T08:32:32.411270818Z System parameter file is /opt/oracle/homes/OraDBHome21cXE/network/admin/listener.ora
2025-05-20T08:32:32.411272401Z Log messages written to /opt/oracle/diag/tnslsnr/63dc2f99ee70/listener/alert/log.xml
2025-05-20T08:32:32.411273526Z Listening on: (DESCRIPTION=(ADDRESS=(PROTOCOL=ipc)(KEY=EXTPROC_FOR_XE)))
[...]

PR #1550, opened 2025-10-13
feat: Throw if container not running

### LINKED ISSUE #1445: [Enhancement]: Wait strategies should detect when a container exits early and throw an exception
### Problem

If a container stops or exits during startup, most wait strategies will keep retrying to detect readiness, even though it will never succeed. But if the container has already stopped or exited, we know there's no reason retrying it. Instead, we should immediately throw an exception.

### Solution

Ideally, we should also try to get the container logs and include them in the exception to help with debugging.

### Benefit

This change would significantly improve the developer experience. Developers would get immediate feedback when something goes wrong.

It would also improve our CI process. ~~For example, the Pulsar module test sometimes fails because a container exits during startup, but we don't know why. From running tests in GitHub Codespaces (Dev Containers), it looks like a memory issue, most likely the CI runner is running OOM~~ (the issue has been fixed in the meantime).

### Alternatives

\-

Yes

HofmeisterAn:
## What does this PR do?

This PR introduces wait strategy modes. Wait strategy modes define how Testcontainers for .NET handles wait strategies. Normally, a wait strategy indicates when a container (and the service inside it) is ready to use. These strategies assume the container stays running.

However, sometimes a container fails to start and exits during startup. In those cases, Testcontainers kept retrying the wait strategy indefinitely, never indicating readiness. From a developer's point of view, it looks like the container startup is just hanging.

This PR updates the internal behavior of wait strategies. They now expect the container to be running. If the container exits unexpectedly, Testcontainers for .NET will collect the exit code and logs, then throw a meaningful exception: `ContainerNotRunningException`, which includes that information.

Some scenarios do not require the container to stay running. For example, running database migrations in a sidecar container during startup. In those cases, it is fine if the container exits after finishing its work. To support this, Testcontainers for .NET now provides wait strategy modes.

If you do not want an exception when a container exits unexpectedly, set the wait strategy mode to `OneShot`.

**TODOs**

- [x] Add integration tests.
- [x] Add docs.

## Why is it important?

\-

## Related issues

- Closes #1445
- Closes #1451

### DIFF
--- a/docs/api/wait_strategies.md
+++ b/docs/api/wait_strategies.md
@@ -20,6 +20,17 @@ _ = Wait.ForUnixContainer()
 
 Besides configuring the wait strategy, cancelling a container start can always be done utilizing a [CancellationToken](create_docker_container.md#canceling-a-container-start).
 
+## Wait strategy modes
+
+Wait strategy modes define how Testcontainers for .NET handles container readiness checks. By default, wait strategies assume the container remains running throughout the startup. If a container exits unexpectedly during startup, Testcontainers for .NET will throw a `ContainerNotRunningException` containing the exit code and logs.
+
+Some containers are intended to stop after completing short-lived tasks like migrations or setup scripts. In these cases, the container exit is expected, not a failure. Use `WaitStrategyMode.OneShot` to treat a normal exit as successful rather than throwing an exception.
+
+```csharp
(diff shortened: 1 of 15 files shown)


0xced:
Thanks for taking the time to address this issue, Andre!

PR #1566, opened 2025-11-06
fix: Split ALL_CHANGED_FILES on any whitespace

HofmeisterAn:
## What does this PR do?

This PR fixes an issue in the incremental build and test setup. The GH Action step that collects all changed files doesn't use `\n` as a separator. It uses spaces. I updated the split implementation to split on any whitespace.

Previously, the `allChangedFiles` variable only contained a single string with all changed files, which caused the wrong test configurations to run. @digital88, we should double-check whether filenames with spaces might break this. According to the docs, the [GH Action](https://github.com/tj-actions/changed-files) step lets you configure the separator, and the default is supposed to be `\n`, but that doesn't seem to be the case (probably I misunderstood the docs).

## Why is it important?

To select and run the appropriate test projects.

## Related issues

\-

### DIFF
--- a/.github/scripts/Filter-TestProjects.ps1
+++ b/.github/scripts/Filter-TestProjects.ps1
@@ -117,7 +117,7 @@ function Filter-TestProjects {
 
     Begin {
         $script:branch = $env:GITHUB_REF_NAME
-        $script:allChangedFiles = $env:ALL_CHANGED_FILES -Split "`n"
+        $script:allChangedFiles = $env:ALL_CHANGED_FILES -Split "[\s\t\n]"
         $script:filteredModules = @()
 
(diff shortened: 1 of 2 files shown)


digital88:
@HofmeisterAn 
~~Maybe something like `.Split([Environment]::NewLine)` would be 'safer'? WDYT? Or add `[Environment]::NewLine` as additional argument to array of `\s\t\n`.~~

I beleive this is what we need and it's indeed equals to `" "` (space) by default:
image

HofmeisterAn:
Yep, I noticed that too after going through the docs again. I even tried it, but I ran into another issue, which was probably a misconfiguration with the string split. I haven't had the time to look into it yet, so I just stuck with this quick fix.

PR #1567, opened 2025-11-09
fix: Set Kusto wait strategy encoding to UTF-8

MattKotsenas:
## What does this PR do?

The Kusto wait strategy uses the default encoding. For .NET Framework, this resolves to Windows-1252, which causes the Kusto emulator to return a 415 Media Type not Supported error. In .NET Core, it is UTF-8.

This allows .NET Framework tests to run by setting the encoding to always be UTF-8.

## Why is it important?

Required to use Testcontainers in a .NET Framework app. Otherwise, the Testcontainer waits forever.


HofmeisterAn (approved):
Thanks! That's something we should update across the whole repo.

Issue #1540, opened 2025-09-30
[Enhancement]: Force user provided image name/tag when creating Builder


digital88:
### Problem

As per discussion: https://github.com/testcontainers/testcontainers-dotnet/discussions/1470
Related: https://github.com/testcontainers/testcontainers-dotnet/discussions/1407

By default, Builders use hardcoded default image versions. Those versions deprecate quite rapidly, so many other TC implementations opt to force user to provide desiriable image tag explicitly when creating Builder instance.

### Solution

Mark current parameterless constructor as `[Obsolete]`. Introduce new `Builder(DockerImage)` constructor. Rewrite old constructor implementation to call this new constructor with default hardcoded image for compatibility.

### Benefit

No need to support outdated/insecure versions of docker images.

### Alternatives

Continue to support old images indefinitely.

### Would you like to help contributing this enhancement?

Yes

HofmeisterAn:
Thanks for creating this issue, it is a great first step in addressing the topic.

As you mentioned, other language implementations already follow this approach, so aligning .NET makes perfect sense. It also helps stabilize test environments through version pinning. Although, I am a little concerned some developers might misunderstand the decision or find it harder to get started with Testcontainers.

Here are some initial thoughts on what we need to cover in this issue (there could be more):

- Mark the default constructor as obsolete (with a message that points to the expected default repository).
- Each module should include XML documentation that references the expected default repository.
- For our module tests, we could use a shared `Dockerfile` that contains all required images, and then map them to our tests.
    - That way, we could use Dependabot to keep images updated, similar to how Testcontainers for Node does it.
    - We should discuss whether this is the best approach for us, or if there is a better workflow.

WDYT?

digital88:
Good points. I too like TC for Node approach with standalone dockerfile + dependabot.

I think we also need to define strategy what to do if the new docker image for module breaks compatibility, like what happened with kafka recently. For how long should previous version be supported or should it be supported at all?

HofmeisterAn:
> For how long should previous version be supported or should it be supported at all?

For now, I'd stick with the proposal that's part of the discussion. We should also add the information to the module documentation to make it clearer and more concrete.

HofmeisterAn:
## TODOs

- [x] Fix https://github.com/testcontainers/testcontainers-dotnet/issues/1598.
- [ ] Add workflow to update the `Dockerfile` (using Dependabot).
- [x] Prepare a draft PR that updates the docs.
[...]

PR #1584, opened 2025-11-18
feat: Require explicit container image when creating container builder

digital88:
## What does this PR do?

See https://github.com/testcontainers/testcontainers-dotnet/issues/1540

## Why is it important?

See https://github.com/testcontainers/testcontainers-dotnet/issues/1540


digital88:
Again random Elastic test failure. Need to merge https://github.com/testcontainers/testcontainers-dotnet/pull/1593 first I think.

digital88:
Small overview:

Added two new constructors to each module. `ctor(string)` and `ctor(IImage)`. They both call `Init().WithImage()` and pass the argument. Changed existing parameterless constructor to explicitly use old image like `Init().WithImage(MODULE_DEFAULT_IMAGE)`. The old parameterless constructor is deprecated.

Added xml docs to point where image tags can be located, usually `hub.docker.com`.

Added same two new constructors and deprecated old constructor in `ContainerBuilder`. I changed almost all internal references to one of the new constuctors.

Added `Dockerfile` to each test project and wrote a new `TestSession.GetImageFromDockerfile()` helper. I tried to add latest version of every module in this Dockerfile.
Not all modules currently support latest versions of their images, so I had to put non-latest working image tag for some of them.

HofmeisterAn:
Wow, this looks great! Thanks for your hard work. The only issue I noticed is with the xUnit.net [module](https://github.com/testcontainers/testcontainers-dotnet/blob/8aa6520dc6fda67a7d001285054f0dc0e8bf2b29/src/Testcontainers.Xunit/ContainerLifetime.cs#L19). It won't work anymore in the future. We probably need to use reflection here, even though I try to avoid it as much as possible. We need to think about whether there's another way. WDYT?

digital88:
> The only issue I noticed is with the xUnit.net [module](https://github.com/testcontainers/testcontainers-dotnet/blob/8aa6520dc6fda67a7d001285054f0dc0e8bf2b29/src/Testcontainers.Xunit/ContainerLifetime.cs#L19).

Yes, this is valid problem, I think we need to make a new issue for it, and make a separate PR. We could for example pass a factory or "creator" delegate if reflection is not an option. Or maybe even use `InternalsVisibleTo`, and make constructor internal instead of public 😄

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
```diff
@@ -17,10 +17,31 @@ public sealed class ArtemisBuilder : ContainerBuilder<ArtemisBuilder, ArtemisCon
     /// <summary>
     /// Initializes a new instance of the <see cref="ArtemisBuilder" /> class.
     /// </summary>
+    [Obsolete("Use constructor with image as a parameter instead.")]
     public ArtemisBuilder()
         : this(new ActiveMqConfiguration())
     {
-        DockerResourceConfiguration = Init().DockerResourceConfiguration;
+        DockerResourceConfiguration = Init().WithImage(ArtemisImage).DockerResourceConfiguration;
+    }
```
HofmeisterAn, line 25: We can replace this with:

```csharp
[Obsolete("Use constructor with image as a parameter instead.")]
public ArtemisBuilder()
    : this(ArtemisImage)
{
}
```

I think I can change this with a regex.

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
```diff
@@ -17,10 +17,31 @@ public sealed class ArtemisBuilder : ContainerBuilder<ArtemisBuilder, ArtemisCon
     /// <summary>
     /// Initializes a new instance of the <see cref="ArtemisBuilder" /> class.
     /// </summary>
+    [Obsolete("Use constructor with image as a parameter instead.")]
```
HofmeisterAn: I think we should add the GH issue link to the obsolete message and include more information about the migration (in the issue), e.g.:

    Use the constructor with the image argument instead: https://github.com/testcontainers/testcontainers-dotnet/issues/1540.

src/Testcontainers.ActiveMq/Testcontainers.ActiveMq.csproj
```diff
@@ -4,9 +4,9 @@
         <LangVersion>latest</LangVersion>
     </PropertyGroup>
     <ItemGroup>
-        <PackageReference Include="JetBrains.Annotations" VersionOverride="2023.3.0" PrivateAssets="All"/>
+        <PackageReference Include="JetBrains.Annotations" VersionOverride="2023.3.0" PrivateAssets="All" />
     </ItemGroup>
     <ItemGroup>
-        <ProjectReference Include="../Testcontainers/Testcontainers.csproj"/>
+        <ProjectReference Include="../Testcontainers/Testcontainers.csproj" />
     </ItemGroup>
```
HofmeisterAn, line 11: I believe these changes aren't necessary. Lets revert them with search and replace.

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
HofmeisterAn: Somehow the files got converted to UTF-8 with BOM (probably your IDE). Can you convert them back? I can do it too, I have a script.

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
```diff
@@ -7,17 +7,22 @@
...
-        <ProjectReference Include="../Testcontainers.Commons/Testcontainers.Commons.csproj"/>
+        <ProjectReference Include="../../src/Testcontainers.Toxiproxy/Testcontainers.Toxiproxy.csproj" />
+        <ProjectReference Include="../../src/Testcontainers.Redis/Testcontainers.Redis.csproj" />
+        <ProjectReference Include="../Testcontainers.Commons/Testcontainers.Commons.csproj" />
+    </ItemGroup>
+    <ItemGroup>
+      <None Update="Dockerfile">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      </None>
     </ItemGroup>
```
HofmeisterAn, line 27: Inconsistent indentation. The modules use 4 spaces. I can update them too.

tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs
```diff
@@ -120,7 +120,7 @@ public MongoDbV5Configuration()
     public sealed class MongoDbV4Configuration : MongoDbContainerTest
     {
         public MongoDbV4Configuration()
-            : base(new MongoDbBuilder().WithImage("mongo:4.4").Build(), true /* Replica set status returns "ok" in MongoDB 4.4 without initialization. */)
+            : base(new MongoDbBuilder(TestSession.GetImageFromDockerfile()).WithImage("mongo:4.4").Build(), true /* Replica set status returns "ok" in MongoDB 4.4 without initialization. */)
         {
         }
     }
```
HofmeisterAn, line 126: I think we should be able to handle different versions from a Dockerfile. We could use the stages feature to define names, which the method can then resolve:

```Dockerfile
FROM mongo:8 AS mongo_latest
FROM mongo:4.4 AS mongo_v4_4
```

```csharp
Dockerfile.Resolve("mongo_latest");
```

Then we probably can also replace the `CommonImages` classes with that.
But I'm not sure if this works well with Dependabot 🤔.

WDYT?

HofmeisterAn (commented):
As I mentioned, the PR looks pretty good. I just have a few small things we should address before merging. We also need to keep the docs in mind and update them afterward (having a branch ready for the release).

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
digital88: Can you share script please? That's probably Visual Studio messing up as usual.

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
digital88: This is auto ident that Visual Studio does.. We probably need an `.editorconfig` to auto resolve this in the future. Can you please sahre the script or update this idents?

tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs
digital88: I'm not sure about dependabot too, but `GetImageFromDockerfile` helper can already read any line with offset passed as parameter. I can make it search for "stage" (param) instead, good idea.

src/Testcontainers.ActiveMq/Testcontainers.ActiveMq.csproj
digital88: OK, I'll fix it.

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
digital88: I'll fix this

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
digital88: I'm not that good in regex voodoo, probably would be faster for me to manually update it. 😄

tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs
HofmeisterAn: Probably we can reuse some implementations from [here](https://github.com/testcontainers/testcontainers-dotnet/blob/8aa6520dc6fda67a7d001285054f0dc0e8bf2b29/src/Testcontainers/Images/DockerfileArchive.cs#L154-L160).

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
HofmeisterAn: Yes, I can share it when I'm back home.

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
digital88: +

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
digital88: +

src/Testcontainers.ActiveMq/Testcontainers.ActiveMq.csproj
digital88: +

digital88:
Unrelated, but looks like Kafka tests also randomly fail. I already had this failure happen twice in this PR (for unrelated changes).

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
digital88: Looks like we already have `.editorconfig` in the repo :)

I managed to make VS apply it by adding it to the solution.

digital88: Fixed idents by configuring VS

HofmeisterAn:
> Unrelated, but looks like Kafka tests also randomly fail. I already had this failure happen twice in this PR (for unrelated changes).

The best way is to test flaky modules in GH Codespaces with a script that runs the specific module tests indefinitely. I just took a quick look: the Kafka tests (seems to) finish in about a minute, but the overall test process doesn't exit and gets cancelled after five minutes.

tests/Testcontainers.MongoDb.Tests/MongoDbContainerTest.cs
digital88: +

digital88:
> > Unrelated, but looks like Kafka tests also randomly fail. I already had this failure happen twice in this PR (for unrelated changes).
> 
> The best way is to test flaky modules in GH Codespaces with a script that runs the specific module tests indefinitely. I just took a quick look: the Kafka tests (seems to) finish in about a minute, but the overall test process doesn't exit and gets cancelled after five minutes.

Looks like some resource(s) are not disposed properly? We probably need separate issue to track and fix this behaviour.

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
digital88:
```
$path = "C:\Users\digital\github\testcontainers-dotnet\tests"
cd $path
$csprojs = Get-ChildItem -Path "**\*.csproj"
foreach ($proj in $csprojs){
  $s = Get-Content -Raw $proj
  [System.IO.File]::WriteAllLines($proj, $s)
}
```

This did the trick.

digital88:
@HofmeisterAn 
Does something else need to be done/fixed or we are good to merge this?

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
HofmeisterAn: > @HofmeisterAn 
> Does something else need to be done/fixed or we are good to merge this?

I'll have more time tomorrow, so I can probably finish the review then. I think we should also remove these API calls.

tests/Testcontainers.Toxiproxy.Tests/Testcontainers.Toxiproxy.Tests.csproj
HofmeisterAn: 👍 This is similar to what I use. I usually run `git clean -ffdx` beforehand because of the many build files (be careful, it deletes unstaged files).

```powershell
Get-ChildItem -Path . -File -Recurse | ForEach-Object {
    $bytes = [System.IO.File]::ReadAllBytes($_.FullName)
    if ($_.Length -gt 2 -and $bytes[0] -eq 0xEF -and $bytes[1] -eq 0xBB -and $bytes[2] -eq 0xBF) {
        $content = Get-Content -LiteralPath $_.FullName -Encoding UTF8
        [System.IO.File]::WriteAllText($_.FullName, $content, [System.Text.Encoding]::UTF8)
    }
}
```

src/Testcontainers.ActiveMq/ArtemisBuilder.cs
digital88: Makes sense, I removed this line from modules.

digital88:
My dockerfile parsing code in `GetImageFromDockerfile()` is very straightforward, so adding comments in first line(s) broke it :) Probably should ignore lines starting with `#`. I'll fix it.

HofmeisterAn:
> My dockerfile parsing code in `GetImageFromDockerfile()` is very straightforward, so adding comments in first line(s) broke it :) Probably should ignore lines starting with `#`. I'll fix it.

Ah, I just saw your comment, no worries. I've set the images in the Dockerfile to the versions used in the builder classes for now. We should update the images in a follow-up PR (65a2fc005e8bceba21f7e6eed2b80bfd4899c9dd). OK?

digital88:
> > My dockerfile parsing code in `GetImageFromDockerfile()` is very straightforward, so adding comments in first line(s) broke it :) Probably should ignore lines starting with `#`. I'll fix it.
> 
> Ah, I just saw your comment, no worries. I've set the images in the Dockerfile to the versions used in the builder classes for now. We should update the images in a follow-up PR ([65a2fc0](https://github.com/testcontainers/testcontainers-dotnet/commit/65a2fc005e8bceba21f7e6eed2b80bfd4899c9dd)). OK?

Yes, but maybe let dependabot update them instead? :)

As a sidenote, while I was searching for latest images I encountered some obstacles such as:
- Some non-docker hub registries is a pain to navigate and look up image tags. For example I couldn't figure IBM's `icr` and how to see available tags of Db2.
- Some image vendors do not push LTS or even v{Major} tags. They may push current build tagged yyyy-mm-dd. Some even only push `latest` (Microsoft emulators)!
- Even the vendors who push more or less semvered tags do not always list what is currently supported and/or LTS versions are. I need to guess or go to official docs site.
- Some newer images do not work with our tests at all :) Kafka for example.

HofmeisterAn:
I've reviewed everything and made a couple of small changes. LMKWYT. If you're okay with them, we can merge the PR.

> Yes, but maybe let dependabot update them instead? :)

👍 Ideally, we can use Dependabot (for most of the images) and rely on its auto-update capabilities.

> As a sidenote, while I was searching for latest images I encountered some obstacles such as:

Yeah, unfortunately there isn't much we can do here. There just aren't strict guidelines for images.

> Some newer images do not work with our tests at all :) Kafka for example.

Yes, that's expected. It's one of the reasons we're doing this.

These are the follow-ups we should add as TODOs to the issue so we don't forget:

- Prepare a draft PR to update the docs.
- Prepare a draft PR to update the Testcontainers module catalog.
- Update the examples after the `4.10.0` release.

Again, thanks for the work 💪!

digital88:
> LMKWYT. If you're okay with them, we can merge the PR.

I think it's ready for merge. 👍

PR #1610, opened 2025-12-22
feat: Add Platform property to IImage interface

### LINKED ISSUE #1587: [Enhancement]: Support customizing image pull behavior to include a "platform" override
### Problem

This is a copy of this issue in the Java repo: https://github.com/testcontainers/testcontainers-java/issues/9908

### Solution

Have some sort of `WithPlatform` configuration option

### Benefit

Have consistent images pulled for teams with different architectures (ARM64/AMD64 etc). Some images rightfully do not have an ARM64 architecture available and these just don't work at all on ARM-based machines like Macs/MacBook Pros

### Alternatives

None, it doesn't work. I have to disable those tests on my M4 MBP

Yes

HofmeisterAn:
## What does this PR do?

This PR extends the `IImage` interface by adding the `Platform` property. The new property allows developers to specify which platform the container runtime should pull the image for. This is especially important on macOS (Apple Silicon), where some images are not natively available for the host architecture.

When using the Docker Engine API, the runtime does not automatically fall back, e.g. for `linux/arm64` to `linux/amd64`. This is the client's responsibility.

The new property makes it possible to explicitly set the platform when creating a new `IImage` instance using the `DockerImage` implementation, for example:

`_ = new DockerImage("postgres:15.1", "linux/amd64")`

The platform flag is also evaluated when building an image. If the platform is specified in a `Dockerfile`, Testcontainers now respects it and pulls the correct image.

@0xced LMKWYT

## Why is it important?

The PR allows pulling images that do not match the container runtime host architecture.

## Related issues

- Closes #1587

## Summary by CodeRabbit

* **New Features**
  * Platform-aware image support added across the API, including a Platform type for explicit architecture targeting (e.g., linux/amd64, linux/arm64, linux/arm/v6, linux/arm/v7).
  * Dockerfile parsing now recognizes FROM-ARG/platform entries to derive image platforms.

* **Bug Fixes**
  * Platform is propagated to image pull/create and container creation requests for correct platform selection.

* **Documentation**
  * Guidance added for configuring container images with platform-aware options and spec format.

* **Tests**
  * Updated and added tests for platform parsing, serialization, selection, and behavior.

* **Chores**
  * Minor test cleanup (removed unused imports).

### DIFF
--- a/src/Testcontainers/Images/DockerImage.cs
+++ b/src/Testcontainers/Images/DockerImage.cs
@@ -31,12 +31,15 @@ namespace DotNet.Testcontainers.Images
     [CanBeNull]
     private readonly string _digest;
 
+    [CanBeNull]
+    private readonly string _platform;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="DockerImage" /> class.
     /// </summary>
     /// <param name="image">The image.</param>
     public DockerImage(IImage image)
-      : this(image.Repository, image.Registry, image.Tag, image.Digest)
+      : this(image.Repository, image.Registry, image.Tag, image.Digest, image.Platform)
     {
     }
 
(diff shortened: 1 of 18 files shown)


src/Testcontainers/Images/DockerImage.cs
```diff
@@ -50,26 +53,48 @@ public DockerImage(string image)
...
+    /// <param name="image">The image.</param>
+    /// <param name="platform">The platform.</param>
+    /// <example><c>fedora/httpd:version1.0</c> where <c>fedora/httpd</c> is the repository and <c>version1.0</c> the tag.</example>
+    public DockerImage(
+      string image,
+      string platform)
+      : this(GetDockerImage(image))
+    {
+      _platform = TrimOrDefault(platform);
+    }
```
HofmeisterAn, line 68: We should consider a different approach. This is a breaking change and will affect existing configurations.

HofmeisterAn: Either we can:

1. Use a static factory method (`IImage FromImageAndPlatform(string image, string platform)`).
2. Use a distinct parameter type in the ctor (`DockerImage(string image, Platform platform)`).

PR #1619, opened 2026-01-10
(feat): Allow to copy tar stream to container before it starts.

### LINKED ISSUE #1180: [Bug]: Can't copy large files to container
### Testcontainers version

3.8.0

Yes

### Host OS

Windows

### Host arch

x86

### .NET version

8.0.3

### Docker version

```shell
Client:       Podman Engine
Version:      4.9.2
API Version:  4.9.2
Go Version:   go1.21.6
Git Commit:   f9a48ebcfa9a39144be0f86f4ba842752835f945
Built:        Sat Feb  3 00:29:04 2024
OS/Arch:      windows/amd64

Server:       Podman Engine
Version:      4.7.0
API Version:  4.7.0
Go Version:   go1.20.8
Built:        Wed Sep 27 20:24:38 2023
OS/Arch:      linux/amd64
```

### Docker info

```shell
host:
  arch: amd64
  buildahVersion: 1.32.0
  cgroupControllers: []
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.1.7-2.fc38.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.7, commit: '
  cpuUtilization:
    idlePercent: 99.76
    systemPercent: 0.07
    userPercent: 0.17
  cpus: 16
  databaseBackend: boltdb
  distribution:
    distribution: fedora
    variant: container
    version: "38"
  eventLogger: journald
  freeLocks: 2036
  hostname: mlse2068
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 524288
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 524288
      size: 65536
  kernel: 5.15.146.1-microsoft-standard-WSL2
  linkmode: dynamic
  logDriver: journald
  memFree: 30362726400
  memTotal: 33512914944
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: aardvark-dns-1.8.0-1.fc38.x86_64
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.8.0
    package: netavark-1.8.0-2.fc38.x86_64
    path: /usr/libexec/podman/netavark
    version: netavark 1.8.0
  ociRuntime:
    name: crun
    package: crun-1.9.2-1.fc38.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.9.2
      commit: 35274d346d2e9ffeacb22cc11590b0266a23d634
      rundir: /run/user/1000/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
  os: linux
  pasta:
    executable: /usr/bin/pasta
    package: passt-0^20231004.gf851084-1.fc38.x86_64
    version: |
      pasta 0^20231004.gf851084-1.fc38.x86_64
      Copyright Red Hat
      GNU General Public License, version 2 or later
        <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>
      This is free software: you are free to change and redistribute it.
      There is NO WARRANTY, to the extent permitted by law.
  remoteSocket:
    exists: true
    path: /run/user/1000/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: true
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: false
  serviceIsRemote: true
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.2.1-1.fc38.x86_64
    version: |-
      slirp4netns version 1.2.1
      commit: 09e31e92fa3d2a1d3ca261adaeb012c8d75a8194
      libslirp: 4.7.0
      SLIRP_CONFIG_VERSION_MAX: 4
      libseccomp: 2.5.3
  swapFree: 8589934592
  swapTotal: 8589934592
  uptime: 25h 7m 1.00s (Approximately 1.04 days)
  variant: ""
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries:
  search:
  - docker.io
store:
  configFile: /home/user/.config/containers/storage.conf
  containerStore:
    number: 0
    paused: 0
    running: 0
    stopped: 0
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /home/user/.local/share/containers/storage
  graphRootAllocated: 1081101176832
  graphRootUsed: 31850049536
  graphStatus:
    Backing Filesystem: extfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Supports shifting: "false"
    Supports volatile: "true"
    Using metacopy: "false"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 54
  runRoot: /run/user/1000/containers
  transientStore: false
  volumePath: /home/user/.local/share/containers/storage/volumes
version:
  APIVersion: 4.7.0
  Built: 1695839078
  BuiltTime: Wed Sep 27 20:24:38 2023
  GitCommit: ""
  GoVersion: go1.20.8
  Os: linux
  OsArch: linux/amd64
  Version: 4.7.0
```

### What happened?

Trying to use a large file with `.WithResourceMapping` results in an IOException stating that the stream is too long. The file I'm copying is a few gigabytes in size.

### Relevant log output

```shell
System.IO.IOException: Stream was too long.
   at System.IO.MemoryStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.MemoryStream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
--- End of stack trace from previous location ---
   at ICSharpCode.SharpZipLib.Tar.TarBuffer.WriteRecordAsync(CancellationToken ct, Boolean isAsync)
   at ICSharpCode.SharpZipLib.Tar.TarBuffer.WriteBlockAsync(Byte[] buffer, Int32 offset, CancellationToken ct, Boolean isAsync)
   at ICSharpCode.SharpZipLib.Tar.TarOutputStream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken, Boolean isAsync)
   at System.IO.Stream.<CopyToAsync>g__Core|27_0(Stream source, Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at System.IO.Strategies.BufferedFileStreamStrategy.CopyToAsyncCore(Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at DotNet.Testcontainers.Containers.TarOutputMemoryStream.AddAsync(DirectoryInfo directory, FileInfo file, UnixFileModes fileMode, CancellationToken ct) in /_/src/Testcontainers/Containers/TarOutputMemoryStream.cs:line 145
```

### Additional information

_No response_

digital88:
## What does this PR do?

This PR adds a new method to `ContainerBuilder` which allows to copy a `Stream` containing tar archive into container, leveraging already available method `IDockerContainerOperations.ExtractArchiveToContainerAsync`. I changed `IDockerContainerOperations.ExtractArchiveToContainerAsync` signature to match underlying `Docker.Dotnet` API which takes a base `Stream` instead of `TarOutputmemoryStream`.

The catch is that the user must build tar stream on it's own.

At first I was thinking about adding some class/methods to allow users build tar archive stream using TC library, but then I decided not to because:

- There is a couple of options to create tar archive, which gives users more freedom to choose from.
- This is not something I think TC library should do out of the box anyway (not in our "domain").
- We currently use `SharpZipLib` library in TC internals, but it looks like stopped being maintained. Probably not a good idea to create more functionality on top of it. Not using this library further may also allow us to drop it as a dependency later.
- The `SharpZipLib` API is not that trivial to build custom functionality on top of.
- I made some tests which show how to create and manipulate tar archives using `SharpZipLib` library and built-in `System.IO` (starting from NET 7). We may add this tests as a documentation to show users some guidance.

Should close https://github.com/testcontainers/testcontainers-dotnet/issues/1180

Also may help in cases like this: https://github.com/testcontainers/testcontainers-dotnet/issues/1486

## Why is it important?

This change will allow uses to add arbitrary files and directories to container. The existing API `WithResourceMapping` currently have some limitations:

- it will not copy files larger than ~2Gb.
- some overloads inner implementation differs from other overloads, users have to carefully read each overload method documentation (some overloads assume `target` is directory path, other assume it's file path).

## Summary by CodeRabbit

* **New Features**
  * Stream-based tar archive copy support: copy tar archives into containers with configurable target path and new public APIs and configuration options to declare tar-archive resource mappings.
  * Containers now process tar-archive mappings during startup and runtime copy workflows.

* **Tests**
  * Added end-to-end tests covering tar archive copies using FileStream, MemoryStream, and library-generated tars; verify files inside containers and UID handling.

### DIFF
--- a/src/Testcontainers/Configurations/Containers/TarArchiveMapping.cs
+++ b/src/Testcontainers/Configurations/Containers/TarArchiveMapping.cs
@@ -1,0 +1,17 @@
+using System.IO;
+
+namespace DotNet.Testcontainers.Configurations.Containers
+{
+  public sealed record TarArchiveMapping
+  {
+    public TarArchiveMapping(Stream tarArchive, string containerPath)
+    {
+      TarArchive = tarArchive;
+      ContainerPath = containerPath;
+    }
+
+    public Stream TarArchive { get; }
+
+    public string ContainerPath { get; }
+  }
+}
--- a/src/Testcontainers/Builders/ContainerBuilder`3.cs
+++ b/src/Testcontainers/Builders/ContainerBuilder`3.cs
@@ -9,6 +9,7 @@ namespace DotNet.Testcontainers.Builders
   using System.Threading.Tasks;
   using Docker.DotNet.Models;
   using DotNet.Testcontainers.Configurations;
+  using DotNet.Testcontainers.Configurations.Containers;
   using DotNet.Testcontainers.Containers;
   using DotNet.Testcontainers.Images;
   using DotNet.Testcontainers.Networks;
@@ -269,6 +270,13 @@ namespace DotNet.Testcontainers.Builders
       {
         return WithResourceMapping(new UriResourceMapping(source, target, uid, gid, fileMode));
       }
+    }
+
+    /// <inheritdoc />
+    public TBuilderEntity WithCopyTarArchive(Stream tarArchive, string containerPath = "/")
+    {
+      var tarArchiveMappings = new[] { new TarArchiveMapping(tarArchive, containerPath) };
+      return Clone(new ContainerConfiguration(tarArchiveMappings: tarArchiveMappings));
     }
 
     /// <inheritdoc />
--- a/src/Testcontainers/Builders/IContainerBuilder`2.cs
+++ b/src/Testcontainers/Builders/IContainerBuilder`2.cs
@@ -324,6 +324,14 @@ namespace DotNet.Testcontainers.Builders
     TBuilderEntity WithResourceMapping(Uri source, string target, uint uid = 0, uint gid = 0, UnixFileModes fileMode = Unix.FileMode644);
 
     /// <summary>
+    /// Copies a tar archive contents to the container before it starts.
+    /// </summary>
+    /// <param name="tarArchive">The <see cref="Stream"/> with the tar archive contents.</param>
+    /// <param name="containerPath">The path where tar archive contents should be placed.</param>
+    /// <returns>A configured instance of <typeparamref name="TBuilderEntity" />.</returns>
+    TBuilderEntity WithCopyTarArchive(Stream tarArchive, string containerPath = "/");
+
+    /// <summary>
     /// Assigns the mount configuration to manage data in the container.
     /// </summary>
     /// <param name="mount">The mount configuration.</param>
--- a/src/Testcontainers/Builders/IContainerBuilder`2.cs
+++ b/src/Testcontainers/Builders/IContainerBuilder`2.cs
@@ -326,6 +326,9 @@ namespace DotNet.Testcontainers.Builders
     /// <summary>
     /// Copies a tar archive contents to the container before it starts.
     /// </summary>
+    /// <remarks>
+    /// Set the <paramref name="tarArchive"/> property <see cref="Stream.Position"/> to 0 before calling this method.
+    /// </remarks>
     /// <param name="tarArchive">The <see cref="Stream"/> with the tar archive contents.</param>
     /// <param name="containerPath">The path where tar archive contents should be placed.</param>
     /// <returns>A configured instance of <typeparamref name="TBuilderEntity" />.</returns>
--- a/src/Testcontainers/Builders/IContainerBuilder`2.cs
+++ b/src/Testcontainers/Builders/IContainerBuilder`2.cs
@@ -34,6 +34,7 @@ namespace DotNet.Testcontainers.Builders
     /// <param name="acceptLicenseAgreement">A boolean value indicating whether the license agreement is accepted.</param>
     /// <returns>A configured instance of <typeparamref name="TBuilderEntity" />.</returns>
     /// <exception cref="InvalidOperationException">Thrown when the module does not require a license agreement.</exception>
+    [PublicAPI]
     TBuilderEntity WithAcceptLicenseAgreement(bool acceptLicenseAgreement);
 
     /// <summary>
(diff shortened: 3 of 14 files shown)


src/Testcontainers/Configurations/Containers/TarArchiveMapping.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+  public sealed record TarArchiveMapping
+  {
+    public TarArchiveMapping(Stream tarArchive, string containerPath)
+    {
+      TarArchive = tarArchive;
+      ContainerPath = containerPath;
+    }
+
+    public Stream TarArchive { get; }
```
HofmeisterAn, line 13: I was wondering if we should change [`Task<byte[]> GetAllBytesAsync(CancellationToken)`](https://github.com/testcontainers/testcontainers-dotnet/blob/f776ff935625ae9af2a7f0bc92f70574fac3a966/src/Testcontainers/Configurations/Volumes/IResourceMapping.cs#L36C9-L36C10) to return a `Stream` instead, after reviewing the PR. Then we need a stream that can handle more than 2GB to build the tarball. Wouldn't that improve a lot?

HofmeisterAn (commented):
Thanks a lot for all the effort you've put in so far. I'm a bit uncertain, especially when I look at the tests and the number of steps developers need to take care of. I believe that (like some other TC features) this is something Docker.DotNet should ideally support OOB. Now that we have control over the version included in TC for .NET, we can add these kinds of features to Docker.DotNet or at least improve the API to make it easier for developers and us to integrate it into TC for .NET. WDYT?

> This is not something I think TC library should do out of the box anyway (not in our "domain")

We should offer an API that accepts a Stream, I agree. However, one of the reasons TC became so popular and successful is because it's easy to use and handles many things developers don't need to worry about. So, I believe it's important to still offer an API that makes copying files to and from the container as easy as possible (like it is today, but without the 2GB limit).

> * I made some tests which show how to create and manipulate tar archives using `SharpZipLib` library and built-in `System.IO` (starting from NET 7). We may add this tests as a documentation to show users some guidance.

This is a good idea. Maybe we can add a proper abstraction so we can use `System.IO` for everything new, and maintain backward compatibility for legacy frameworks and old versions using `SharpZipLib`.

src/Testcontainers/Configurations/Containers/TarArchiveMapping.cs
digital88: Yes, this will improve a lot. The `FileStream` probably will work best as a backing store of `TarOutputMemoryStream`.

digital88:
I have some more ideas in this PR's context:

- We should drop `TarOutputMemoryStream` and introduce a new `Builder`-like internal API. Reason is that every `IResourceMapping` makes a separate `TarOutputMemoryStream`, writes into it and then copies it into container. This can be batched in single `Stream` to reduce overhead.

src/Testcontainers/Configurations/Containers/TarArchiveMapping.cs
digital88: We should probably hide inner tar `Stream` from user, that will give use more freedom to (re)design internal APIs like `TarOutputMemoryStream`. At first I had the feeling that out API should require user to provide `Stream` of thier own and designed this PR around this idea.

PR #1624, opened 2026-01-17
fix(EventHubs): Workaround health API timeout

HofmeisterAn:
## What does this PR do?

Since the EventHubs image is not versioned, we unfortunately ran into an issue with the latest release. It appears that the health API call fails when the HTTP request is sent too early, causing the client to time out.

While we could catch the timeout or cancellation exception and retry once, the default HTTP timeout is 100 seconds, which would make users wait unnecessarily. It would be better for the EventHubs team to address this issue properly.

The same issue also occurs when using the CLI.

You can find more information here: https://github.com/Azure/azure-event-hubs-emulator-installer/issues/69#issuecomment-3762895979.

## Why is it important?

Unblock our CI.

## Related issues

\-

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Event Hubs container initialization reliability with an enhanced wait strategy that now includes a startup delay before performing health status checks. This addresses timing-related issues that could prevent containers from starting properly and ensures more stable and predictable initialization behavior across deployment and integration testing workflows.

### DIFF
--- a/src/Testcontainers.EventHubs/EventHubsBuilder.cs
+++ b/src/Testcontainers.EventHubs/EventHubsBuilder.cs
@@ -168,8 +168,10 @@ public sealed class EventHubsBuilder : ContainerBuilder<EventHubsBuilder, EventH
         return base.Init()
             .WithPortBinding(EventHubsPort, true)
             .WithPortBinding(EventHubsHttpPort, true)
-            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request =>
-                request.ForPort(EventHubsHttpPort).ForPath("/health")));
+            .WithWaitStrategy(Wait.ForUnixContainer()
+                .AddCustomWaitStrategy(new WorkaroundGhIssue69())
+                .UntilHttpRequestIsSucceeded(request =>
+                    request.ForPort(EventHubsHttpPort).ForPath("/health")));
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


HofmeisterAn:
**Edit**: A better version without an artificial delay:

https://github.com/testcontainers/testcontainers-dotnet/blob/dde205fbfb719b1291c32481d5a7b8701bfd4f42/src/Testcontainers.EventHubs/EventHubsBuilder.cs#L171-L175

TC users can work around this by overriding the default wait strategy and adding an artificial delay ([`WorkaroundGhIssue69`](https://github.com/Azure/azure-event-hubs-emulator-installer/issues/69#issuecomment-3762895979)) before sending the first health API call.

```csharp
using DotNet.Testcontainers.Configurations;
using DotNet.Testcontainers.Containers;

...

_ = new EventHubsBuilder("mcr.microsoft.com/azure-messaging/eventhubs-emulator:latest")
    .WithAcceptLicenseAgreement(true)
    // Override the default wait strategy.
    .WithWaitStrategy(Wait.ForUnixContainer()
        .AddCustomWaitStrategy(new WorkaroundGhIssue69())
        .UntilHttpRequestIsSucceeded(request =>
            request.ForPort(EventHubsBuilder.EventHubsHttpPort).ForPath("/health")))
    .Build();
        
...

private sealed class WorkaroundGhIssue69 : IWaitUntil
{
    {
        await Task.Delay(TimeSpan.FromSeconds(5))
            .ConfigureAwait(false);

        return true;
    }
}
```

src/Testcontainers.EventHubs/EventHubsBuilder.cs
```diff
@@ -168,8 +168,10 @@ protected override EventHubsBuilder Init()
         return base.Init()
             .WithPortBinding(EventHubsPort, true)
             .WithPortBinding(EventHubsHttpPort, true)
-            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(request =>
-                request.ForPort(EventHubsHttpPort).ForPath("/health")));
+            .WithWaitStrategy(Wait.ForUnixContainer()
+                .AddCustomWaitStrategy(new WorkaroundGhIssue69())
+                .UntilHttpRequestIsSucceeded(request =>
```
scrocquesel, line 173: @HofmeisterAn could it be possible to loop on http call with a smaller (configurable) http timeout.

HofmeisterAn: No, I don't think this is possible without changing the `HttpWaitStrategy` implementation.

PR #1635, opened 2026-02-14
feat: Add Temporal module

### LINKED ISSUE #1530: [Enhancement]: Support Temporalio dev container
### Problem

Temporalio dev container: https://hub.docker.com/r/temporalio/temporal
Temporal: https://temporal.io/

### Solution

Create a new nuget package `Testcontainers.Temporalio`

### Benefit

Standard temporalio testcontainer package, helping integration tests using temporal dev server on CI.

### Alternatives

Build on your own _all the time_:

```csharp
container = new ContainerBuilder()
            .WithImage("temporalio/temporal:latest")
            .WithCommand("server", "start-dev", "--ip", "0.0.0.0")
            .WithPortBinding(GrpcPort, true)
            .WithPortBinding(HttpPort, true)
            .WithWaitStrategy(Wait.ForUnixContainer().UntilHttpRequestIsSucceeded(r => r
                .ForPath("/api/v1/namespaces")
                .ForPort(HttpPort)))
            .Build();
```

Yes

bgener:
## What does this PR do?

Adds a new Testcontainers.Temporalio module that wraps the https://temporal.io/ dev server for integration testing.

## Why is it important?

Developers using Temporal for workflow orchestration need a reliable way to spin up a Temporal server for integration tests. Today, each team manually configures Docker containers which is repetitive, error-prone, and time consuming.

## Related issues

- Closes #1530 

## How to test this PR

Build and run tests via build script:

` ./build.sh --target Test --test-project Testcontainers.Temporalio`

 Or run tests directly:

`dotnet test tests/Testcontainers.Temporalio.Tests/`

## Summary by CodeRabbit

* **New Features**
  * Added Temporal container support to spin up Temporal instances for testing, with gRPC and Web UI endpoints, configurable ports, and built-in readiness checks.
* **Tests**
  * Added an integration test suite to validate Temporal container lifecycle and basic workflow operations.


src/Testcontainers.Temporalio/TemporalBuilder.cs
```diff
@@ -0,0 +1,176 @@
...
+    {
+        return Merge(DockerResourceConfiguration, new TemporalConfiguration(dbFilename: path));
+    }
+
+    /// <inheritdoc />
+    public override TemporalContainer Build()
+    {
+        Validate();
+
+        var command = new List<string> { "server", "start-dev", "--ip", "0.0.0.0" };
```
HofmeisterAn: Please move this to init. The `WithCommand()` API appends commands it does not repleace them OOB.

src/Testcontainers.Temporalio/TemporalBuilder.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        if (DockerResourceConfiguration.Namespaces != null && DockerResourceConfiguration.Namespaces.Any())
+        {
+            foreach (var ns in DockerResourceConfiguration.Namespaces)
+            {
+                command.Add("--namespace");
+                command.Add(ns);
+            }
+        }
```
HofmeisterAn: These can be moved to the actual public API and set the command set right away (`WithCommand(...)`). There is no need to store the value in the configuration instance. We only store values there if we need them in or after `Build()`. This is not the case here.

We only provide APIs for configurations that are absolutely necessary. If users can extend the default module builder configuration using the standard APIs, we don't add extra ones on top, so we don't have to maintain them across different container image versions (breaking changes, backwards compatibility).

src/Testcontainers.Temporalio/TemporalBuilder.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// <inheritdoc />
+    protected override TemporalBuilder Init()
+    {
+        return base.Init()
+            .WithPortBinding(TemporalGrpcPort, true)
+            .WithPortBinding(TemporalHttpPort, true)
+            .WithConnectionStringProvider(new TemporalConnectionStringProvider())
+            .WithWaitStrategy(Wait.ForUnixContainer()
+                .UntilInternalTcpPortIsAvailable(TemporalGrpcPort)
+                .UntilInternalTcpPortIsAvailable(TemporalHttpPort)
```
HofmeisterAn:
```suggestion
```

Are they reeally necessary?

tests/Testcontainers.Temporalio.Tests/TemporalContainerTest.cs
```diff
@@ -0,0 +1,56 @@
+namespace Testcontainers.Temporalio;
```
HofmeisterAn: The tests are very detailed, which is great to see. However, I think we could simplify the builder module and reduce the tests to the essentials, just verifying that the container starts and running one or two basic workflows to confirm it handles requests properly.

As a maintainer, my priority is to focus on what is necessary to ensure a reliable service instance for the main use cases.

HofmeisterAn (changes requested):
Thanks, the PR looks overall very good. I have a few minor requests/questions.

src/Testcontainers.Temporalio/TemporalBuilder.cs
bgener: Removed the HTTP port check, but kept the gRPC check because the health endpoint's success does not guarantee the external gRPC listener on port 7233 has finished binding

src/Testcontainers.Temporalio/TemporalBuilder.cs
bgener: Done, moved to the Init

src/Testcontainers.Temporalio/TemporalBuilder.cs
bgener: Totally makes sense, I've performed the code cleanup

tests/Testcontainers.Temporalio.Tests/TemporalContainerTest.cs
bgener: Simplified the tests and removed the excessive tests

HofmeisterAn (commented):
Thanks for the quick adjustments! I slightly aligned some parts with the repo standards during my lunch break. If you're okay with that, I'm happy to merge the PR. Thanks again.

bgener:
> Thanks for the quick adjustments! I slightly aligned some parts with the repo standards during my lunch break. If you're okay with that, I'm happy to merge the PR. Thanks again.

@HofmeisterAn It looks good, let's go ahead, thanks

HofmeisterAn (approved):
Thanks again.

Issue #1642, opened 2026-03-02
[Bug]: Resource Reuse doesn't work for ServiceBusContainer


franciscosamuel:
### Testcontainers version

4.10.0

### Using the latest Testcontainers version?

Yes

### Host OS

Windows

### Host arch

x86_64

### .NET version

8

### Docker version

```shell
Client:       Podman Engine
Version:      5.7.1
API Version:  5.7.1
Go Version:   go1.25.5
Git Commit:   f845d14e941889ba4c071f35233d09b29d363c75
Built:        Wed Dec 10 14:54:51 2025
OS/Arch:      windows/amd64

Server:       Podman Engine
Version:      5.2.1
API Version:  5.2.1
Go Version:   go1.22.5
Built:        Wed Aug 14 01:00:00 2024
OS/Arch:      linux/amd64
```

### Docker info

```shell
Client:
  APIVersion: 5.7.1
  Built: 1765378491
  BuiltTime: Wed Dec 10 14:54:51 2025
  GitCommit: f845d14e941889ba4c071f35233d09b29d363c75
  GoVersion: go1.25.5
  Os: windows
  OsArch: windows/amd64
  Version: 5.7.1
host:
  arch: amd64
  buildahVersion: 1.37.1
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.10-1.fc40.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.10, commit: '
  cpuUtilization:
    idlePercent: 99.81
    systemPercent: 0.11
    userPercent: 0.09
  cpus: 20
  databaseBackend: sqlite
  distribution:
    distribution: fedora
    variant: container
    version: "40"
  eventLogger: journald
  freeLocks: 2039
  hostname: DESKTOP-0LLKJQJ
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 6.6.87.2-microsoft-standard-WSL2
  linkmode: dynamic
  logDriver: journald
  memFree: 12629606400
  memTotal: 16654704640
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: aardvark-dns-1.12.1-1.fc40.x86_64
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.12.1
[...]

PR #1643, opened 2026-03-02
fix(ServiceBus): Ignore runtime property to support reuse

franciscosamuel:
## What does this PR do?

Makes the configuration hash of ServiceBusContainer stable by removing the IDatabaseContainer from the hashed configuration.

## Why is it important?

The IDatabaseContainer contains runtime information about the database container such as the start time which changes everytime, resulting in an always changing configuration hash, which breaks the Resource Reuse feature for ServiceBus containers.

## Related issues

- Relates #1642

## How to test this PR

Since this concerns cross-execution tests, an automated test would become quite convoluted. The easiest way to test this is to temporarily modify the configurations on ServiceBusCustomMsSqlConfiguration and DatabaseFixture to both use .WithReuse(true) and running the test twice. The first execution should create the containers and the second should reuse them.

## Follow-ups

An analogous change needs to be made for EventHubs. I will do it in a separate PR

## Summary by CodeRabbit

* **Chores**
  * Improved Service Bus configuration JSON serialization handling to prevent unintended property export.

### DIFF
--- a/src/Testcontainers.ServiceBus/ServiceBusConfiguration.cs
+++ b/src/Testcontainers.ServiceBus/ServiceBusConfiguration.cs
@@ -57,5 +57,6 @@ public sealed class ServiceBusConfiguration : ContainerConfiguration
     /// <summary>
     /// Gets the database container.
     /// </summary>
+    [JsonIgnore]
     public IDatabaseContainer DatabaseContainer { get; }
 }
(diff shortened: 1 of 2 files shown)


HofmeisterAn (approved):
Thanks for the contribution!

Something came to mind while I was thinking about this again. It's probably best to use the reuse hash from the actual dependent container instead of ignoring it.

We can merge this for now to unblock your use case, and then add it properly as a feature in a separate PR.

franciscosamuel:
I didn't mention it in the issue, but while investigating I also thought of that and my thought process was:

1. the ReuseHash is calculated at the configuration level 
2. the configuration only knows about the container itself, not its configuration
3. we are already relying on the user to pass in a configured container in order to use the Resource Reuse feature

So I believe this is actually the correct way to do it, at least until we have a pattern to pass down WithReuse to dependant containers.

Issue #1677, opened 2026-04-13
[Enhancement]: Nvidia Triton test container


ezsilmar:
### Problem

No wrapper for Nvidia Triton, an open source inference server:
* https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/index.html
* https://github.com/triton-inference-server/server

### Solution

Implement `Testcontainers.Triton` exposing its http and grpc endpoints.

### Benefit

Ease integration testing of Triton from C# code. Whoever needs it would benefit from an already existing package.

### Alternatives

Using the container builder directly. Keeping the Triton wrapper hidden in the internal codebase.

### Would you like to help contributing this enhancement?

Yes

ezsilmar:
Hello, just to give some context at my company we've migrated direct FluentDocker usages to test containers. It was a really nice experience as this project saves a lot of code and things like port mapping and resource cleanup are handled automatically. The only technology we needed to write a custom wrapper for happened to be Triton. 

I'd like to contribute that upstream basically to not keep it in our codebase and in hope it can be relevant to someone else as well. It's not that popular tech like Cassandra or Temporal but I still think it could be interesting for the community.

Thanks!

HofmeisterAn:
Thanks for the feedback and the contribution. I'll review the PR by the end of the week. If I haven't reviewed it by then, please feel free to ping me.

ezsilmar:
Closing as I realized the generic container builder is good enough for my usecase and that I don't have a complex enough setup to create a good wrapper.

PR #1678, opened 2026-04-13
Introduce Testcontainers.Triton

ezsilmar:
## What does this PR do?

Introduce `Testcontainers.Triton` package which wraps Nvidia Triton inference server:
* https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/index.html
* https://github.com/triton-inference-server/server

## Why is it important?

Avoids setting up a custom wrapper, gives a pre-configured alternative to anyone who may need it.

## Related issues

Solves https://github.com/testcontainers/testcontainers-dotnet/issues/1677

## Summary by CodeRabbit

* **New Features**
  * Added Triton inference server container support with configurable HTTP, gRPC, and metrics endpoints and endpoint helpers
  * New solution projects enable Triton runtime and test integration

* **Chores**
  * Centralized gRPC package versions updated to 2.76.0

* **Tests**
  * Added tests covering readiness, metrics, gRPC health, connection string, and a simple inference scenario


src/Testcontainers.Triton/TritonBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <inheritdoc />
+    protected override TritonBuilder Init()
+    {
+        return base.Init()
+            .WithPortBinding(TritonHttpPort, true)
+            .WithPortBinding(TritonGrpcPort, true)
+            .WithPortBinding(TritonMetricsPort, true)
+            .WithConnectionStringProvider(new TritonConnectionStringProvider())
+            .WithWaitStrategy(Wait.ForUnixContainer()
+                .UntilMessageIsLogged("Started GRPCInferenceService"));
```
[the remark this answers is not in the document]
ezsilmar: So the problem here is that Triton needs to have models to return 200 on health request. Models can be loaded in a variety of ways (remote storage uri, local folder, etc.) and the configuration is passed via cmd like `tritonserver --model-repository ./models`. But as `WithCommand`‘s most common overload appends the parameters and the string must also start with `tritonserver` it may create confusion if we put some default path. Current behavior allows the container to start and be reported as healthy without any configuration from the user, however they will still need to call `WithCommand` to specify the models and other parameters to actually use it in tests. I think that’s the least confusing way of implementing that but we can discuss if you see better options.

src/Testcontainers.Triton/TritonBuilder.cs
```diff
@@ -0,0 +1,105 @@
...
+    protected override void Validate()
+    {
+        base.Validate();
+
+        const string message = "Triton Inference Server requires a command to be specified. " +
+            "Use .WithCommand(\"tritonserver\", \"--model-repository=/path/to/models\") to configure the server.";
+
+        _ = Guard.Argument(DockerResourceConfiguration.Command, nameof(DockerResourceConfiguration.Command))
+            .ThrowIf(argument => argument.Value == null || !argument.Value.Any(), _ => new ArgumentException(message));
+    }
```
HofmeisterAn, line 74: I wrote the following before the recent changes, and I think it's better. WDZT about a dedicated API?

---

I think with the current implementation, getting started with the module is difficult. Users probably won't realize they need to pass an additional command to run the container/module.

In Testcontainers, we often take an opinionated approach to modules. I suggest we add something like `WithModelRepository(...)` that applies the necessary pre-configuration (including common setups like a remote host, etc.).

We could then throw an exception on validation if the `Command` property doesn't specify any model, with a clear error message explaining what's missing. That way, users understand what the module expects.

Testcontainers modules don't need to support every possible configuration or edge case, just the most important ones. If users need more flexibility, they can always fall back to the generic container builder or fully override the commands.

ezsilmar: What bothers me is that if I implement some minimal model loading, like

```csharp
var defaultModelsDir = "/some/path/without/collisions"
base.Init()
  .WithTmpfsMount(defaultModelsDir)
  .WithCommand("tritonserver", $"--model-repository={defaultModelsDir}")
```

It would mean that the following `WithModelRepository` would need to modify the command line instead of just calling `WithCommand`. And it may get trickier given that the user might call `WithCommand` for different command line parameters before/after/multiple times, override the command completely with `OverwriteEnumerable`, etc. Not sure how brittle these helpers would be but I agree it's possible and it would help to ease a simple usecase, like `WithLocalModelRepository(string dirToMount)` and `WithRemoteModelRepository(string url)`. 

Regarding the validation I feel like it could be too restrictive. We could check that the command begins with `tritonserver` and includes `--model-repository` but what if the user wants to do something before calling `tritonserver` (the entrypoint in this container seem to allow that). On the other hand with a working zero-conf container we could just skip the validation.

HofmeisterAn: > It would mean that the following `WithModelRepository` would need to modify the command line instead of just calling `WithCommand`.

What we usually do is set the binary in `Init()` and then append or override the command. This is a very common pattern in TC, like:

https://github.com/testcontainers/testcontainers-dotnet/blob/a2f571718f58ae45e321576b93a65a1c782f01ee/src/Testcontainers.Azurite/AzuriteBuilder.cs#L134-L135

> And it may get trickier given that the user might call `WithCommand` for different command line parameters before/after/multiple times, override the command completely with `OverwriteEnumerable`, etc.

This is a general "issue" with pre-configured modules. Their purpose is to cover common use cases, not all kinds of configurations. E.g. if you look at Kafka, it pre-configures a lot, if users start to override the module configuration (entrypoint, cmd, etc.), it might not work anymore. As I said, if users need full flexibility, they can use the generic builder. Modules are pre-configured, opinionated, containing best practices and serve as a good example. Usually I expect modules to add some kind of additional value.

Does `tritonserver` always require a model repository? If so, I would recommend at least validating `--model-repository` and providing an API to apply the configuration. Otherwise, we could check whether any `Command` is set and, if not, throw a meaningful exception that explains what is expected from the developer (not my favorite).

WDYT?

ezsilmar: Hello, after looking more into it and investigating our internal usecase deeper, I believe that a generic `ContainerBuilder` would suit us better this time.

I realized the larger community would certainly expect to run GPU models with Testcontainers.Triton which requires significant investment to make things easy: smth like `.WithGpuSupport()` which will construct correct `DeviceRequests` (and there are driver parameters involved). Same for model repository, indeed implementing dedicated helpers for the most common options would be more aligned with the testcontainers purpose. I also learnt that a relatively common case is to use multiple repositories at the same time, with combined auth configurations, etc.

Internally, even though we use GPU models, we don't really have a usecase to test them from inside the application, i.e. mocked CPU models are enough, as well as locally mounted model repository. So I unfortunately do not have a dogfooding setup complex enough to create a good "opinionated" container wrapper. Otherwise, I'd have pushed forward with this PR (and I will if the situation changes and our container builder code starts to grow / I have something I'm sure is useful).

So sorry for taking your time to review this. I'm very grateful for the discussion and learnt a lot from it.

HofmeisterAn: No worries. Then we can close the issue and PR?

PR #1707, opened 2026-06-16
feat(CosmosDb): Add get method AccountEndpoint

### LINKED ISSUE #1704: [Enhancement]: CosmosDb: Add getters for AccountEndpoint and AccountKey
### Problem

In my test setup, I need to get the ComsmosDb container's `AccountEndpoint` and `AccountKey` properties to configure my services.

`CosmosDbContainer` exposes the connection string that contains these properties. But it would be more convenient to have those properties directly exposed rather than having to parse the connection string

https://github.com/testcontainers/testcontainers-dotnet/blob/638a926de1253b444192d66c6b2c88993552e4f1/src/Testcontainers.CosmosDb/CosmosDbContainer.cs#L20-L26

### Solution

Add getters or get methods to return both `AccountEndpoint` and `AccountKey` and rewrite `GetConnectionString` to use those getters.

### Benefit

Getting the properties in a more convenient way, while keeping the `CosmosDbContainer` class easy to read and maintain.

### Alternatives

Get them by parsing the connection string as done today.

Yes

srollinet:
## What does this PR do?
Add getters for Cosmos DB AccountEndpoint and AccountKey

## Why is it important?

Add a convenient way to access these properties

## Related issues

- Closes #1704 

## Summary by CodeRabbit

* **New Features**
  * Added a public helper to retrieve the Cosmos DB emulator account endpoint URI.

* **Refactor**
  * Updated Cosmos DB connection string generation to populate the account endpoint using the new helper, rather than building it inline.

* **Documentation**
  * Refreshed XML documentation for Azurite endpoint helper methods (no behavior changes).

### DIFF
--- a/src/Testcontainers.CosmosDb/CosmosDbContainer.cs
+++ b/src/Testcontainers.CosmosDb/CosmosDbContainer.cs
@@ -21,7 +21,7 @@ public sealed class CosmosDbContainer : DockerContainer
     {
         var properties = new Dictionary<string, string>();
         properties.Add("AccountEndpoint", GetAccountEndpoint());
-        properties.Add("AccountKey", GetAccountKey());
+        properties.Add("AccountKey", CosmosDbBuilder.DefaultAccountKey);
         return string.Join(";", properties.Select(property => string.Join("=", property.Key, property.Value)));
     }
 
@@ -42,15 +42,6 @@ public sealed class CosmosDbContainer : DockerContainer
     public string GetAccountEndpoint()
     {
         return new UriBuilder(Uri.UriSchemeHttp, Hostname, GetMappedPublicPort(CosmosDbBuilder.CosmosDbPort)).ToString();
-    }
-    
-    /// <summary>
-    /// Gets the CosmosDb account key
-    /// </summary>
-    /// <returns>The CosmosDb account key</returns>
-    public string GetAccountKey()
-    {
-        return CosmosDbBuilder.DefaultAccountKey;
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Testcontainers.CosmosDb/CosmosDbContainer.cs
```diff
@@ -34,6 +34,24 @@ public string GetConnectionString()
...
+    }
+    
+    /// <summary>
+    /// Gets the CosmosDb account key
+    /// </summary>
+    /// <returns>The CosmosDb account key</returns>
+    public string GetAccountKey()
+    {
+        return CosmosDbBuilder.DefaultAccountKey;
+    }
```
HofmeisterAn:
```suggestion
```

The builder already makes it publicly available.

srollinet: My rational to have an explicit getter is to avoid questioning
1) you have to know that the `DefaultAccountKey` exists on the builder
2) you have to know that the actual account key is always the default one.

So I thought it was clearer to put it on the container.
If you think it is not necessary, I can remove it.

HofmeisterAn: All default configurations in the modules are exposed through the builder. The container instances IIRC only provide runtime information. For now, I would prefer to keep this consistent. We can add a remarks section explaining the differences in the get methods for the connection string and account endpoint, including a reference to the key.

HofmeisterAn (approved):
Thanks

PR #1719, opened 2026-06-29
fix: Restore tar archive write performance regressed by padding trim

HofmeisterAn:
## What does this PR do?

The block factor of 1 introduced to fix the Podman EPIPE race (https://github.com/podman-container-tools/buildah/pull/6678) was inadvertently applied to the Dockerfile build context path (`DockerfileArchive`), which writes to a FileStream on disk. At a block factor of 1, SharpZipLib flushes on every 512-byte block, significantly degrading image build performance.

/cc @Rob-Hague @artiomchi

## Why is it important?

\-

## Related issues

- Relates https://github.com/testcontainers/testcontainers-dotnet/pull/1684#issuecomment-4825305495

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tar archive creation to reduce intermittent failures when generating or uploading tar-based build contexts in container environments.

* **Documentation**
  * Expanded guidance on tar block factor usage, including why a value of `1` is recommended to avoid padding-related issues that can surface as broken-pipe errors, and cautioned against using the setting for Dockerfile build context due to performance impact.

### DIFF
--- a/src/Testcontainers/TarArchiveDefaults.cs
+++ b/src/Testcontainers/TarArchiveDefaults.cs
@@ -4,18 +4,18 @@ namespace DotNet.Testcontainers
   {
     // A block factor of 1 keeps the record size equal to the block size (512 B),
     // so SharpZipLib writes no extra zero padding beyond the two standard EOF
-    // blocks. The default factor of 20 produces ~8 KB of trailing zeros, which
-    // can trigger a race in Podman's archive handler. The tar subprocess exits
-    // after the EOF blocks while the HTTP sender is still flushing the padding,
-    // resulting in an EPIPE (HTTP 500 "broken pipe"). See:
-    // https://github.com/testcontainers/testcontainers-dotnet/issues/1683.
+    // blocks. The default factor of 20 produces up to 9 KiB of trailing zero
+    // padding, which can trigger a race in Podman's archive handler. The tar
+    // subprocess exits after the EOF blocks while the HTTP sender is still
+    // flushing the padding, resulting in an EPIPE (HTTP 500 "broken pipe").
+    // See: https://github.com/testcontainers/testcontainers-dotnet/issues/1683.
     //
     // The race only affects PUT /containers/{id}/archive (resource mappings via
     // a MemoryStream). Buildah's copierHandlerPut returned immediately after
     // tar.Reader reached io.EOF without draining the pipe. POST /build is immune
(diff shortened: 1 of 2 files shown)


src/Testcontainers/TarArchiveDefaults.cs
```diff
@@ -2,13 +2,26 @@ namespace DotNet.Testcontainers
...
+    // tar.Reader reached io.EOF without draining the pipe. POST /build is immune
+    // because containers/storage's chrootarchive.Untar always drains stdin after
+    // extraction. The server-side fix landed in buildah. See:
+    // https://github.com/podman-container-tools/buildah/pull/6678.
+    // Keeping the block factor at 1 stays correct for older Podman versions.
+    //
+    // Do NOT use this for the Dockerfile build context, which is written to a
+    // FileStream on disk: at 512 B per record, SharpZipLib flushes on every
+    // block, which regresses image build performance. That path intentionally
+    // keeps SharpZipLib's default block factor.
```
HofmeisterAn, line 24:
```suggestion
    // A block factor of 1 keeps the record size equal to the block size (512 B),
    // so SharpZipLib writes no extra zero padding beyond the two standard EOF
    // blocks. The default factor of 20 produces up to 9 KiB of trailing zero
    // padding, which can trigger a race in Podman's archive handler. The tar
    // subprocess exits after the EOF blocks while the HTTP sender is still
    // flushing the padding, resulting in an EPIPE (HTTP 500 "broken pipe").
    // See: https://github.com/testcontainers/testcontainers-dotnet/issues/1683.
    //
    // The race only affects PUT /containers/{id}/archive (resource mappings via
    // a MemoryStream). Buildah's copierHandlerPut returned immediately after
    // tar.Reader reached io.EOF without draining the pipe. POST /build is immune
    // because containers/storage's chrootarchive.Untar always drains stdin after
    // extraction. The server-side fix landed in buildah.
    // See: https://github.com/podman-container-tools/buildah/pull/6678.
    // Keeping the block factor at 1 stays correct for older Podman versions.
    //
    // Do NOT use this for the Dockerfile build context, which is written to a
    // FileStream on disk: at 512 B per record, SharpZipLib flushes on every
    // block, which regresses image build performance. That path intentionally
    // keeps SharpZipLib's default block factor.
```

Rob-Hague:
Thanks 👍

PR #1731, opened 2026-08-07
fix(MongoDb): Treat an already initiated replica set as success

arnelirobles:
Fixes #1722

## The problem

`UnsafeStartAsync` invokes the startup callback on every start, so a reused container runs `InitiateReplicaSetAsync` against a replica set that is already initiated.

The script it runs is not idempotent:

```js
var r = rs.initiate({_id:"rs0", members:[{_id:0, host:"127.0.0.1:27017"}]});
quit(r.ok === 1 ? 0 : 1);
```

On a set that already exists, `rs.initiate` **throws** rather than returning a result, so `var r` is never assigned and mongosh exits non-zero:

```
MongoServerError: already initialized
  code: 23
  codeName: 'AlreadyInitialized'
```

`WaitStrategy.WaitUntilAsync` then retries the script on its interval. The default wait strategy timeout is one hour, which is why the reporter saw the same `mongosh` exec repeat "without ever resolving".

## The fix

Catch the error and treat `AlreadyInitialized` as success, so the startup callback is idempotent:

```js
try { var r = rs.initiate({...}); quit(r.ok === 1 ? 0 : 1); }
catch (e) { quit(e.codeName === "AlreadyInitialized" ? 0 : 1); }
```

Any other server error still exits non-zero and is still retried, so a genuinely failing initiation behaves as before.

## Testing

`MongoDbReplicaSetReuseTest.ReusedContainerStartsAgain` starts the same replica set container twice with `WithReuse(true)`, asserts the second start reused the first container, and asserts the replica set is healthy afterwards.

- Before the change: times out (bounded to 3 minutes in the test so a regression fails rather than hangs).
- After the change: passes in about 7 seconds.
- Full `Testcontainers.MongoDb.Tests` suite: 19 passed, 0 failed.

The test bounds its own wait because the default wait strategy timeout is one hour; without a bound a regression would stall a CI run rather than fail it.

## A related defect, not addressed here

While testing this I first wrote the regression test as a stop/start restart rather than reuse, and hit a different problem in `WaitIndicateReadiness`:

```csharp
var (stdout, stderr) = await container.GetLogsAsync(since: container.StoppedTime, timestampsEnabled: false);
return _count.Equals(... .Count(line => line.Contains("Waiting for connections")));
```

It asserts the count is **exactly** 1 or 2 over logs taken since `StoppedTime`, which is not restart-safe in either direction:

- After a restart the window spans the previous and the new boot, so the count exceeds `_count` and the equality can never hold. The container start hangs.
[...]

PR #1656, opened 2026-03-08
fix(MongoDb): Wait for post-init startup readiness before replica set initiation

HofmeisterAn:
## What does this PR do?

The PR changes how the MongoDB replica set is initiated. Instead of initiating it immediately, it now waits until the final process is running.

The official MongoDB image starts `mongod` twice during the first-time initialization (bootstrap + final process). Waiting prevents the replica set from being initiated while this handover is in progress, which could otherwise cause the container startup to fail (container exits).

## Why is it important?

We occasionally saw MongoDB tests fail because the container exited during startup with:

    {"msg":"Error setting up listener","attr":{"error":{"code":9001,"codeName":"SocketException","errmsg":"Address already in use"}}}

## Related issues

- Closes #1636

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MongoDB replica set initialization reliability by adding a readiness check before starting the replica set setup, preventing potential startup conflicts.

Issue #1636, opened 2026-02-14
[Bug]: MongoDB stops while starting


HofmeisterAn:
### Testcontainers version

4.10.0

### Using the latest Testcontainers version?

Yes

### Host OS

Linux

### Host arch

x86

### .NET version

10.0.100

### Docker version

```shell
-
```

### Docker info

```shell
-
```

### What happened?

Sometimes our MongoDB CI tests fail. The issue is reproducible in GH Codespaces by building the project and running the tests repeatedly until they fail. It looks like a race condition during the MongoDB container startup, likely caused by high load on the CI runner. I wasn't able to reproduce it locally or on a larger GH Codespaces machine.

The MongoDB container stops while starting. When comparing the logs with a successful run, this line stands out.

    {"msg":"Error setting up listener","attr":{"error":{"code":9001,"codeName":"SocketException","errmsg":"Address already in use"}}}

It seems the socket isn't properly closed.

From what I understand, MongoDB creates a temporary socket to initialize the admin user, then stops and starts again.

It looks like there's a chance that when:

> MongoDB init process complete; ready for start up.

the socket hasn't been fully closed, and starting MongoDB fails.

I don't think there's much we can do. I created the bug for documentation purposes. This is probably something the MongoDB team needs to address and fix in their container image.

### Relevant log output

```shell
2026-02-13T20:13:52.225535110Z about to fork child process, waiting until server is ready for connections.
2026-02-13T20:13:52.231093960Z forked process: 27
2026-02-13T20:13:52.231944406Z
2026-02-13T20:13:52.231957030Z {"t":{"$date":"2026-02-13T20:13:52.231+00:00"},"s":"I",  "c":"CONTROL",  "id":20698,   "ctx":"-","msg":"***** SERVER RESTARTED *****"}
2026-02-13T20:13:52.240485549Z {"t":{"$date":"2026-02-13T20:13:52.240+00:00"},"s":"I",  "c":"CONTROL",  "id":23285,   "ctx":"-","msg":"Automatically disabling TLS 1.0, to force-enable TLS 1.0 specify --sslDisabledProtocols 'none'"}
[...]

PR #1735, opened 2026-08-09
fix(MongoDb): Probe the server instead of counting log messages before initiating

### LINKED ISSUE #1732: [Bug]: MongoDb WaitIndicateReadiness counts log lines from the previous run, so a restarted container either hangs or is reported ready too early
### Testcontainers version

4.11.0 (`develop`, `1d329ad1`)

Yes

### Host OS

macOS

### Host arch

arm64

### .NET version

10

### Docker version

Docker Desktop, engine 28.x

### What happened?

`WaitIndicateReadiness` in `MongoDbBuilder` is not safe across a container restart, in both directions.

```csharp
// src/Testcontainers.MongoDb/MongoDbBuilder.cs
{
    var (stdout, stderr) = await container.GetLogsAsync(since: container.StoppedTime, timestampsEnabled: false)
        .ConfigureAwait(false);

    return _count.Equals(Array.Empty<string>()
        .Concat(stdout.Split(LineEndings, StringSplitOptions.RemoveEmptyEntries))
        .Concat(stderr.Split(LineEndings, StringSplitOptions.RemoveEmptyEntries))
        .Count(line => line.Contains("Waiting for connections")));
}
```

`_count` is 1 or 2 depending on whether credentials are configured, and the comparison is **exact equality** over the logs since `StoppedTime`. After a restart that window covers the previous run as well as the new one, so:

- **Too many matches: the start hangs.** The count exceeds `_count` and the equality can never become true. For a replica set container this surfaces inside `InitiateReplicaSetAsync`, which awaits this readiness check before initiating. The default wait strategy timeout is one hour, so the caller appears to hang rather than fail.

- **Too few matches, at the wrong time: the start returns early.** A single stale `Waiting for connections` line from the previous run can satisfy the check before the new `mongod` is listening. The next command then fails with `MongoNetworkError: connect ECONNREFUSED 127.0.0.1:27017`.

### Steps to reproduce

Add a restart test to `MongoDbContainerTest`:

```csharp
[Fact]
public async Task RestartsSuccessfully()
{
    using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(2));

    await _mongoDbContainer.StopAsync(cts.Token);
    await _mongoDbContainer.StartAsync(cts.Token);

    var execResult = await _mongoDbContainer.ExecScriptAsync("db.adminCommand({ping:1}).ok;", cts.Token);
    Assert.True(0L.Equals(execResult.ExitCode), execResult.Stderr);
}
```

Observed across the existing fixtures:

| Fixture | Result |
|---|---|
| `MongoDbNoAuthConfiguration` | fails in ~3s, `MongoNetworkError: connect ECONNREFUSED 127.0.0.1:27017` |
| `MongoDbReplicaSetDefaultConfiguration` | hangs until the bound, `TimeoutException` from `WaitStrategy.WaitUntilAsync` via `InitiateReplicaSetAsync` |
| `MongoDbNamedReplicaSetConfiguration` | same as above |
| `MongoDbV5Configuration` | same as above |
| `MongoDbDefaultConfiguration`, `MongoDbV4Configuration` | pass |

### Expected result / actual result

**Expected:** a stopped container that is started again becomes ready, and the readiness check reflects the current run.

**Actual:** the readiness check counts log lines from the previous run, so a restart either hangs until the wait strategy times out or is reported ready before `mongod` is accepting connections.

### Additional information

Two things worth deciding together:

1. Comparing with `>=` instead of exact equality would fix the hang but not the early return, since a stale line still counts.
2. Scoping the log window to the current run (rather than `StoppedTime`, which refers to the previous stop) addresses both, but I am not sure what the intended semantics of `StoppedTime` are here.

Found while working on #1722; deliberately kept out of that pull request since it is a separate defect. Happy to open a PR once the preferred direction is clear.

arnelirobles:
Fixes #1732

## The problem

The readiness check that runs before the replica set is initiated counts occurrences of a log message and compares that count for **equality**:

```csharp
var (stdout, stderr) = await container.GetLogsAsync(since: container.StoppedTime, timestampsEnabled: false);
return _count.Equals(... .Count(line => line.Contains("Waiting for connections")));
```

`_count` is 1 or 2. The count only grows, and the module does not control what else writes that text, so the check fails in both directions:

- **Too many occurrences and the start hangs.** The equality can never become true. The default wait strategy timeout is one hour, so the caller does not fail, it appears to hang.
- **A stale occurrence and the start returns early.** The check passes before `mongod` is serving, and the next command fails with `MongoNetworkError: connect ECONNREFUSED 127.0.0.1:27017`.

@xamir82's reproduction on the issue shows this needs no restart: extra markers written by an init script during a single run are enough, and whether it fails at all depends on when the once-per-second poll samples.

This arrived in 4.11.0. #1656 (`ec6b50f6`) put this check on the replica set path; before that the path did not use it.

## The fix

Ask the server instead of reading its logs:

```csharp
private const string ScriptContent = "try{rs.status();quit(0);}catch(e){quit(e.codeName===\"NotYetInitialized\"?0:1);}";
```

This still satisfies what #1656 needed for #1636. The official image forks a temporary `mongod` during first-time initialization to create the root user, and that process is **not** started with `--replSet`, so it never reports `NotYetInitialized`. Only the final server does.

Measured on `mongo:6.0.27` with the module's own configuration (auth plus the generated keyfile), polling twice a second from container start:

| t | authenticated ping | `rs.status()` | "Waiting for connections" |
|---|---|---|---|
| 1.0s | fails | no reply | 1 |
| 2.0s | succeeds | `NotYetInitialized` | 2 |

A plain `ping` is not sufficient on its own; against a container without the keyfile it succeeds during the bootstrap phase. `rs.status()` is what distinguishes the two processes.

It is also idempotent for an already initiated set, which returns `ok: 1` and exits zero.

## Scope

Limited to the replica set path, which is where #1656 introduced this. `Build()` still uses the log-counting strategy for non-replica-set containers, and it has the same weakness, but changing it alters long-standing behaviour for every MongoDb container and seemed better decided separately. Happy to extend this PR if you would prefer both in one go.

## Testing

`MongoDbReplicaSetReadinessTest` follows @xamir82's reproduction: a replica set container with an init script that writes the marker twice more.

- Before: times out (bounded to three minutes in the test so a regression fails rather than stalls the run).
- After: passes in about seven seconds.
- Full `Testcontainers.MongoDb.Tests`: 19 passed, 0 failed, including every replica set fixture, which is what covers the first-time initialization path from #1636.

Separate from #1731, which is about `rs.initiate` not being idempotent. The two touch the same method but neither depends on the other.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MongoDB replica-set startup detection by verifying replication status directly.
  - Containers now start reliably when readiness messages appear multiple times.
  - Replica sets in the intermediate `NotYetInitialized` state are handled correctly.

- **Tests**
  - Added regression coverage for duplicate readiness messages and successful replica-set initialization.

### DIFF
--- a/src/Testcontainers.MongoDb/MongoDbBuilder.cs
+++ b/src/Testcontainers.MongoDb/MongoDbBuilder.cs
@@ -198,7 +198,7 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
             return;
         }
 
-        var readiness = new WaitIndicateReadiness(configuration);
+        var readiness = new WaitReplicationEnabled();
 
         // This is a simple workaround to use the default options, which can be configured
         // with custom configurations as needed.
@@ -227,6 +227,32 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
     }
 
     /// <inheritdoc cref="IWaitUntil" />
+    /// <inheritdoc cref="IWaitUntil" />
+    private sealed class WaitReplicationEnabled : IWaitUntil
+    {
+        // rs.status() only answers once the final mongod is serving. The official image forks a
+        // temporary mongod during first-time initialization to create the root user, and that
+        // process is not started with --replSet, so it never reports NotYetInitialized. This
+        // preserves the handover guarantee from #1636 without counting log messages, whose
+        // number depends on log content the module does not control: #1732.
+        private const string ScriptContent = "try{rs.status();quit(0);}catch(e){quit(e.codeName===\"NotYetInitialized\"?0:1);}";
+
+        /// <inheritdoc />
+        public Task<bool> UntilAsync(IContainer container)
+        {
+            return UntilAsync(container as MongoDbContainer);
+        }
+
+        /// <inheritdoc cref="IWaitUntil.UntilAsync" />
+        private static async Task<bool> UntilAsync(MongoDbContainer container)
+        {
+            var execResult = await container.ExecScriptAsync(ScriptContent)
+                .ConfigureAwait(false);
+
+            return 0L.Equals(execResult.ExitCode);
+        }
+    }
+
     private sealed class WaitIndicateReadiness : IWaitUntil
     {
         private static readonly string[] LineEndings = { "\r\n", "\n" };
--- a/src/Testcontainers.MongoDb/MongoDbBuilder.cs
+++ b/src/Testcontainers.MongoDb/MongoDbBuilder.cs
@@ -130,7 +130,7 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
         }
         else
         {
-            waitUntil = new WaitInitiateReplicaSet();
+            waitUntil = new WaitReplicaSetPrimary();
         }
 
         // If the user does not provide a custom waiting strategy, append the default MongoDb waiting strategy.
@@ -198,7 +198,7 @@ public sealed class MongoDbBuilder : ContainerBuilder<MongoDbBuilder, MongoDbCon
             return;
         }
 
-        var readiness = new WaitReplicationEnabled();
+        var readiness = new WaitReplicaSetEnabled();
 
         // This is a simple workaround to use the default options, which can be configured
         // with custom configurations as needed.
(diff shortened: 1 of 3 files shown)


arnelirobles:
One more argument for this direction that I missed when opening the PR: **testcontainers-go already gates replica set readiness on `rs.status()` rather than on log output.**

```go
// modules/mongodb/mongodb.go
req.WaitingFor = wait.ForAll(
    req.WaitingFor,
    wait.ForExec(cli.eval("rs.status().ok")),
).WithDeadline(60 * time.Second)
```

https://github.com/testcontainers/testcontainers-go/blob/main/modules/mongodb/mongodb.go#L207

So this is not a new idea, it brings the .NET module in line with how the Go one already handles the same problem. Same command, and the same reason: `rs.status()` answers only once the final `mongod` is serving, and it does not depend on log content.

HofmeisterAn (approved):
Thanks

xamir82:
Sorry if this isn't the right place to ask this, but when will this fix be release as a NuGet package version?

PR #1612, opened 2025-12-30
feat: Require explicit container image in Testcontainers.Xunit

0xced:
## What does this PR do?

This pull request introduce a new parameterless `Configure()` method on the `ContainerLifetime` class, hence also on the derived `(Db)ContainerFixture` and `(Db)ContainerTest` classes.

It also makes the `TBuilderEntity Configure(TBuilderEntity builder)` method obsolete.

## Why is it important?

This matches the changes discussed in #1540 and introduced in #1584.

## Related issues

This is a follow-up to #1584. Closes #1598.

/cc @digital88

## How to test this PR

All relevant tests have been adapted to use the new parameterless `Configure()` method.

## Follow-ups

When the parameterless constructors (now obsolete) are eventually removed from the builders, the following changes will need to be performed.

1. Remove the `new()` constraint on `ContainerLifetime`, `ContainerFixture`, `ContainerTest`, `DbContainerFixture` and `DbContainerTest`
2. Make the `ContainerFixture` class abstract
3. Change `protected virtual TBuilderEntity Configure() => new();` to `protected abstract TBuilderEntity Configure();` on ContainerLifetime, forcing consumers to provide the image name/tag when creating builders.
4. Delete the obsolete methods: `protected virtual TBuilderEntity Configure(TBuilderEntity builder)`

## Summary by CodeRabbit

* **Refactor**
  * Standardized container configuration to a parameterless builder initialization across fixtures, simplifying how builders are created and extended.

* **Bug Fixes**
  * Updated derived fixtures to obtain base configuration before applying additional options (e.g., wait strategies, credentials).

* **Documentation**
  * Added XML docs and examples for the new configuration extension point and marked the old overload as deprecated.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1767, opened 2026-09-17
feat: Add Testcontainers.TUnit package

### LINKED ISSUE #1766: [Enhancement]: Support for TUnit
### Problem

There doesn't seem to be a TestContainers.TUnit package on NuGet.

You should.  Please support TUnit.

### Solution

Publish a TestContainers.TUnit package on NuGet

### Benefit

Use TestContainers with TUnit so all our tests can be TUnit rather than XUnit.v3

### Alternatives

Have to keep our TestContainers Integration tests on XUnit.v3 which we would prefer not to.

Yes

zcsizmadia:
## What does this PR do?

Adds a `Testcontainers.TUnit` package, the TUnit counterpart to `Testcontainers.Xunit` and `Testcontainers.XunitV3`, together with a test project, documentation, and the build changes needed to run it in CI.

The package deliberately follows the design of the existing xUnit.net test framework packages: the same class names, the same lifecycle semantics, the same ADO.NET helpers, and the same example tests and documentation structure. Anyone familiar with `Testcontainers.Xunit` should be able to read the TUnit package without learning a new model. Where the two differ, it is because TUnit offers a native mechanism (class data sources, `[DependsOn]`, ambient test context) for something xUnit.net needs a sink, an orderer, or an extension for.

**Package.** The public surface mirrors the xUnit.net packages: `ContainerTest`, `ContainerFixture`, `DbContainerTest`, and `DbContainerFixture`, built on a shared `ContainerLifetime` base class.

- `ContainerLifetime` implements TUnit's `IAsyncInitializer` and `IAsyncDisposable`. TUnit starts the container before a test runs and disposes of it when the owning scope ends. As in the xUnit.net packages, a start failure is captured and rethrown on `Container` access so the test fails with the actual cause.
- `ContainerTest` gives each test its own container, because TUnit creates a new test class instance per test. `ContainerFixture` is injected with `[ClassDataSource<TFixture>]` and supports every `SharedType` (per class, per assembly, per test session, keyed) without extra attributes or extensions.
- Container startup honors `TestContext.Current.Execution.CancellationToken`, matching the xUnit.net v3 behavior from #1431.
- Log messages are written to the output of the TUnit context that is current when the message is logged. Logs emitted by a shared fixture during a test therefore appear in that test's output instead of a separate diagnostic stream.
- Because TUnit sets injected properties before it initializes an instance, a fixture can declare its own `[ClassDataSource<...>]` properties (for example a network or a dependent container) and use them inside `Configure()`. This is documented in the guide.
- `Configure()` is abstract. The obsolete `Configure(TBuilderEntity)` overload and the `configure` constructor callback from the xUnit.net packages are not carried over, since parameterless module builders are obsolete as well (#1612).
- The package references `TUnit.Core` 1.68.4, the package intended for libraries that provide TUnit infrastructure without the runner. The source generator, polyfills, and implicit usings are disabled for the library. Target frameworks match the xUnit.net packages.
- The framework-neutral files (`DbContainerTestMethods`, `IDbContainerTestMethods`, `Logger`, `NullScope`) are copied rather than shared, because the xUnit.net sources are bound to the `Testcontainers.Xunit` namespace and `Testcontainers.XunitV3` already relies on compiling them as they are.

**Tests.** `tests/Testcontainers.TUnit.Tests` ports the Redis and PostgreSQL examples of the xUnit.net test project. The shared Redis example orders the dependent test with `[DependsOn]` instead of a custom test case orderer, and the PostgreSQL example uses the injected `CancellationToken`. Since the xUnit.net module tests exercise the xUnit.net base classes as a side effect and the TUnit package does not benefit from that, the project also contains a shared PostgreSQL `DbContainerFixture` example that uses every ADO.NET helper method, and a test for the postponed start exception. 16 tests, about 90% block coverage of the library.

**Build.** On the .NET 10 SDK, `dotnet test` refuses to run Microsoft.Testing.Platform projects through VSTest. The Cake `Test` task now detects projects that set `TestingPlatformDotnetTestSupport` (TUnit sets it to `true`, xunit.v3 sets it to `false`) and runs them with `dotnet run`, passing the platform's equivalents of the VSTest options: TRX report, code coverage, results directory, and test filter. Coverage is produced in the Visual Studio coverage XML format, and `sonar.cs.vscoveragexml.reportsPaths` is added next to the existing OpenCover path. The xUnit.net projects are unaffected and still run through VSTest.

**Docs.** A `Testing with TUnit` guide is added next to the xUnit.net guide, built from the same snippet markers in the test project. Package versions are managed centrally and the test project is registered for CI via `.runs-on` and the database module filter.

## Why is it important?

TUnit is increasingly used for integration tests and has first-class support for shared, asynchronously initialized fixtures, which is exactly the pattern the xUnit.net packages exist for. Without a package, TUnit users have to hand-roll the lifetime management, the ADO.NET helpers, and the log forwarding that `Testcontainers.Xunit` already provides. Modeling the new package on the xUnit.net one keeps the two consistent for users who maintain tests in both frameworks, and keeps the maintenance surface familiar for the project.

## Related issues

- Closes #1766

## How to test this PR

```shell
./build.sh --target=Build --test-project=Testcontainers.TUnit
./build.sh --target=Test --test-project=Testcontainers.TUnit
```

The `Test` target writes `Testcontainers.TUnit.Tests.trx` and `Testcontainers.TUnit.Tests.coverage.xml` to `test-results`. To confirm the VSTest path is unchanged, run the same two targets with `--test-project=Testcontainers.XunitV3` and check that `coverage.opencover.xml` is still produced.

The package can be inspected with `dotnet pack src/Testcontainers.TUnit -c Release`. Its only dependencies are `Testcontainers` and `TUnit.Core`.

## Follow-ups

- NuGet Trusted Publishing: the first push of a brand-new package ID depends on the owner's policy on nuget.org allowing new packages. Worth checking before the release that includes this PR.
- The `--blame-hang-timeout` safeguard has no direct Microsoft.Testing.Platform equivalent and is not applied to platform projects. The TUnit test project only runs on the Linux runner, where hangs have not been an issue.
- Attaching container logs to the TUnit test output on failure would be a natural TUnit-specific addition, left out here to keep the scope aligned with the xUnit.net packages.

## Summary by CodeRabbit

- **New Features**
  - Added TUnit integration with isolated per-test and shared container fixtures.
  - Added database container helpers for connections, commands, batches, and asynchronous operations.
  - Added Redis and PostgreSQL examples for container and database testing.
- **Documentation**
  - Added TUnit documentation covering fixtures, database testing, configuration, and test lifecycles.
- **Build & Testing**
  - Added TUnit projects to the solution and CI test execution, filtering, and coverage reporting.


zcsizmadia:
@HofmeisterAn The remaining CI failure is unrelated to this PR: the `Testcontainers.Minio` tests fail because Docker Hub no longer serves `minio/minio` (or `minio/mc`). The repositories return 404 now; the "pull access denied" message is just how Docker Hub reports a missing repository to anonymous pulls. The images disappeared between 2026-09-11 and 2026-09-12 (the last green `develop` run on 2026-09-10 still pulled them), which lines up with MinIO winding down its Docker Hub publishing after archiving the community edition. The remaining red jobs are fail-fast cancellations caused by that one failure.

The same tag the module pins, `RELEASE.2023-01-31T02-24-19Z`, is still available at `quay.io/minio/minio` without login, so switching `MinioBuilder.MinioImage` to the Quay address would fix `develop` and every open PR. I can send that as a separate PR if you like.

zcsizmadia:
One idea that came out of the Minio failure, for a separate PR if it is of interest: a small CI job that runs `docker manifest inspect` against every module's default image constant before the test matrix starts. It needs no Docker daemon state and no image pulls, finishes in well under a minute, and would report a removed or renamed image as a single named failure instead of surfacing it inside an unrelated PR after a fail-fast cancellation of the whole matrix. Sixteen module defaults already live outside Docker Hub (Quay, GHCR, MCR, GCR, ICR), so the check would cover all registries the same way.

zcsizmadia:
@dependabot rebase

PR #1768, opened 2026-09-18
fix: Prevent Docker ignore patterns from matching path prefixes

Benziza:
## What does this PR do?

Match Docker ignore patterns against complete path segments. The matcher previously used a regex word boundary, which also matches before punctuation such as a hyphen. The first pattern had no boundary check at all.

Add regression cases for similarly named directories and for the built-in `.idea` and `.vs` patterns.

## Why is it important?

With `node_modules` in `.dockerignore`, Testcontainers currently removes `node_modules-old/file.txt` from the Docker build context. The file belongs to a different directory and should remain available to Dockerfile `COPY` instructions.

The built-in `**/.idea` pattern can likewise remove files under `src/.ideality/`.

## How to test this PR

- Run `dotnet test tests/Testcontainers.Tests/Testcontainers.Tests.csproj --filter FullyQualifiedName~IgnoreFileTest --configuration Release`.
- Verify `node_modules/file.txt` is excluded and `node_modules-old/file.txt` is included.
- Verify `src/.idea/file.json` is excluded and `src/.ideality/file.json` is included.

The focused test run passed (46 tests). Docker integration tests were not run because the Docker daemon is unavailable in the local environment.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ignore-pattern matching to respect complete path segments.
  - Prevented similarly named paths from being incorrectly ignored, including `node_modules-old`, `bin-old`, `obj-old`, `.idea-old`, and `.vs-old`.

- **Tests**
  - Expanded coverage for valid ignore patterns, including `node_modules`, `.idea`, and `.vs` paths.


HofmeisterAn:
Looks like MinIO dropped the Docker Hub image... I'll remove the module later and merge this PR. Thanks for the contribution!

HofmeisterAn (approved):

PR #1770, opened 2026-09-20
docs: Clarify that the startup callback replaces module defaults

### LINKED ISSUE #1661: [Bug]: User's module builder configuration gets overwritten by the final `Build()`
### Testcontainers version

4.10.0

Yes

### Host OS

Windows

### Host arch

x86

### .NET version

10

### Docker version

```shell
Docker version 29.2.1, build a5c7197
```

### Docker info

```shell
Client:
 Version:    29.2.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  agent: create or run AI agents (Docker Inc.)
    Version:  v1.27.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-agent.exe
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.18.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-ai.exe
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.31.1-desktop.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v5.1.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-compose.exe
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.47
    Path:     C:\Program Files\Docker\cli-plugins\docker-debug.exe
  desktop: Docker Desktop commands (Docker Inc.)
    Version:  v0.3.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-desktop.exe
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.31
    Path:     C:\Program Files\Docker\cli-plugins\docker-extension.exe
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v1.4.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-init.exe
  mcp: Docker MCP Plugin (Docker Inc.)
    Version:  v0.40.2
    Path:     C:\Program Files\Docker\cli-plugins\docker-mcp.exe
  model: Docker Model Runner (Docker Inc.)
    Version:  v1.1.5
    Path:     C:\Program Files\Docker\cli-plugins\docker-model.exe
  offload: Docker Offload (Docker Inc.)
    Version:  v0.5.56
    Path:     C:\Program Files\Docker\cli-plugins\docker-offload.exe
  pass: Docker Pass Secrets Manager Plugin (beta) (Docker Inc.)
    Version:  v0.0.24
    Path:     C:\Program Files\Docker\cli-plugins\docker-pass.exe
  sandbox: Docker Sandbox (Docker Inc.)
    Version:  v0.12.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-sandbox.exe
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-sbom.exe
  scout: Docker Scout (Docker Inc.)
    Version:  v1.20.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-scout.exe

Server:
 Containers: 12
  Running: 0
  Paused: 0
  Stopped: 12
 Images: 137
 Server Version: 29.2.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Discovered Devices:
  cdi: docker.com/gpu=webgpu
 Swarm: inactive
 Runtimes: nvidia runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: dea7da592f5d1d2b7755e3a161be07f43fad8f75
 runc version: v1.3.4-0-gd6d73eb8
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
 Kernel Version: 5.15.153.1-microsoft-standard-WSL2
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 20
 Total Memory: 15.46GiB
 Name: docker-desktop
 ID: eac52a52-36e3-4060-a6fc-2f4c8d68b4d7
 Debug Mode: false
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 No Proxy: hubproxy.docker.internal
 Labels:
  com.docker.desktop.address=npipe://\\.\pipe\docker_cli
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables

WARNING: No blkio throttle.read_bps_device support
WARNING: No blkio throttle.write_bps_device support
WARNING: No blkio throttle.read_iops_device support
WARNING: No blkio throttle.write_iops_device support
WARNING: Support for cgroup v1 is deprecated and planned to be removed by no later than May 2029 (https://github.com/moby/moby/issues/51111
```

### What happened?

``` chsarp 
Kafka = new KafkaBuilder("confluentinc/cp-kafka:7.8.1")
     .WithEnvironment("KAFKA_CREATE_TOPICS", "topic-name")
     .WithStartupCallback((container, ct) => { return Task.CompletedTask; })
     .WithPortBinding(9092, true)
     .Build();
// ...
await Kafka.StartAsync()
```
When I successfully start the container the startup callback isn't invoked.
I suspect the reason for that is that my custom callback is potentially [overwritten here](https://github.com/testcontainers/testcontainers-dotnet/blob/87e0b929d26f27c960436d3fc7d49e4a9ae6071c/src/Testcontainers.Kafka/KafkaBuilder.cs#L238),

If this is the case this bug also applies to other containers.

### Relevant log output

```shell
-
```

### Additional information

_No response_

arnelirobles:
## What does this PR do?

Documents that the startup callback is replace-only, in the three places someone would look: the XML docs on both `WithStartupCallback` overloads, the API table and composition section in the docs, and `KafkaBuilder.Build()`.

Closes #1661.

## Why is it important?

From the issue:

> That said, we should make it clearer for developers when a configuration is overwritten by `Build()`, so users are aware of this behavior.

Reading through it, the overwrite goes both ways, and the direction the issue reports is the rarer one.

`ContainerConfiguration` combines `StartupCallback` with the scalar `BuildConfiguration.Combine`, which is last write wins. Nine module builders set a callback. Eight of them set it in `Init()`, so a user calling `WithStartupCallback` replaces the module's own provisioning: MongoDb never runs `rs.initiate`, MsSql never creates the database, Pulsar, Redpanda and FakeGcsServer never get the startup script their command waits for. Nothing errors. The wait strategy just times out, and nothing in the message points at the callback.

Kafka is the one that matches the issue title. It sets the callback in `Build()` at line 238, after the user's calls, so the module wins there.

While confirming that I found a second one in Kafka that is not in the issue. If the user has not set a consensus protocol, `Build()` calls `WithKRaft()` or `WithZooKeeper()`, and both end with `WithWaitStrategy(...)`. `WaitStrategies` is combined with the scalar overload too, so a wait strategy the user set is dropped. Calling `WithKRaft()` or `WithZooKeeper()` explicitly avoids it. Documented on `Build()` alongside the callback note.

## How was it tested?

Docs only, no behaviour change. `Testcontainers` and `Testcontainers.Kafka` both build clean, no `CS1574` or other cref warnings from the new `<see>` references.

## Notes

I kept this to documentation because that is what the issue asked for, and because modules being opinionated is a deliberate design choice.

If you would rather Kafka composed instead, it is a small local change: capture `kafkaBuilder.DockerResourceConfiguration.StartupCallback` before setting the module's own, then invoke it from the new callback using the three argument overload. Roughly eight lines, contained to `KafkaBuilder`, no effect on the other modules. Happy to add it here or in a follow-up if you want it, just say the word.

Could you add the `documentation` label so release drafter files it correctly.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that startup callbacks run after container startup and before wait strategies.
  * Documented last-call-wins behavior for builder configurations, including how callbacks can replace module provisioning.
  * Warned that replacing provisioning may leave containers unprovisioned and cause wait-strategy timeouts.
  * Recommended the generic container builder when full startup callback control is required.


arnelirobles:
Good catch, and it was worse than the arity. Fixed in 9efea9a.

`IContainerBuilder` does take three type parameters, so `WithWaitStrategy` was not resolving. That one now uses the three parameter form and resolves.

The `WithStartupCallback` overload is a separate problem. It takes `Func<TContainerEntity, CancellationToken, Task>`, an open type parameter of the declaring interface, and I could not get a cref to bind to it from inside `Testcontainers.Kafka`. Closing it with `KafkaContainer` does not match the signature, and leaving it open does not resolve from another assembly. Rather than keep guessing I dropped it to `<c>WithStartupCallback</c>`, since the sentence names the method clearly enough.

Worth noting the original build was clean because this repo does not emit XML docs by default, so `CS1574` never fires. I checked this one by building with `GenerateDocumentationFile=true` and grepping the generated XML for unresolved `cref="!:` entries. Zero across all five target frameworks now, and `Testcontainers` itself is clean too.

HofmeisterAn (commented):
Better documentation of the builder API and how it behaves sounds like a good idea. The issue you mention isn't specific to Kafka or the `WithStartupCallback`. It's a general design decision that affects all modules and all builder (container, image, network, volume) configurations:

- Building a container is immutable, it always returns a new configuration. This is useful for [A/B testing](https://dotnet.testcontainers.org/api/create_docker_container/#reusing-builder-configurations), for example.
- A `WithX` builder call overwrites the previously configured value.
- The exception is lists and dictionaries, where new values are appended by default.
- Lists and dictionaries cannot be modified by default.
- The exception is those that support [`ComposableEnumerable<T>`](https://dotnet.testcontainers.org/api/create_docker_container/#composing-command-arguments) through the builder.
- Modules come pre-configured, and that configuration is opinionated. If users override it using the default container builder APIs, we "don't" support it.

arnelirobles:
Makes sense, thanks. Reworked in 2abf6b2: one general "How builder configurations combine" section built from the rules you listed, and I dropped the startup callback and Kafka specific notes since overriding module defaults isn't supported anyway.

arnelirobles:
@HofmeisterAn heads up on CodeRabbit's latest outside-diff note, which suggests documenting `WithWaitStrategy` as appending. I left the wording as is because it replaces: `ContainerConfiguration` merges `WaitStrategies` with the single-value `Combine<IEnumerable<WaitStrategy>>`. I also checked it at runtime: one `WithWaitStrategy` call stores 2 strategies (running check plus custom), and a second call still leaves 2, not 4. Modules like MongoDb and PostgreSql also skip their own wait strategy when one is already set, so the "not supported" example still holds.

